PR #26, opened 2025-08-09
Feature/customize themes

s0up4200:


s0up4200:
After reconsideration, this will not make the cut.

PR #27, opened 2025-08-10
feat(config): add configurable database path

s0up4200:
- Added optional DatabasePath configuration to Config struct
- Support custom database location via config file or QUI__DATABASE_PATH
- Maintain backward compatibility (database next to config if not specified)
- Useful for read-only config directories
- Updated documentation with new configuration option
- Added configuration loading and path determination logic

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,10 @@ QUI__SESSION_SECRET=...  # Auto-generated if not set
 
 # Logging
 QUI__LOG_LEVEL=INFO      # Options: ERROR, DEBUG, INFO, WARN, TRACE
+QUI__LOG_PATH=...        # Optional: log file path
+
+# Storage
+QUI__DATABASE_PATH=...   # Optional: custom database location (default: next to config)
 ```
 
 ## API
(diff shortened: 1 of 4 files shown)


cmd/server/main.go
```diff
@@ -54,6 +56,8 @@ multiple qBittorrent instances with support for 10k+ torrents.`,
 func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVar(&cfgFile, "config", "", "config file (default is OS-specific: ~/.config/qui/config.toml or %APPDATA%\\qui\\config.toml)")
+	rootCmd.PersistentFlags().StringVar(&dbPath, "database", "", "database file path (default is next to config file)")
```
zze0s: I would rename at least the database key to `database-dir` or `data-dir`. `database` becomes ambiguous if there's ever another engine supported.

And config could become `config-dir` in that case as well. Leaves room for more than just setting it directly to the file.

It's a breaking change but it's such a limited number of users so far so I wouldn't care too much about that.

s0up4200: Will sort that out, thanks 👍

PR #34, opened 2025-08-11
fix: prevent database corruption from stale schema during migrations

s0up4200:
SQLite connection pooling during ALTER TABLE operations could cause 'no such table' errors when the schema changed. This fix ensures migrations run with a single connection to prevent stale schema views, then enables connection pooling for normal operations afterward.

The issue occurred when users updated instance credentials and the health check tried to update the database while connections had stale schema information from migration 003's ALTER TABLE operations.

Changes:
- Use single connection (MaxOpenConns=1) during migrations
- Enable connection pooling (MaxOpenConns=25) after migrations complete
- Set connection lifetime to 5 minutes for fresh schema views


zze0s (commented):
The migration needs to run in a transaction so it only commits if it's all OK.

We do this in all the other projects AFAIK, so I'd pick a solution and run with it.

EDIT: it does each single applyMigration in a transaction but I would run all the available ones in a single transaction.

PR #38, opened 2025-08-11
feat: migrate instances to single URL field

s0up4200:
## Breaking Change
Migrates instance configuration from separate host/port fields to a single URL field. Users will need to recreate their database.

## Changes
- Replace host/port columns with single URL field in database schema
- Add URL validation on backend (HTTP/HTTPS only)
- Improve frontend URL validation with IPv4/IPv6 support
- Default username to 'admin' (qBittorrent default)
- Update tests to use testify assertions

## Migration
This is a breaking change for early adopters. Users must delete their existing database and let it recreate with the new schema. BACKUP LICENSE KEYS!


internal/models/instance.go
```diff
@@ -99,7 +100,47 @@ func (s *InstanceStore) decrypt(ciphertext string) (string, error) {
...
+	} else {
+		// No scheme, add http://
+		rawURL = "http://" + rawURL
+	}
+	
+	// Parse the URL
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return "", err
+	}
```
zze0s: You can check for u.Schema after url.Parse, so no need for custom strings.Contains

internal/qbittorrent/client.go
```diff
@@ -23,44 +23,10 @@ type Client struct {
...
-			hostURL = fmt.Sprintf("%s:%d", host, port)
-		}
-	} else {
-		// Fallback to original behavior
-		hostURL = fmt.Sprintf("%s:%d", host, port)
-	}
-	
+func NewClient(instanceID int, instanceURL, username, password string, basicUsername, basicPassword *string) (*Client, error) {
+	// Clean up trailing slash if present
+	hostURL := strings.TrimSuffix(instanceURL, "/")
```
zze0s: Not needed, the client takes care of this with `url.JoinPath`

web/src/components/instances/InstanceForm.tsx
```diff
@@ -99,52 +97,58 @@ export function InstanceForm({ instance, onSuccess, onCancel }: InstanceFormProp
...
+              
+              // Require port for IP addresses (except localhost)
+              // This helps prevent common mistakes where users forget the port
+              if ((isIPv4 || isIPv6) && !isLocalhost && !url.port) {
+                return 'Port is required when using an IP address (e.g., :8080)'
+              }
+              
+              return undefined
+            } catch {
+              return 'Please enter a valid URL'
```
zze0s: You could utilize validation with Zod. It has a `ZodURL` type which should work here, I think. https://zod.dev/packages/core

PR #41, opened 2025-08-12
feat(torrents): redesign torrent table pane

martylukyy:


web/src/components/layout/Header.tsx
```diff
@@ -25,18 +35,140 @@ interface HeaderProps {
 
 export function Header({ children, sidebarCollapsed = false }: HeaderProps) {
   const { user, logout } = useAuth()
+  const location = useLocation()
+  const navigate = useNavigate()
+  const routeSearch = useSearch({ strict: false }) as any
+
+  const isInstanceRoute = useMemo(() => location.pathname.includes('/instances/'), [location.pathname])
```
martylukyy: If someone has a more elegant way for instance detection here feel free to suggest a refactor.

s0up4200 (commented):
I like it!

Pointers:
Should we move the all time stats and the current speed stat to the top bar? And those two can be made more compact as well I think. And we could remove the |separators between each stat and speed, since the chevrons are explanatory on their own. And to separate stats from current speed, maybe a - would look cleaner.

For speeds per torrent (mobile UI): We could move percentage to be inline with the progress bar, and then have the speeds use that space instead. (Or simply not show speeds per torrent on mobile)

s0up4200 (commented):
This should be moved to top bar as well.

Image

martylukyy:
> Should we move the all time stats and the current speed stat to the top bar? And those two can be made more compact as well I think. And we could remove the |separators between each stat and speed, since the chevrons are explanatory on their own. And to separate stats from current speed, maybe a - would look cleaner.

Personally, as long as we decide to keep the footer for the torrent table, i wouldn't move too much stuff into the header.
I removed the separators between the stats and speeds.
Since they spread out evenly anyway i didn't opt for a `-` as a separator for those.
Please speak your mind if you disagree.

> For speeds per torrent (mobile UI): We could move percentage to be inline with the progress bar, and then have the speeds use that space instead. (Or simply not show speeds per torrent on mobile)

I didn't touch that part, and to be honest i am not quite sure on what to do so that it will not overflow anymore...
For the idea with the percentage inline to the progress bar - would you move the ETA like this aswell in that case?
I am not a friend of not showing brr if you know what i mean 😄

s0up4200:
> Personally, as long as we decide to keep the footer for the torrent table, i wouldn't move too much stuff into the header. I removed the separators between the stats and speeds. Since they spread out evenly anyway i didn't opt for a `-` as a separator for those. Please speak your mind if you disagree.

No, I think I agree with you now.

> I didn't touch that part, and to be honest i am not quite sure on what to do so that it will not overflow anymore... For the idea with the percentage inline to the progress bar - would you move the ETA like this as well in that case? I am not a friend of not showing brr if you know what i mean 😄

I misread the initial question I think. Maybe we shouldn't show totals on mobile at all. We need decide what to show and what not to show on mobile because of limited space, and this is the perfect time 😄 I think we should drop total stats on mobile UI. That's my take at least.

For the cards, we could likely clean them up, but don't feel obliged to do it in this PR.

martylukyy:
I removed the all time stats from the mobile view now and centered the speed.
Let's consider the cards for mobile out of scope now and rework the overflows in a separate PR.
Here is a screenshot of the changes:

image

PR #46, opened 2025-08-12
fix: sort tags alphabetically case-insensitive

s0up4200:
## Summary
- Sort tags alphabetically using case-insensitive comparison for consistent UI ordering

## Before/After
- Before: Tags appeared in inconsistent order due to case-sensitive sorting
- After: Tags sorted alphabetically regardless of capitalization (e.g., "AITHER", "aither" group together)

### DIFF
--- a/internal/qbittorrent/sync_manager.go
+++ b/internal/qbittorrent/sync_manager.go
@@ -533,6 +533,10 @@ func (sm *SyncManager) GetTags(ctx context.Context, instanceID int) ([]string, e
 		return nil, fmt.Errorf("failed to get tags: %w", err)
 	}
 
+	sort.Slice(tags, func(i, j int) bool {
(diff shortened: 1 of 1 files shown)


internal/qbittorrent/sync_manager.go
```diff
@@ -533,6 +533,10 @@ func (sm *SyncManager) GetTags(ctx context.Context, instanceID int) ([]string, e
 		return nil, fmt.Errorf("failed to get tags: %w", err)
 	}
 
+	sort.Slice(tags, func(i, j int) bool {
+		return strings.ToLower(tags[i]) < strings.ToLower(tags[j])
+	})
+
```
nuxencs: should maybe use `slices.SortFunc` or `slices.SortStableFunc` instead 🤔

PR #90, opened 2025-08-25
fix(qbittorrent): stable sort order for torrents with equal values

### LINKED ISSUE #75: Shuffling ordered torrents
**Version**
0.2.0 latest

**Describe the bug**

On one of my added instances it seems to shuffle the order of torrents. I have sorted by "Date added", but it seems to be resorting every couple of seconds those torrents that were added in the same batch so all have the same "date added". See the video:

https://limewire.com/d/SiMI2#I6mXEDcplo

**Desktop (please complete the following information):**

Macos Firefox

s0up4200:
When multiple torrents have the same value for the sort field (e.g., same "added_on" timestamp), they would randomly shuffle on each request. This was causing the UI to visually shuffle torrents every 3 seconds when the frontend polled for updates.

The fix adds a hash-based tiebreaker to ensure deterministic, stable ordering when primary sort values are equal. This resolves the shuffling issue reported in #75.

Also improves string comparisons by using strings.EqualFold for case-insensitive equality checks.

Fixes #75

### DIFF
--- a/web/eslint.config.js
+++ b/web/eslint.config.js
@@ -7,7 +7,7 @@ import { globalIgnores } from 'eslint/config'
 import stylistic from '@stylistic/eslint-plugin'
 
 export default tseslint.config([
-  globalIgnores(['dist', '@web/pnpm-lock.yaml']),
+  globalIgnores(['dist', '@web/pnpm-lock.yaml', 'vite.config.ts']),
   {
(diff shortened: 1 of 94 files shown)


s0up4200 (commented):
Might be better to check the version instead of checking for error first.

PR #91, opened 2025-08-25
feat(web): add expandable dashboard all-time statistics

### LINKED ISSUE #69: Dashboard stats
On the dashboard, make the all time stats clickable so it expands to rows for each individual instance. Simple replace "All time statistics" with the instance name per row and leave the right side stats the same

s0up4200:
Implement feature request #69 to make all-time statistics clickable and expandable to show per-instance breakdown.

Features:
- Click all-time statistics bar to expand/collapse
- Table view showing individual instance stats
- Smooth animations with rotating chevron icon
- Hover effects and visual feedback
- Responsive design for mobile and desktop
- Consistent padding with header layout

Closes #69

CleanShot 2025-08-25 at 11 40 23

### DIFF
--- a/web/src/components/ui/table.tsx
+++ b/web/src/components/ui/table.tsx
@@ -1,0 +1,114 @@
+import * as React from "react"
+
+import { cn, } from "@/lib/utils"
+
+function Table({ className, ...props }: React.ComponentProps<"table">,) {
+  return (
+    <div
+      data-slot="table-container"
+      className="relative w-full overflow-x-auto"
+    >
+      <table
+        data-slot="table"
+        className={cn("w-full caption-bottom text-sm", className,)}
+        {...props}
+      />
+    </div>
+  )
+}
+
+function TableHeader({ className, ...props }: React.ComponentProps<"thead">,) {
+  return (
+    <thead
+      data-slot="table-header"
+      className={cn("[&_tr]:border-b", className,)}
+      {...props}
+    />
+  )
+}
+
+function TableBody({ className, ...props }: React.ComponentProps<"tbody">,) {
+  return (
+    <tbody
+      data-slot="table-body"
+      className={cn("[&_tr:last-child]:border-0", className,)}
+      {...props}
+    />
+  )
+}
+
+function TableFooter({ className, ...props }: React.ComponentProps<"tfoot">,) {
+  return (
+    <tfoot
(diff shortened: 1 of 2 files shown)


web/src/components/ui/table.tsx
martylukyy, line 1: Is there a reason we spin up our own table component instead of using the table lib we already use?

s0up4200: the one used for the torrenttable is a customized version. Using the original table from shadcn here is proper imo.

web/src/pages/Dashboard.tsx
```diff
@@ -444,72 +446,132 @@ function GlobalAllTimeStats({ statsData, }: { statsData: Array<{ instance: Insta
   }
 
   return (
-    <div className="rounded-lg border bg-card p-4">
-      {/* Mobile layout */}
-      <div className="sm:hidden">
-        <div className="flex items-center justify-between mb-3">
-          <h3 className="text-sm font-medium text-muted-foreground">All-Time Statistics</h3>
```
martylukyy, line 448: Chevron is missing in mobile view even though you can expand here.

martylukyy:
I would also center the row entries and get rid of the Peers columns since the peers are not really a all-time stat.
That way it look cleaner on mobile views...
Thoughts?

web/src/pages/Dashboard.tsx
s0up4200: Hm, where do you mean exactly?

CleanShot 2025-08-25 at 19 35 53@2x

s0up4200:
> I would also center the row entries and get rid of the Peers columns since the peers are not really a all-time stat. That way it look cleaner on mobile views... Thoughts?

We could rename it to "Server Statistics" for example. People wanted the Peers info so.

We could hide peers on mobile though, for sure.

s0up4200:
CleanShot 2025-08-25 at 21 10 49@2x
CleanShot 2025-08-25 at 21 11 00@2x
CleanShot 2025-08-25 at 21 11 04@2x

nuxencs:
the down arrow for download and the down arrow for the dropdown should be different imo

s0up4200:
> the down arrow for download and the down arrow for the dropdown should be different imo

Is this okay?

CleanShot 2025-08-26 at 11 55 35

PR #96, opened 2025-08-25
fix(lint): relax comma dangle rule

martylukyy:
We went a little overboard with the comma dangle rule which seems to have changed since the last time i used it.
It added trailing commas on all imports and functions which we do not need and want to have.
This PR adapts the rule in our eslint config and a format has been run.


s0up4200 (approved):
Nice 👍

PR #104, opened 2025-08-27
fix(docker): add curl for health check

s0up4200:
- Add curl to runtime dependencies in ci.Dockerfile
- Replace wget with curl -f in HEALTHCHECK command
- Fixes Docker reporting container as unhealthy


s0up4200:
Closing as wget is available. Issue reported was due to user error.

PR #122, opened 2025-08-31
feat(auth): migrate to SCS database-backed session storage

s0up4200:
- Replace gorilla/sessions with alexedwards/scs for better reliability
- Add SQLite3 session store implementation following autobrr pattern
- Create database migration for sessions table
- Update auth handlers to use SCS session manager
- Add "Remember me" functionality to login form
- Configure secure session cookies with 7-day lifetime
- Set custom cookie name to "qui_session"

### DIFF
--- a/cmd/qui/main.go
+++ b/cmd/qui/main.go
@@ -15,6 +15,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/alexedwards/scs/v2"
 	"github.com/go-chi/chi/v5"
 	"github.com/rs/zerolog"
 	"github.com/rs/zerolog/log"
@@ -31,6 +32,7 @@ import (
 	"github.com/autobrr/qui/internal/qbittorrent"
 	"github.com/autobrr/qui/internal/services"
 	"github.com/autobrr/qui/internal/web"
+	"github.com/autobrr/qui/pkg/sqlite3store"
 	webfs "github.com/autobrr/qui/web"
 )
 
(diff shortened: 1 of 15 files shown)


internal/auth/service.go
zze0s, line 1: I would skip adding the sessionManager in the auth service because the only thing it does is return it.

If it's not there then the cli handlers doesn't need it either.

And pass it explicitly to the auth middleware, auth handler and server deps.

cmd/qui/main.go
```diff
@@ -209,7 +211,11 @@ If no --config-dir is specified, uses the OS-specific default location:
 			}
 			defer db.Close()
 
-			authService := auth.NewService(db.Conn(), cfg.Config.SessionSecret)
+			// Create minimal session manager for CLI command
+			sessionManager := scs.New()
+			sessionManager.Store = sqlite3store.New(db.Conn())
```
zze0s: No longer needed.

cmd/qui/main.go
```diff
@@ -300,7 +306,11 @@ If no --config-dir is specified, uses the OS-specific default location:
...
```
zze0s: No longer needed.

cmd/qui/user_commands_test.go
```diff
@@ -107,7 +109,11 @@ func TestRunCreateUserCommand(t *testing.T) {
 				db, err := database.New(cfg.GetDatabasePath())
 				require.NoError(t, err)
 
-				authService := auth.NewService(db.Conn(), cfg.Config.SessionSecret)
+				// Create minimal session manager for test
+				sessionManager := scs.New()
+				sessionManager.Store = sqlite3store.New(db.Conn())
+
```
zze0s: No longer needed.

cmd/qui/user_commands_test.go
```diff
@@ -237,7 +243,11 @@ func TestRunChangePasswordCommand(t *testing.T) {
...
```
zze0s: No longer needed.

s0up4200:
Should be good now, sorry about the last round 😄 @zze0s

PR #123, opened 2025-08-31
feat(domain): add secret redaction helper functions

s0up4200:
Add RedactString and IsRedactedValue functions for masking sensitive data in API responses.

### DIFF
--- a/web/src/components/instances/InstanceForm.tsx
+++ b/web/src/components/instances/InstanceForm.tsx
@@ -74,11 +74,11 @@ export function InstanceForm({ instance, onSuccess, onCancel }: InstanceFormProp
         submitData = data
       }
     } else {
-      // Basic auth disabled - preserve existing basic auth if editing
+      // Basic auth disabled - clear basic auth credentials
       submitData = {
         ...data,
-        basicUsername: instance?.basicUsername ?? undefined,
-        basicPassword: undefined, // Don't send basicPassword to preserve existing
+        basicUsername: "",
+        basicPassword: "",
       }
     }
 


internal/api/handlers/auth.go
```diff
@@ -33,18 +34,53 @@ type SetupRequest struct {
...
+	type Alias SetupRequest
+	return json.Marshal(&struct {
+		*Alias
+		Password string `json:"password"`
+	}{
+		Password: domain.RedactString(s.Password),
+		Alias:    (*Alias)(&s),
+	})
+}
+
```
zze0s: The MarshalJSON method is used when we turn this INTO json to send via the api, and I think this only takes data in, unless it also responds with the same data this is not needed.

internal/api/handlers/auth.go
```diff
@@ -33,18 +34,53 @@ type SetupRequest struct {
...
+	type Alias LoginRequest
+	return json.Marshal(&struct {
+		*Alias
+		Password string `json:"password"`
+	}{
+		Password: domain.RedactString(l.Password),
+		Alias:    (*Alias)(&l),
+	})
+}
+
```
zze0s: See above. Keep if the LoginRequest is returned as is (but it should not return the password via the api on login).

s0up4200:
Should be good! @zze0s

s0up4200:
> Only minor change for the api docs, rest looks good.
> 
> The instance update method will however not allow to remove a set basic auth password. It likely won't be needed often and users can just add a new client without it, but if there's more state stored in the future it should be fixed to allow empty. Then it needs to fetch the existing instance first and compare before it sets it.

This should be fixed now. The UI displays a redacted placeholder for existing basic auth passwords and doesn't send an update for the password field unless the user modifies it. If the user tries to clear the field while basic auth is enabled, the field will be marked as required and won't allow submission with an empty value.

PR #131, opened 2025-09-02
fix(web): don't expand sidebar on dropdown click

nuxencs:
This removes the automatic unhiding of the sidebar when clicking on any items in the top right dropdown that are not an instance.

### DIFF
--- a/web/src/components/layout/Header.tsx
+++ b/web/src/components/layout/Header.tsx
@@ -6,11 +6,7 @@
 import { useAuth } from "@/hooks/useAuth"
 import { Button } from "@/components/ui/button"
 import { Input } from "@/components/ui/input"
-import {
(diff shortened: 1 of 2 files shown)


s0up4200:
This was actually a feature to reduce amount of clicks needed from going from torrenttable with closed sidebar to Dashboard and back etc. 😄 What do you think @martylukyy ?

nuxencs:
The clicks are the exact same going from the table to the dashboard and back 🤔
Top right menu -> dashboard -> click on instance name

s0up4200:
> The clicks are the exact same going from the table to the dashboard and back 🤔 Top right menu -> dashboard -> click on instance name

You make a solid argument there lol

s0up4200 (approved):

PR #185, opened 2025-09-11
fix(auth): correctly use `withBasePath` for login/setup check and redirect

cmd430:
fixes inability to login if using a base path

### DIFF
--- a/web/src/lib/api.ts
+++ b/web/src/lib/api.ts
@@ -31,7 +31,7 @@ class ApiClient {
     })
 
     if (!response.ok) {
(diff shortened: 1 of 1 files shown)


cmd430:
related but not in scope of this PR, the swizzin theme logo image has same issue

s0up4200:
> related but not in scope of this PR, the swizzin theme logo image has same issue

How? 🤔 We don't use that on the login/setup screens.

cmd430:
sorry should have clarified, once logged in, the one that replaces the autobrr bunny on top left

s0up4200:
I see it now, I'll commit that fix separately. Thanks for the report 👍

PR #196, opened 2025-09-13
fix: sync table view and selection state when no longer match

s0up4200:
- Update torrents list even when backend returns empty array
- Clear selection state when filtered torrents are removed from view
- Prevent stale torrent display after state changes (pause/resume)


s0up4200:
Superseded by #217

Issue #120, opened 2025-08-31
FR: Disable Auth


jtdroste:
**Is your feature request related to a problem? Please describe.**
Kinda? I'd prefer to not manage credentials within an application.

**Describe the solution you'd like**
I'd like the ability to disable authentication for QUI - essentially delegating it to my reverse proxy.

An alternative that's probably more annoying is OIDC ;)

**Describe alternatives you've considered**

**Additional context**

s0up4200:
Hey! For now we'll do OIDC, but might revisit the thought of bypassing auth for local network etc. in the future. 

I'll leave this FR open for that.

jfgoncalves:
I’m also interested in this, to shield it behind my Cloudflare Access and avoid a redundant double auth. If considered, using an environment variable could be envisaged. Or, some projects like Sonarr even require to edit a specific config file to make sure it’s not trivial for users to disable if that’s the concern.

ppfeister:
Created #298 to track OIDC as an RFE independent of disabling auth

PR #200, opened 2025-09-13
feat(auth): add oidc support

s0up4200:
Re #120 

Tested and works with:

- [x] https://github.com/pocket-id/pocket-id
- [x] https://github.com/tailscale/tailscale/tree/main/cmd/tsidp


internal/api/handlers/auth.go
```diff
@@ -260,14 +288,55 @@ func (h *AuthHandler) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	// For OIDC users, we might not have a user_id
+	userID := h.sessionManager.GetInt(r.Context(), "user_id")
+	authMethod := h.sessionManager.GetString(r.Context(), "auth_method")
+
+	response := map[string]any{
 		"username": username,
+	}
+
+	// Only include ID if it exists (for built-in auth users)
+	if userID != 0 {
```
KyleSanderson, line 300: This seems weird, shouldn't it be a number?

s0up4200: We only include the field if user_id from the session is non-zero which only happens for built-in users

PR #205, opened 2025-09-15
fix(filters): untagged & uncategorized not working

nuxencs:
This fixes the `Uncategorized` and `Untagged` filter options in the sidebar. They did not work on their own, only together with another category/tag.


s0up4200 (approved):
Perfect 👍

PR #223, opened 2025-09-21
test(database): harden migration contract

s0up4200:


KyleSanderson:
We really need to work on our sqlite paths in general across all of the projects. This is actually more correct than autobrr, but we need to really start using cached stmts where possible and a single writer with a reader pool.

internal/database/db.go
```diff
@@ -24,6 +25,50 @@ type DB struct {
...
+	}
+
+	for _, pragma := range pragmas {
+		if err := exec(ctx, pragma); err != nil {
+			return fmt.Errorf("apply connection pragma %q: %w", pragma, err)
+		}
+	}
+
+	return nil
+}
```
zze0s, line 70: Why do we want it done this way without the _ import vs what we do in every other codebase? @KyleSanderson

zze0s: Without the default import with _ to import side-effects an init functions it will have some sort of issue on arm64 apparently.

The init calls a func https://gitlab.com/cznic/sqlite/-/blob/master/sqlite.go?ref_type=heads#L60-L62

amd64
https://gitlab.com/cznic/sqlite/-/blob/master/lib/hooks.go?ref_type=heads#L18

arm64
https://gitlab.com/cznic/sqlite/-/blob/master/lib/hooks_linux_arm64.go?ref_type=heads#L23-L26

PR #226, opened 2025-09-21
fix(server): respect baseUrl when serving web UI

s0up4200:
Respect baseUrl when serving web UI
- mount the embedded frontend router beneath the configured baseUrl so `/qui/assets/...` resolves correctly
- add an explicit SPA root handler to serve `index.html` for both `/qui` and `/qui/`


zze0s (approved):
This is good for now! Will refactor the web handler at a later point 👍

PR #233, opened 2025-09-22
fix(build): use the correct ldflags for build info

saltydk:
Also changed the default dev version to something compatible with semver.

### DIFF
--- a/.air.toml
+++ b/.air.toml
@@ -5,7 +5,7 @@ tmp_dir = "tmp"
 [build]
   args_bin = ["serve"]
   bin = "./tmp/main"
-  cmd = "go build -ldflags \"-X main.Version=dev -X main.PolarOrgID=$POLAR_ORG_ID\" -o ./tmp/main ./cmd/qui"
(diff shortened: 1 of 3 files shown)


s0up4200:
Do you mind adding this as well? 👍

```patch
diff --git a/pkg/version/version.go b/pkg/version/version.go
index 058369d..2cb6302 100644
--- a/pkg/version/version.go
+++ b/pkg/version/version.go
@@ -8,6 +8,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"slices"
 	"strings"
 	"time"
 
@@ -165,12 +166,10 @@ func isDevelop(version string) bool {
 		return true
 	}
 
-	tags := []string{"dev", "develop", "master", "latest", ""}
-	for _, tag := range tags {
-		if version == tag {
-			return true
-		}
+	if strings.HasSuffix(version, "-dev") || strings.HasSuffix(version, "-develop") {
+		return true
 	}
 
-	return false
+	tags := []string{"dev", "develop", "master", "latest", ""}
+	return slices.Contains(tags, version)
 }
```

PR #243, opened 2025-09-22
feat(mobile): add compacted view options

Audionut:
- Adds an option to the filter panel (suggestions welcome) to show compact view (only available on mobile)
- Scrollable filename
- Basic elements only

Screenshot_20250922-215438 Screenshot_20250922-202020


KyleSanderson:
I'm not going to bikeshed you, this is already an excellent quality of life improvement.

Can you take it even further (another mode, past this one), and have a single line per item. If it's seeding you know it's 100% already, and if it's a % maybe show the state beside it (76% errored / downloading / seeding / etc.) tags and cats I don't know if they're needed for the incredibly dense view.

Audionut:
Category is useful here.

The current layout keeps it consistent between filters.

My only blocker to your suggestion, is how to option the further compact view. Suggestions?

KyleSanderson:
Triple toggle maybe? I think ultimately this will end up as a bitfield with people picking what they want is relevant to them, but it doesn't need to start there to land.

Audionut:
Added single line (ultra compact) view mode.

{scrollable_filename} {torrent.state} {percentage{if not 100%)} {speed(if applicable)}

Audionut:
Seems the view setting is not quite correct. the tags/category line is not being handled correctly in ultra mode.

Will fix tomorrow

Audionut:
Couldn't help myself, fixed.

Audionut:
- [x] speed then state
- [x] refactor the panel height code, since additional passes are causing minor display changes after initial display

s0up4200:
Hold on with this one. We might keep all 3.

KyleSanderson:
> Hold on with this one. We might keep all 3.

I think your first instinct was right, have 2.

Is there anything more meaningful we can show on 100% completed torrents then the size / size twice ( 54.2GB / 54.2GB ).

s0up4200:
This is great @Audionut 👍 Well done.

Let's merge this now, then decide if we want to keep all 3 modes later.

PR #249, opened 2025-09-22
fix(web): normalize torrent hash fallbacks

s0up4200:
Fix:
- centralize torrent hash normalization so v1/v2 fields fall back to the legacy hash when servers omit them
- reuse the helper in the details panel and context menu so copy + display work on qBittorrent 4.3.x without extra API calls

New:
- keep the peers query tab-gated while leaving existing reopen behavior untouched


s0up4200:
Tested and works for both v4.3.9 and v5+

PR #257, opened 2025-09-22
refactor(proxy): tighten reverse proxy request preparation

s0up4200:
Refactor proxy request preparation
- preflight proxy requests so we can fail fast with a 502 when prerequisites like instance lookup or pool acquisition fail
- reuse authenticated cookie jars and propagate stored HTTP Basic credentials to avoid 401s against protected instances
- normalize forwarded headers and centralize proxy error responses for cleaner downstream behavior

### DIFF
--- a/internal/proxy/handler.go
+++ b/internal/proxy/handler.go
@@ -4,6 +4,7 @@
 package proxy
 
 import (
+	"context"
 	"fmt"
 	"net/http"
 	"net/http/httputil"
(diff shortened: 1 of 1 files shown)


internal/proxy/handler.go
```diff
@@ -184,3 +205,84 @@ func (h *Handler) Routes(r chi.Router) {
...
+		Str("method", r.Method).
+		Str("path", r.URL.Path).
+		Logger()
+
+	if clientAPIKey != nil {
+		logger = logger.With().Str("client", clientAPIKey.ClientName).Logger()
+	}
+
+	if instanceID == 0 || clientAPIKey == nil {
+		logger.Error().Msg("Proxy request missing instance ID or client API key")
```
KyleSanderson: Too high of a log level for these, it's a DoS if someone just slams the endpoint.

s0up4200: Downgraded to warn, and another commit to try cover the DoS slams

PR #268, opened 2025-09-23
feat(web): clickable web links

Audionut:
- Make links on torrent comment fields, clickable.
- Uses a new util to convert plain text urls into react elements.


un2t:
thank you so much, guys.

PR #272, opened 2025-09-23
feat(web): add missing torrent table columns

nuxencs:
This adds support for almost all the missing torrent columns that are available in the qBit WebUI.

The only ones missing are currently:

- [ ] Incomplete Save Path // seems like it is not exposed by the qBit Web API
- [x] Reannounce In // not available in go-qbittorrent yet
- [x] Private // not available in go-qbittorrent yet

Needs https://github.com/autobrr/go-qbittorrent/pull/81 to be merged.


s0up4200:
Updated `DEFAULT_COLUMN_VISIBILITY` to match  https://github.com/qbittorrent/qBittorrent/blob/3de2a9f486a77a911fab986daabbe50ce7c85b04/src/webui/www/private/scripts/dynamicTable.js#L1104-L1151 for posterity.

nuxencs:
> Updated `DEFAULT_COLUMN_VISIBILITY` to match  https://github.com/qbittorrent/qBittorrent/blob/3de2a9f486a77a911fab986daabbe50ce7c85b04/src/webui/www/private/scripts/dynamicTable.js#L1104-L1151 for posterity.

makes sense, I didn't look hard enough to find that initially haha

PR #283, opened 2025-09-24
feat(web): report tracker error in torrent table

Audionut:
This actually works much quicker than I was expecting, with no noticeable delay.

Fetch torrent tracker status and report when trackers are in error.
Appropriate check for multiple tracker torrents.

Screenshot 2025-09-24 215114


Audionut:
Handled with https://github.com/autobrr/qui/pull/359

PR #316, opened 2025-09-26
feat(torrents): fetch tracker icons

s0up4200:
Fetch tracker icons
- add backend tracker icon service that fetches favicons, caches them on disk, and exposes `/api/tracker-icons/{tracker}`
- queue icon fetches from the sync manager whenever trackers change or sync
- render tracker icons in the torrent sidebar and torrent table (new icon column)


cmd/qui/main.go
```diff
@@ -475,7 +481,7 @@ func (app *Application) runServer() {
 	defer clientPool.Close()
 
 	// Initialize managers
-	syncManager := qbittorrent.NewSyncManager(clientPool)
+	syncManager := qbittorrent.NewSyncManager(clientPool, trackerIconService)
```
KyleSanderson: This isn't right, it should be a global thing.

internal/api/handlers/tracker_icons.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// NewTrackerIconHandler constructs a new handler for tracker icons.
+func NewTrackerIconHandler(service TrackerIconProvider) *TrackerIconHandler {
+	return &TrackerIconHandler{service: service}
+}
+
+// GetTrackerIcon resolves or fetches a tracker icon and streams it back to the client.
+func (h *TrackerIconHandler) GetTrackerIcon(w http.ResponseWriter, r *http.Request) {
+	tracker := strings.TrimSpace(chi.URLParam(r, "tracker"))
```
KyleSanderson: this should be all the icons (trackers, not tracker).

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,644 @@
...
+	} else {
+		svc.ua = "qui/dev"
+	}
+
+	return svc, nil
+}
+
+// QueueFetch schedules a background fetch for the given tracker host. trackerURL is optional
+// and may provide a more specific origin URL to discover icons from.
+func (s *Service) QueueFetch(host, trackerURL string) {
```
KyleSanderson: The API seems bonkers, there should be a single exposed element (GetIcon), fetching and similar are internal to the service.

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,644 @@
...
+	if err != nil {
+		return err
+	}
+
+	defer func() {
+		tmpFile.Close()
+		os.Remove(tmpFile.Name())
+	}()
+
+	if err := png.Encode(tmpFile, img); err != nil {
```
KyleSanderson: can't this be done in memory? it seems very unnecessary.

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,644 @@
...
+	s.failureMu.Unlock()
+}
+
+func (s *Service) clearFailure(host string) {
+	s.failureMu.Lock()
+	delete(s.lastFailure, host)
+	s.failureMu.Unlock()
+}
+
+func sanitizeHost(host string) string {
```
KyleSanderson, line 538: We aren't smart enough to do this. filepath.Clean filepath.Base and see if there's one that removes utf8 and similar stuff.

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,644 @@
...
+
+	decoded, err := base64.StdEncoding.DecodeString(payload)
+	if err != nil {
+		return nil, "", err
+	}
+
+	return decoded, contentType, nil
+}
+
+func decodeImage(data []byte, contentType, originalURL string) (image.Image, error) {
```
KyleSanderson, line 698: what on earth. there's got to be someone who does this safer and better than us.

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,644 @@
...
+			return nil, err
+		}
+		return ico.Decode(reader)
+	}
+
+	tryGeneric := func() (image.Image, error) {
+		if _, err := reader.Seek(0, io.SeekStart); err != nil {
+			return nil, err
+		}
+		img, _, err := image.Decode(reader)
```
KyleSanderson: https://pkg.go.dev/image#DecodeConfig

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,644 @@
...
+		return nil, fmt.Errorf("svg icons are not supported")
+	}
+
+	reader := bytes.NewReader(data)
+
+	tryICO := func() (image.Image, error) {
+		if _, err := reader.Seek(0, io.SeekStart); err != nil {
+			return nil, err
+		}
+		return ico.Decode(reader)
```
KyleSanderson: Completely unnecessary? https://github.com/mat/besticon/blob/v3.12.0/ico/ico.go#L259

cmd/qui/main.go
s0up4200: What's the benefit from moving it exactly?

KyleSanderson: it's effectively an image caching tool, it doesn't necessarily belong on sync.

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,733 @@
...
+	if img, _, err := image.Decode(reader); err == nil {
+		return img, nil
+	}
+
+	// Fallback to ICO
+	if _, err := reader.Seek(0, io.SeekStart); err != nil {
+		return nil, fmt.Errorf("failed to reset reader for ico decode: %w", err)
+	}
+
+	if cfg, err := ico.DecodeConfig(reader); err == nil {
```
KyleSanderson: what if there is an error? shouldn't we die here? also check for width and heigh being < 1.

internal/services/trackericons/service.go
```diff
@@ -679,15 +679,28 @@ func decodeImage(data []byte, contentType, originalURL string) (image.Image, err
...
+		default:
+			return nil
+		}
+	}
 
 	reader := bytes.NewReader(data)
 
 	// Check dimensions before expensive decode to avoid decompression bombs
+	// If validation succeeds here we can skip re-checking the same dimensions later.
+	validated := false
```
KyleSanderson, line 719: If the Error is Not NIL, we need to abort. This is not a valid image.

internal/services/trackericons/service.go
```diff
@@ -0,0 +1,790 @@
...
+		ctx, cancel := context.WithTimeout(context.Background(), fetchTimeout)
+		defer cancel()
+		if _, err := svc.GetIcon(ctx, h, tracker); err != nil {
+			// Intentionally ignore errors here; they are tracked internally for cooldown.
+		}
+	}(sanitized, trackerURL)
+}
+
+// ListIcons returns all cached tracker icons as base64-encoded data URLs.
+func (s *Service) ListIcons(ctx context.Context) (map[string]string, error) {
```
Audionut, line 155: If I could dump a list like this into a txt file (with acronym updated to tracker url), that would be nice.

https://github.com/Audionut/add-trackers/blob/8db05c0e822f9b3afa46ca784644c4e7e400c92b/ptp-add-filter-all-releases-anut.js#L768

Audionut: Most all tracker base64 icons are here
https://github.com/Purfview/IMDb-Scout-Mod/blob/15eb4a80ea076142d59632dc9ab740f17e597d72/IMDb_Scout_Mod.user.js#L2221

s0up4200: Done

PR #320, opened 2025-09-27
fix(torrents): context menu aligment and bar recheck icon

nuxencs:
This fixes the alignment in the torrent context menu and also changes the torrent management bar to use the same icon for rechecking as the context menu.

Before:
Screenshot 2025-09-27 at 10 42 37

After:
Screenshot 2025-09-27 at 10 42 12


s0up4200 (commented):
This is good, but should we put the Speed/Share Limits into a sub-menu while we're at it? Like we do for Queue.

nuxencs:
They are already submenus themselves

s0up4200 (approved):

PR #324, opened 2025-09-27
fix(torrents): preload limit dialogs with current values

s0up4200:
- preload share and speed limit dialogs with the selected torrents' current values and add a global limits toggle
- send qBittorrent's expected sentinel values when toggles are disabled, fixing the silent speed limit failure
- pass torrent snapshots into the dialogs, extend the torrent type definitions, and bump go-qbittorrent to expose the new fields

New global limit override:

PixelSnap 2025-09-27 at 22 27 54@2x

### DIFF
--- a/web/src/components/torrents/TorrentDialogs.tsx
+++ b/web/src/components/torrents/TorrentDialogs.tsx
@@ -3,29 +3,6 @@
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
 
-import { memo, useState, useEffect, useRef, useCallback } from "react"
-import type { ChangeEvent, KeyboardEvent } from "react"
-import { Button } from "@/components/ui/button"
-import { Input } from "@/components/ui/input"
-import { Label } from "@/components/ui/label"
-import { Checkbox } from "@/components/ui/checkbox"
-import { Switch } from "@/components/ui/switch"
-import { ScrollArea } from "@/components/ui/scroll-area"
-import {
-  Select,
-  SelectContent,
-  SelectItem,
-  SelectTrigger,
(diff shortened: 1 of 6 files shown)


web/src/components/torrents/TorrentDialogs.tsx
```diff
@@ -837,65 +837,168 @@ export const EditTrackerDialog = memo(function EditTrackerDialog({
...
+  inactive_seeding_time_limit?: number
+  max_ratio?: number
+  max_seeding_time?: number
+  max_inactive_seeding_time?: number
+}
+
+type SpeedLimitTorrentSnapshot = {
+  dl_limit?: number
+  up_limit?: number
+}
```
nuxencs: why define additional types if `Torrent` already has the exact same fields? 🤔

web/src/components/torrents/TorrentDialogs.tsx
```diff
@@ -837,65 +837,168 @@ export const EditTrackerDialog = memo(function EditTrackerDialog({
...
+  ratioLimit: number
+  seedingTimeEnabled: boolean
+  seedingTimeLimit: number
+  inactiveSeedingTimeEnabled: boolean
+  inactiveSeedingTimeLimit: number
+}
+
+const normalizeShareSignature = (torrent: ShareLimitTorrentSnapshot): string => {
+  const safe = (value: number | undefined, fallback: number) =>
+    typeof value === "number" ? value : fallback
```
nuxencs: could be moved out of `normalizeShareSignature` and reused across the other `const`s

nuxencs (approved):

PR #334, opened 2025-09-28
fix(torrents): show ratio limit correctly

nuxencs:
This changes how the ratio limit gets shown in the table and correctly displays the global limit if required.


s0up4200 (commented):
Should we check the enable flag before rendering the value? That way we’d still show something like `2.00` when the global limit is active, but fall back to `∞` when it’s switched off. 

`max_ratio_enabled`

nuxencs:
I think I have all cases cover, but feel free to prove me wrong haha

PR #337, opened 2025-09-28
fix(qbittorrent): enforce exact tag filtering

### LINKED ISSUE #323: selecting tags that are part of other tags will filter both.
**Describe the bug**
one tag is 'bc' the other one is 'abcd' 
selecting 'bc' ins the filter tab will also show 'abcd' torrents in the list 

**Expected behavior**
only 'bc' shows up if you select 'bc' in the filter tab 

**Additional context**
i think the cause of this is that the qbit api endpoint returns a string of tags per torrent 'bc,abcd,abcde' and you have to manually split and turn it into an array.

s0up4200:
- force an exact match when filtering by a single tag so selecting `bc` no longer leaks torrents tagged only `abcd`
- add targeted unit tests to lock in the regression scenario and cover multi-tag torrents that legitimately include `bc`
- add a benchmark to track any future changes to tag filtering performance

Closes #323

### DIFF
--- a/internal/qbittorrent/sync_manager.go
+++ b/internal/qbittorrent/sync_manager.go
@@ -242,6 +242,12 @@ func (sm *SyncManager) GetTorrentsWithFilters(ctx context.Context, instanceID in
 
 		// Use library filtering and sorting
 		filteredTorrents = syncManager.GetTorrents(torrentFilterOptions)
+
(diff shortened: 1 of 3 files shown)


s0up4200:
Will be superseded by https://github.com/autobrr/go-qbittorrent/pull/83

PR #338, opened 2025-09-28
feat: webhooks

xpkmb:
# Summary

This feature gives qui the ability to send certain webhooks whenever qbitorrent starts/completes a torrent download.

image


Audionut:
It would be super nice if this had some other conditions for trigger.

Personally, `tags` and `category` match would be useful. Just an idea.

xpkmb:
@Audionut currently qui only refreshes when the user is actively looking at the UI. So for the moment we are limited to external actions like qBitorrent sending a request to qui whenever a new torrent is added/completed.

The Webhooks will contain all relevant information though, like `category` and `tags`.

PR #341, opened 2025-09-29
fix(web): client advanced settings limits

Audionut:
Fixes client settings to only show client specific settings
Fixes all settings to respect client bounds
Updates the validation handlers
Adds mouseover help

Checking Memory Use: https://github.com/qbittorrent/qBittorrent/blob/eed0e56d1ab200070baeae5caac7935f3634a8b0/src/webui/www/private/views/preferences.html#L3094

Application specifics: https://github.com/qbittorrent/qBittorrent/blob/eed0e56d1ab200070baeae5caac7935f3634a8b0/src/webui/www/private/views/preferences.html#L3194

TODO:

- [x] err, actually get application information
- [x] double check all application checks
- [x] automagic the panel layout based on available settings
- [x] Show `0`, not empty


s0up4200:
https://github.com/autobrr/go-qbittorrent/blob/main/methods.go#L75 this should have what you need, but its missing zlib and platform in the struct currently.

Audionut:
This still needs panel layout handling.

I'm thinking a two-step solution:

1. make all inputs the same size (one-third width)
2. move the variable (can hide) inputs to the end of their respective sections.

I'll have to double check, but there's probably some section that has > 3 can_hide inputs, which could leave dangling inputs on a new row without existing row = 3 inputs.

s0up4200:
I'd like z/kyle to look over the backend bit before we merge this. Ideally we should do a single call for whatever the Dashboard needs. Right now there's at least 3 calls firing here.

[2068c09](https://github.com/autobrr/qui/pull/341/commits/2068c09a2bf7014ae49c762a43d93c777f4cdb9b) exposes qbit and webapi version on the instance cards
CleanShot 2025-10-04 at 23 31 00@2x

Audionut:
Compressed connection settings panel:

Screenshot 2025-10-05 173547 Screenshot 2025-10-05 173408

Audionut:
Screenshot 2025-10-06 203528

Requires https://github.com/autobrr/go-qbittorrent/pull/88

Audionut:
Advanced settings as per instance above:

Screenshot 2025-10-07 204725

s0up4200 (approved):
🚀

PR #347, opened 2025-09-29
feat(web): add torrent table density toggle

### LINKED ISSUE #339: FR: Compact view for desktop users
Would be great to have a much more compacted view for mouse and keyboard warriors.
At the moment, qui appears to be targeting touchscreens, which looks great, but a lot of screen space is wasted on a classic monitor. qui needs almost double the screen size to display the same amount of info.

Below is the comparison between qui and qbit

Image

s0up4200:
This is for desktop

- Add toolbar toggle to cycle between normal, compact, and ultra-compact density modes
- Persist view mode preference across sessions with localStorage
- Optimize render performance with pre-computed className lookups
- Move row height constants to module scope to eliminate useMemo overhead

### DIFF
--- a/web/src/components/torrents/DraggableTableHeader.tsx
+++ b/web/src/components/torrents/DraggableTableHeader.tsx
@@ -3,17 +3,34 @@
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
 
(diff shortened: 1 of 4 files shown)


s0up4200:
Superseded by #643

PR #351, opened 2025-09-30
feat(torrents): create torrents

s0up4200:
Adds native Torrent Creator
Works for qBittorrent v5+

TODO:

- [x] Wire up creator button properly in the UI
- [x] Wire up the Tasks button properly in the UI

Targets [autobrr/go-qbittorrent@`9150109` (#85)](https://github.com/autobrr/go-qbittorrent/pull/85/commits/9150109e3685c1a296a3ce9002f890002bf2e380)

Screenshots:
PixelSnap 2025-09-30 at 11 19 13@2x

CleanShot 2025-09-30 at 11 20 24@2x

### DIFF
--- a/web/src/components/torrents/TorrentCreatorDialog.tsx
+++ b/web/src/components/torrents/TorrentCreatorDialog.tsx
@@ -32,9 +32,9 @@ import { Tooltip, TooltipContent, TooltipTrigger } from "@/components/ui/tooltip
 import { api } from "@/lib/api"
 import type { TorrentCreationParams, TorrentFormat } from "@/types"
 import { useForm } from "@tanstack/react-form"
-import { useMutation, useQueryClient } from "@tanstack/react-query"
+import { useMutation, useQuery, useQueryClient } from "@tanstack/react-query"
 import { AlertCircle, ChevronDown, Info, Loader2 } from "lucide-react"
-import { useState } from "react"
+import { useEffect, useState } from "react"
 import { toast } from "sonner"
 import { pieceSizeOptions, TorrentPieceSize } from "./piece-size"
 
@@ -48,6 +48,13 @@ export function TorrentCreatorDialog({ instanceId, open, onOpenChange }: Torrent
   const [error, setError] = useState<string | null>(null)
   const [advancedOpen, setAdvancedOpen] = useState(false)
   const queryClient = useQueryClient()
+
+  // Fetch active trackers for the select dropdown
+  const { data: activeTrackers } = useQuery({
+    queryKey: ["active-trackers", instanceId],
+    queryFn: () => api.getActiveTrackers(instanceId),
+    enabled: open, // Only fetch when dialog is open
+  })
 
   const mutation = useMutation({
     mutationFn: async (data: TorrentCreationParams) => {
@@ -114,6 +121,15 @@ export function TorrentCreatorDialog({ instanceId, open, onOpenChange }: Torrent
       mutation.mutate(params)
     },
   })
+
+  // Reset form and error state when dialog closes
+  useEffect(() => {
+    if (!open) {
+      form.reset()
+      setError(null)
+      setAdvancedOpen(false)
+    }
+  }, [open, form])
 
   return (
     <Dialog open={open} onOpenChange={onOpenChange}>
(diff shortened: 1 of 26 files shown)


web/src/components/torrents/TorrentCreatorDialog.tsx
```diff
@@ -0,0 +1,366 @@
...
+              {/* Piece Size */}
+              <form.Field name="pieceSize">
+                {(field) => (
+                  <div className="space-y-2">
+                    <Label htmlFor="pieceSize">Piece Size (bytes)</Label>
+                    <Input
+                      id="pieceSize"
+                      type="number"
+                      placeholder="Auto (leave empty)"
+                      value={field.state.value}
```
Audionut: Prohibit invalid values. Ideally, the steppers would only cycle through valid values. 

Probably needs a hardcoded list. There's _not that many_ valid values, and they're static also.

s0up4200: > Leave empty for auto (recommended). Common: 16384, 32768, 65536, 131072, 262144

These are taken from upstream, and we can hardcode something for those. I'll double check that I got them right though.

web/src/components/torrents/TorrentCreatorDialog.tsx
```diff
@@ -0,0 +1,366 @@
...
+                    </p>
+                  </div>
+                )}
+              </form.Field>
+
+              {/* URL Seeds */}
+              <form.Field name="urlSeeds">
+                {(field) => (
+                  <div className="space-y-2">
+                    <Label htmlFor="urlSeeds">Web Seeds (HTTP/HTTPS)</Label>
```
Audionut, line 330: Would be pretty fancy to have a drop down list of current announce urls in the instances.

web/src/components/torrents/TorrentCreatorDialog.tsx
```diff
@@ -0,0 +1,366 @@
...
+            )}
+          </form.Field>
+
+          {/* Trackers */}
+          <form.Field name="trackers">
+            {(field) => (
+              <div className="space-y-2">
+                <Label htmlFor="trackers">Trackers</Label>
+                <p className="text-sm text-muted-foreground">
+                  One tracker URL per line
```
Audionut, line 241: Would be pretty fancy to have a drop down list of current announce urls in the instances.

web/src/components/torrents/TorrentCreatorDialog.tsx
Audionut: A drop down list (that's more friendly than bytes) would be nice here also.

web/src/components/torrents/TorrentCreatorDialog.tsx
```diff
@@ -0,0 +1,366 @@
...
+          {/* Source Path */}
+          <form.Field name="sourcePath">
+            {(field) => (
+              <div className="space-y-2">
+                <Label htmlFor="sourcePath">
+                  Source Path <span className="text-destructive">*</span>
+                </Label>
+                <Input
+                  id="sourcePath"
+                  placeholder="/path/to/file/or/folder"
```
Audionut, line 168: Since linux users get the love of a valid example, there should be some note for windows users about double backslash requirement.

If https://github.com/autobrr/go-qbittorrent/pull/84/files becomes a thing, it would be even nicer to just handle paths OS dependently.

web/src/components/torrents/TorrentCreatorDialog.tsx
s0up4200: Done

web/src/components/torrents/TorrentCreatorDialog.tsx
s0up4200: Lets revisit this one in a separate PR

web/src/components/torrents/TorrentCreatorDialog.tsx
```diff
@@ -0,0 +1,394 @@
...
+                        <SelectItem value="262144">256 KiB</SelectItem>
+                        <SelectItem value="524288">512 KiB</SelectItem>
+                        <SelectItem value="1048576">1 MiB</SelectItem>
+                        <SelectItem value="2097152">2 MiB</SelectItem>
+                        <SelectItem value="4194304">4 MiB</SelectItem>
+                        <SelectItem value="8388608">8 MiB</SelectItem>
+                        <SelectItem value="16777216">16 MiB</SelectItem>
+                        <SelectItem value="33554432">32 MiB</SelectItem>
+                        <SelectItem value="67108864">64 MiB</SelectItem>
+                        <SelectItem value="134217728">128 MiB</SelectItem>
```
nuxencs: maybe add an enum for these, would be easier to reuse in the future

Audionut:
A `QOL ` improvement could be to validate the torrent file save path before creation, because someone kept inputting only a path when testing. :P

s0up4200:
> A `QOL ` improvement could be to validate the torrent file save path before creation, because someone kept inputting only a path when testing. :P

I checked qBittorrent code and couldnt find anything about it validating the save path, the only time you'll get a propagated error is when you try to download the torrent file.

Audionut:
> > A `QOL ` improvement could be to validate the torrent file save path before creation, because someone kept inputting only a path when testing. :P
> 
> I checked qBittorrent code and couldnt find anything about it validating the save path, the only time you'll get a propagated error is when you try to download the torrent file.

Yeah, I keep forgetting that we're working with remote machines. The example is explicit, and people like me should just read properly.

The tracker drop down is super nice :)

PR #359, opened 2025-10-02
feat(torrents): introduce Unregistered and TrackerDown

s0up4200:


internal/qbittorrent/sync_manager.go
```diff
@@ -654,6 +779,115 @@ func (sm *SyncManager) primaryTrackerDomain(torrent qbt.Torrent) string {
...
+			return true
+		}
+	}
+
+	return false
+}
+
+func (sm *SyncManager) torrentTrackerIsDown(torrent qbt.Torrent) bool {
+	for _, tracker := range torrent.Trackers {
+		if tracker.Status != qbt.TrackerStatusNotWorking {
```
Audionut: Reverse the logic?

If any tracker has a working status, return false, since the torrent has a working tracker.

s0up4200: Good call 👍

Audionut:
Found an edge case. Unregistered status needs to be linked to a working tracker.

Screenshot 2025-10-05 174904

KyleSanderson:
> Found an edge case. Unregistered status needs to be linked to a working tracker.
> 
> Screenshot 2025-10-05 174904

If a single tracker continues to work, it's not unregistered.

Audionut:
> If a single tracker continues to work, it's not unregistered.

Yeah, that's the edge case, it is marked unregistered ;)

s0up4200:
> > If a single tracker continues to work, it's not unregistered.
> 
> Yeah, that's the edge case, it is marked unregistered ;)

Pushed a fix for that

s0up4200 (commented):
We should add a grace period for `unregistered`, like ignore torrents that were added within the latest hour.

selfishraccoon:
This feature works fine in one qbit 5.1.2 instance but doesn't in the one i upgraded on ultra (was 4.5.5, updated to 5.1.2-BETA, the newest version supported by their upgrade script as of now)
as an aside the torrent creation menu also doesn't show up in the ultra instance

I assume there may be api changes in the beta version although the tool I was using to tag the various tracker codes continues to work still.

Let me know if I should open an issue or can help with more investigation from here
Thanks!

s0up4200:
> This feature works fine in one qbit 5.1.2 instance but doesn't in the one i upgraded on ultra (was 4.5.5, updated to 5.1.2-BETA, the newest version supported by their upgrade script as of now)
> 
> as an aside the torrent creation menu also doesn't show up in the ultra instance
> 
> 
> 
> I assume there may be api changes in the beta version although the tool I was using to tag the various tracker codes continues to work still.
> 
> 
> 
> Let me know if I should open an issue or can help with more investigation from here
> 
> Thanks!

Try to restart qui and optionally clear cookies and cache for qui in the browser

selfishraccoon:
as soon as i restarted my docker container it worked where the refreshes before didn't! 
thanks so much!

E: i think i got now why it didnt though: i created the container before the ultra update, i should've thought of that

PR #363, opened 2025-10-04
fix(web): correct stepping to -1 when step is a decimal value

Audionut:
When step is a decimal value like `step="0.05"`, html builtin input behavior prevents stepping correctly from `0` to `-1` and skipping intermediate values such as `-0.3`.

This results in these input boxes being unable to reset to default `-1`, aka `unlimited` state (without passing through invalid values).

This fix tracks the input value state, to force the correct state change.

Also update `max_ratio_enabled` to `0.05` steps to mimic client behavior.

### DIFF
--- a/web/src/components/forms/NumberInputWithUnlimited.tsx
+++ b/web/src/components/forms/NumberInputWithUnlimited.tsx
@@ -3,6 +3,7 @@
  * SPDX-License-Identifier: GPL-2.0-or-later
  */
 
+import React from "react"
 import { Input } from "@/components/ui/input"
 import { Label } from "@/components/ui/label"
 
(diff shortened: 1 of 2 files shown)


web/src/components/forms/NumberInputWithUnlimited.tsx
```diff
@@ -37,6 +38,36 @@ export function NumberInputWithUnlimited({
...
+  React.useEffect(() => {
+    if (!allowUnlimited) return
+    
+    const prevValue = prevValueRef.current
+    const currentValue = value
+    
+    // If we stepped down to exactly 0 from a positive value, transition to unlimited
+    if (prevValue > 0 && currentValue === 0) {
+      // Small delay to let the UI update, then transition to unlimited
+      setTimeout(() => onChange(-1), 50)
```
s0up4200: Is the delay necessary?

Audionut: No, not only is it pointless, it impacts clean state change.

s0up4200 (commented):
Am I supposed to be able to go from `0` to `-1` with this? The only way I can do that is by typing 1 and putting a `-` before it. 🤔

KyleSanderson:
> Am I supposed to be able to go from `0` to `-1` with this? The only way I can do that is by typing 1 and putting a `-` before it. 🤔
> 
> And wouldnt simply toggle it to `off` be enough?

Can't we do better, if it's empty in the webui, it's the default. No weird -1 nonsense.

PR #367, opened 2025-10-04
feat(torrents): column filtering

nuxencs:
Adds column filtering support for pretty much all available columns for finegrained filtering.

Screenshot 2025-10-04 at 16 24 17
Screenshot 2025-10-04 at 16 24 51


nuxencs:
Ah yeah private is the only bool column, will fix that 😄
default width adjustments seem good 👍🏼

internal/qbittorrent/sync_manager.go
```diff
@@ -1367,6 +1371,15 @@ func (sm *SyncManager) applyManualFilters(client *Client, torrents []qbt.Torrent
 		}
 	}
 
+	var program *vm.Program
+	var compileErr error
+	if len(filters.Expr) > 0 {
+		program, compileErr = expr.Compile(filters.Expr, expr.Env(qbt.Torrent{}), expr.AsBool())
```
KyleSanderson: can you use ttlcache here?

nuxencs: was thinking about it, didn't try it yet, will do that tomorrow

PR #369, opened 2025-10-04
perf(web): improve torrent table scroll performance on Chromium browsers

### LINKED ISSUE #368: Poor torrent table scrolling performance in Chromium-based browsers
**Version**

Noticed using docker tag `latest`, but the issue exists on `main` as well.

**Describe the bug**

Scrolling performance is not ideal in Chromium-based browsers (at least, in Chrome and Edge which I was able to test in) in Windows. On other browsers like Safari on an iPad, this issue doesn't seem to persist. The number of torrents does not seem to matter, even if less than the default desktop "page" size of 300.

**To Reproduce**

1. Install Chrome or Edge (or presumably any other Chromium based browsers).
2. Open the torrent table of any instance which has at least a reasonable number of items to scroll through.
3. Observe the laggy scrolling performance.

**Expected behavior**

Scrolling should be equally smooth across all browsers and platforms.

**Desktop (please complete the following information):**
 - OS: Windows 11
 - Browser: Chrome / Edge
 - Version: Chrome 141 / Edge 141

To be clear on this alternative device scrolling behaviour is as expected.

**Smartphone (please complete the following information):**
 - Device: iPad Pro 11" 2nd Generation
 - OS: iOS 26.1 Beta
 - Browser: Safari (effectively all browsers)
 - Version: User agent reports as Safari/604.1

corrosiv-e:
Fixes #368.

Tested in Chrome and Edge and seems to resolve the described issue. Does not seem to introduce any unexpected changes in Firefox on the same machine, or on Safari on iOS.

### DIFF
--- a/web/src/components/torrents/TorrentTableOptimized.tsx
+++ b/web/src/components/torrents/TorrentTableOptimized.tsx
@@ -1093,7 +1093,7 @@ export const TorrentTableOptimized = memo(function TorrentTableOptimized({ insta
       {/* Table container */}
       <div className="flex flex-col flex-1 min-h-0 mt-2 sm:mt-0 overflow-hidden">
         <div
-          className="relative flex-1 overflow-auto scrollbar-thin select-none"
+          className="relative flex-1 overflow-auto scrollbar-thin select-none will-change-transform"
           ref={parentRef}
           role="grid"
           aria-label="Torrents table"


s0up4200:
Thanks for the PR! Will need to look more into this. This fix seems like a "last-resort" thing according to https://developer.mozilla.org/en-US/docs/Web/CSS/will-change

I see "choppiness" in Safari on my macbook, but both Firefox and Chrome seems fine.

Are you using a high refresh rate monitor btw?

corrosiv-e:
Yes, this is on a high refresh rate display.

Yeah, it's not super ideal to resolve the issue this way, just what happened to work for me. I took the idea from this thread that's gotten quite a few comments about the issue: https://github.com/TanStack/virtual/issues/860

I saw some people reporting `contain: paint` also helped, but I didn't end up trying that, not sure if it's preferable.

s0up4200:
Thanks, this is useful info 👍

s0up4200:
Just adding this patch here for future reference as it seems to improve it for Safari as well: https://0.0g.gg/?7f1fef2b4def1b32#9i4aHcpPt2CzFPUdXxmhHpSZbCWdADiRoUXTHVPoqW3J

NOTE: Using the trackpad to scroll is already smooth in Safari on main, but this seemed to help when using a scroll wheel.

PR #385, opened 2025-10-06
feat(torrents): add backup/restore support

s0up4200:
#### Testing
- [x] Backups
- [x] Restore

### DIFF
--- a/internal/api/server_test.go
+++ b/internal/api/server_test.go
@@ -27,6 +27,7 @@ import (
 	"github.com/autobrr/qui/internal/qbittorrent"
 	"github.com/autobrr/qui/internal/services/license"
 	"github.com/autobrr/qui/internal/services/trackericons"
+	"github.com/autobrr/qui/internal/update"
 	"github.com/autobrr/qui/internal/web"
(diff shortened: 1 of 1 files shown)


s0up4200:
This should be good to go for a review if anyone feels brave

internal/api/server.go
```diff
@@ -191,6 +194,10 @@ func (s *Server) Handler() (*chi.Mux, error) {
 	clientAPIKeysHandler := handlers.NewClientAPIKeysHandler(s.clientAPIKeyStore, s.instanceStore)
 	versionHandler := handlers.NewVersionHandler(s.updateService)
 	qbittorrentInfoHandler := handlers.NewQBittorrentInfoHandler(s.clientPool)
+	var backupsHandler *handlers.BackupsHandler
+	if s.backupService != nil {
+		backupsHandler = handlers.NewBackupsHandler(s.backupService)
+	}
```
zze0s: Is there any case where the api won't have some of these methods like BackupsHandler and TrackerIconHandler? The nil checks doesn't really make sense since these should always be available

s0up4200: Yeah these are not needed.

PR #397, opened 2025-10-07
fix(web): mobile tooltips

Audionut:
mobile tooltip


s0up4200 (approved):
Brought back the `children` prop and the arrow. LGTM 👍

IMG_1942

PR #406, opened 2025-10-09
feat(qbittorrent): derive uptime from log buffer

s0up4200:
This change adds qBittorrent session uptime to the dashboard.

- Captures the start timestamp by polling the qBittorrent log endpoint on startup, and whenever the cached value looks stale, taking the earliest log entry. qbit clears its logs on restart, so that first entry should mark session start.
- We use the cached value until the RID counter suggests a restart
- If fetching logs fail, we mark the cache inaccurate and do a five-minute backoff, and leave the uptime fields empty so the UI shows N/A.

CleanShot 2025-10-09 at 12 52 29@2x


s0up4200:
Will be revisited in a simpler way when qbit 5.2 lands

ireun:
5.2 landed ;)

PR #434, opened 2025-10-12
feat(web): display IPv4/IPv6 address in status bar; blur in incognito mode

m04ndj5xc2u3pn4eow:
The external IPv4 (and presumably IPv6) address is currently displayed on the dashboard. However, adding it to the status bar would provide easier access without the need to navigate back to the dashboard each time.

**Screenshots (IP address edited for obvious reasons)**

Non-incognito mode:
ipv4_1

Incognito mode:
ipv4_2


m04ndj5xc2u3pn4eow:
Ah, yeah, that makes sense on taking up real estate on the status bar. Will make the updates 👍

PR #451, opened 2025-10-13
feat(web): drag-and-drop add torrents

s0up4200:
CleanShot 2025-10-13 at 22 19 56@2x

### DIFF
--- a/web/src/components/torrents/AddTorrentDialog.tsx
+++ b/web/src/components/torrents/AddTorrentDialog.tsx
@@ -41,7 +41,7 @@ import { api } from "@/lib/api"
 import type { Torrent } from "@/types"
 import { useForm } from "@tanstack/react-form"
 import { useMutation, useQueryClient } from "@tanstack/react-query"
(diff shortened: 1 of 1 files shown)


Audionut:
Suggestion: Search for torrent in instance.

Currently, you can add a torrent that is already in the instance, and there is no feedback.
I guess ideally it would search for the torrent immediately, overlaying an error if the `Add New Torrent` dialog is still open, or showing a message if the user clicks `add torrent` before the existing search is complete.

nuxencs:
> showing a message if the user clicks `add torrent` before the existing search is complete.

I think that would be a nice way to check for that 👍🏼

s0up4200:
> Suggestion: Search for torrent in instance.
> 
> Currently, you can add a torrent that is already in the instance, and there is no feedback. I guess ideally it would search for the torrent immediately, overlaying an error if the `Add New Torrent` dialog is still open, or showing a message if the user clicks `add torrent` before the existing search is complete.



> > showing a message if the user clicks `add torrent` before the existing search is complete.
> 
> I think that would be a nice way to check for that 👍🏼

Done!

PR #452, opened 2025-10-13
fix(proxy): add base-path support for proxy routing

s0up4200:
Adds support for custom qui baseurl for the client proxy feature.

Introduces a new helper for basePaths that we can port /api etc to later. Not touching that for this.

- [x] Tested and works with Swizzin

CleanShot 2025-10-13 at 23 08 23@2x


s0up4200:
Tested and works with Swizzin

PR #456, opened 2025-10-14
feat(proxy): implement shared transport

KyleSanderson:


internal/proxy/handler.go
```diff
@@ -74,6 +75,7 @@ func NewHandler(clientPool *qbittorrent.ClientPool, clientAPIKeyStore *models.Cl
 		ModifyResponse: h.modifyResponse,
 		BufferPool:     bufferPool,
 		ErrorHandler:   h.errorHandler,
+		Transport:		sharedhttp.Transport,
```
nuxencs: indenting 😛

PR #489, opened 2025-10-20
feat(sync): periodic instance syncing

Audionut:
Implements a per-instance configuration to keep running `autosync` at a periodic (maximum) interval.

The periodic timer resets when any other action is performed in the syncmanager, ensuring that the interval functions correctly, as a maximum autosync interval feature, without interfering with any other actions.

Runs on application start to ensure torrent tables have access to fresh cache on first load.
syncing


s0up4200 (approved):
lgtm

Audionut:
stale

PR #551, opened 2025-11-02
feat(sse): stream torrent and activity updates, replacing background polling

s0up4200:
Replaces the frontend's background polling with a Server-Sent Events stream. A single `/stream` endpoint multiplexes per-instance and cross-instance torrent updates alongside qui's own activity events (cross-seed, dirscan, jackett, orphan/tracker scans, reannounce, backups), so the torrents table, dashboard, title-bar speeds, and instance metadata stay live without a polling loop. Polling stays in place as an automatic fallback whenever the stream drops, and the UI surfaces connection state with retry/backoff.

On the backend, updates are coalesced and snapshots shared across subscribers behind a go-sse provider, with cached app-info/preferences and a prefer-cached query option to avoid redundant qBittorrent calls. Reverse-proxy guidance for SSE (buffering, timeouts, compression) is documented under `documentation/docs/advanced`.


Audionut (dismissed):

KyleSanderson:
Nice one

PR #571, opened 2025-11-09
fix(database): do not release mutex on tx err

KyleSanderson:
## Summary by CodeRabbit

* **Bug Fixes**
  * Writer locks are now reliably released and statement promotion to cache happens only after successful commits; rollbacks never promote statements and always release locks to avoid stalled transactions.

* **Tests**
  * Added concurrency tests covering commit/rollback locking behavior, double-rollback safety, serialization between transactions, and read-only concurrency.

* **Other**
  * No public API or exported signatures changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,15 +10,19 @@ A fast, modern web interface for qBittorrent. Supports managing multiple qBittor
 
 - [Features](#features)
 - [Installation](#installation)
+- [Docker](#docker)
+- [Updating](#updating)
 - [Configuration](#configuration)
-- [API](#api)
-- [Metrics](#metrics)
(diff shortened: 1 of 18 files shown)


internal/database/db.go
```diff
@@ -199,18 +200,20 @@ func (t *Tx) QueryRowContext(ctx context.Context, query string, args ...any) *sq
 // Commit commits the transaction and releases the writer mutex if this is a write transaction.
 // Also promotes any transaction-prepared statements to the DB cache for future use.
 func (t *Tx) Commit() error {
+	defer func() {
```
KyleSanderson: This is wrong, on a failed Commit it doesn't end the transaction.

PR #599, opened 2025-11-17
fix(backups): respect schedule when backups keep failing

s0up4200:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved backup missed detection logic to more reliably determine backup cadence, including when no successful runs exist.
  * Enhanced time calculation handling to consistently extract reference times from backup runs and gracefully fall back to the most recent run when necessary.


s0up4200:
Superseded by #632

PR #606, opened 2025-11-19
feat(torrents): add reannounce monitor

s0up4200:
Introduces a proactive reannounce service for qBittorrent instances to handle "unregistered" torrents and transient tracker outages without flooding trackers.

- Service: Background worker that scans new torrents, checks tracker status against known error patterns, and issues conservative reannounce calls (max 3 retries, configurable intervals).
- Scope: Supports "Monitor All" or targeted monitoring via Category, Tag, and Tracker allowlists/blocklists.
- Proxy: Intercepts `/api/v2/torrents/reannounce` to debounce external requests (e.g., from autobrr) for monitored torrents.
- Config: Per-instance settings for wait times, max age, and scope filters stored in `instance_reannounce_settings`.
- UI: Added reannounce configuration tab to Instance Preferences.

## Summary by CodeRabbit

* **New Features**
  * Tracker Reannounce: per-instance monitoring and conservative automatic reannounces with timing, scope, aggressive/cooldown controls and manual debounced requests.

* **UI**
  * Reannounce tab in Instance Preferences (settings + activity), Instance card indicator, instance form defaults, optional settings button, quick access from torrent table.
  * New Command palette and Multi-select UI components.

* **API**
  * Read endpoints for reannounce activity and current candidates.

* **Documentation**
  * New REANNOUNCE.md guide.

* **Tests**
  * Comprehensive tests for reannounce logic, caching and debounce behavior.

### DIFF
--- a/web/src/components/instances/preferences/TrackerReannounceForm.tsx
+++ b/web/src/components/instances/preferences/TrackerReannounceForm.tsx
@@ -21,7 +21,7 @@ import { cn, copyTextToClipboard } from "@/lib/utils"
 import type { InstanceFormData, InstanceReannounceActivity, InstanceReannounceSettings } from "@/types"
 import { useQuery } from "@tanstack/react-query"
 import { Copy, Info, RefreshCcw } from "lucide-react"
-import { useEffect, useMemo, useState } from "react"
+import { useMemo, useState } from "react"
 import { toast } from "sonner"
 
 interface TrackerReannounceFormProps {
@@ -120,10 +120,6 @@ export function TrackerReannounceForm({ instanceId, onSuccess }: TrackerReannoun
       }
     })
   }
-
-  useEffect(() => {
-    setSettings(cloneSettings(instance?.reannounceSettings))
-  }, [instance?.reannounceSettings, instanceId])
 
   const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
     event.preventDefault()
@@ -311,7 +307,7 @@ export function TrackerReannounceForm({ instanceId, onSuccess }: TrackerReannoun
                           id="monitor-all"
                           checked={settings.monitorAll}
                           onCheckedChange={(v) => {
-                            setSettings(prev => {
+                            setSettings((prev) => {
                               const next = { ...prev, monitorAll: v }
                               // Automatically switch to exclude mode if monitoring all
                               if (v) {
@@ -333,7 +329,7 @@ export function TrackerReannounceForm({ instanceId, onSuccess }: TrackerReannoun
                             <Label htmlFor="scope-categories">Categories</Label>
                             <Tabs
                               value={settings.excludeCategories ? "exclude" : "include"}
-                              onValueChange={(v) => setSettings(prev => ({ ...prev, excludeCategories: v === "exclude" }))}
+                              onValueChange={(v) => setSettings((prev) => ({ ...prev, excludeCategories: v === "exclude" }))}
                               className="h-7"
                             >
                               <TabsList className="h-7">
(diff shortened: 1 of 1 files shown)


internal/services/reannounce/REANNOUNCE.md
```diff
@@ -0,0 +1,67 @@
...
+   - Useful for filtering out public trackers, specific labels, or categories that don't need monitoring.
+7. Save the instance. Settings are remembered per instance.
+
+## What happens once it’s enabled?
+
+- qui watches the instance’s sync data for torrents whose trackers report:
+  - “Unregistered”‑style messages (matched against built‑in patterns).
+  - Known outage phrases (“tracker is down”, “maintenance”, etc.).
+- Only torrents that match your scope (and are not excluded), have a detected tracker problem, are still within the configured max active‑time window, and have no working trackers will be considered for reannounce.
+- For problematic torrents whose trackers are still in an “updating / not contacted yet” state, qui waits up to the configured **Initial tracker wait** to let qBittorrent finish its first announce cycle. If the tracker becomes healthy during this window, no reannounce is issued.
```
buroa: This shouldn't wait forever (if it does). I set a small timeout in qbrr on the wait, like 15 seconds. If you don't... they may sit for a while in that updating status even though they can be announced.

s0up4200: It already handles this with a configurable timeout. If the tracker is still in an updating state after that, it stops waiting and proceeds to reannounce. This is adjustable in the UI 👍 It defaults to 15 seconds.

CleanShot 2025-11-19 at 14 57 21@2x

buroa: LGTM!

internal/database/migrations/016_add_instance_reannounce_settings.sql
```diff
@@ -0,0 +1,28 @@
...
+    updated_at DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP
+);
+
+CREATE TRIGGER IF NOT EXISTS trg_instance_reannounce_settings_updated
+AFTER UPDATE ON instance_reannounce_settings
+BEGIN
+    UPDATE instance_reannounce_settings
+    SET updated_at = CURRENT_TIMESTAMP
+    WHERE instance_id = NEW.instance_id;
+END;
```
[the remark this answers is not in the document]
s0up4200: In our SQLite setup we don’t enable recursive triggers (PRAGMA recursive_triggers is never turned on), and this AFTER UPDATE pattern (issuing a second UPDATE to bump updated_at) is used consistently across other migrations too. In this configuration the inner UPDATE does not re‑fire the trigger, so there’s no recursion. To keep the migrations consistent, I’d prefer to leave this as‑is; if we ever switch to recursive_triggers = ON we’d adjust all of the updated_at triggers in a dedicated follow‑up.

ppkhoa:
Can the field be validated *after* user finished entering in these fields?

image

Not sure how other people do it, but when I change the value, I would delete the current value then enter new value there. However, current behavior prevents me from doing that. For example:
1) Tried to delete the value to enter another valid value (in this case, changing max torrent age from `600` to `300`)
2) The box default to minimum acceptable value (I assume so, `60` in the case of max torrent age)
3) Had to enter `6300`, then delete the `6` after.
Similar behavior can be observed with the other 2 boxes in the screenshot above.

ppkhoa:
I'm not clear on what `Aggressive Mode` does. If it's turned off, wait 2 minutes if re-announce failed/did not change the torrent status (from `Aggressive Mode` description), then where does `Retry Interval` come in? Retry infers previous attempts failed. If `Aggressive Mode` retry immediately, does that mean it will ignore `Retry Interval`?

s0up4200:
> Can the field be validated _after_ user finished entering in these fields?
> image
> 
> Not sure how other people do it, but when I change the value, I would delete the current value then enter new value there. However, current behavior prevents me from doing that. For example:
> 
>     1. Tried to delete the value to enter another valid value (in this case, changing max torrent age from `600` to `300`)
> 
>     2. The box default to minimum acceptable value (I assume so, `60` in the case of max torrent age)
> 
>     3. Had to enter `6300`, then delete the `6` after.
>        Similar behavior can be observed with the other 2 boxes in the screenshot above.

Will get that sorted.

Will also clarify the meaning of aggressive mode in REANNOUNCE.md and the UI.

PR #661, opened 2025-12-01
feat(config): add SESSION_SECRET_FILE env var

undefined-landmark:
autobrr has the option to read all environment variables from a file: https://github.com/autobrr/autobrr#docker-secrets. This makes it easy to pass secrets from docker or a systemd credential. This was introduced in https://github.com/autobrr/autobrr/pull/2061/files. 

I had a go at implementing this in feature in qui as well. For now only for `SESSION_SECRET`. The same logic could be applied to other env vars if convenient.

Let me know if there are any nicer ways to implement this in go/viper. Thanks!

## Summary by CodeRabbit

* **New Features**
  * Support for providing the session secret via a file path using a new environment variable; file-based secret takes precedence over the direct secret env var.
* **Documentation**
  * Configuration docs updated to describe the file-based session secret option and precedence rules.
* **Tests**
  * Added tests validating file-vs-env precedence and correct secret loading (including trimming).

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -625,16 +625,25 @@ func (c *AppConfig) GetEncryptionKey() []byte {
 	return padded
 }
 
-// Sets viper variable if environment variable with _FILE suffix is present
+// bindOrReadFromFile sets the viper variable from a file if the _FILE suffixed
+// environment variable is present, otherwise it binds to the regular env var.
 func (c *AppConfig) bindOrReadFromFile(viperVar string, envVar string) {
 	envVarFile := envVar + "_FILE"
-	if filePath := os.Getenv(envVarFile); filePath != "" {
-		content, err := os.ReadFile(filePath)
-		if err != nil {
-			log.Fatal().Err(err).Str("path", filePath).Msg("Could not read " + envVarFile)
-		}
-		c.viper.Set(viperVar, strings.TrimSpace(string(content)))
-	} else {
+	filePath := os.Getenv(envVarFile)
+	if filePath == "" {
 		c.viper.BindEnv(viperVar, envVar)
+		return
 	}
+
+	content, err := os.ReadFile(filePath)
+	if err != nil {
+		log.Fatal().Err(err).Str("path", filePath).Msg("Could not read " + envVarFile)
+	}
+
+	secret := strings.TrimSpace(string(content))
+	if secret == "" {
+		log.Fatal().Str("path", filePath).Msg(envVarFile + " file is empty")
+	}
+
+	c.viper.Set(viperVar, secret)
 }
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -231,54 +231,66 @@ func TestNewLoadsConfigFromFileOrDirectory(t *testing.T) {
 }
 
 func TestBindOrReadFromFile(t *testing.T) {
-	tmpKeyFile := func(t *testing.T, tmpDir string) (string) {
-				configPath := filepath.Join(tmpDir, "key-file.txt")
-				content := "key-from-file"
-				require.NoError(t, os.WriteFile(configPath, []byte(content), 0o644))
-				return configPath
+	tmpKeyFile := func(t *testing.T, tmpDir string) string {
+		configPath := filepath.Join(tmpDir, "key-file.txt")
+		content := "key-from-file"
+		require.NoError(t, os.WriteFile(configPath, []byte(content), 0o644))
+		return configPath
 	}
 
-	noTmpKeyFile := func(t *testing.T, tmpDir string) (string) {
-				return ""
+	tmpKeyFileWithNewline := func(t *testing.T, tmpDir string) string {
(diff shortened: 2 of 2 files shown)


s0up4200 (commented):
## PR Review Summary

> **Note:** This review was generated by Claude Code (AI assistant).

---

## Critical Issues (2 found)

### 1. `ReadFileIfPath` is NOT called at runtime - only during config generation
**File:** `internal/config/config.go:370`

The `ReadFileIfPath` function is only called within `writeDefaultConfig()`. When a user sets `QUI__SESSION_SECRET_FILE=/run/secrets/session`, the **literal path string** is stored as the session secret, not the file contents.

**Impact:** The encryption key derived in `GetEncryptionKey()` will use the path string `/run/secrets/session` instead of the actual secret, completely breaking the feature.

**Fix:** Add secret resolution after config unmarshaling:
```go
// After viper.Unmarshal in New()
c.Config.SessionSecret = ReadFileIfPath(c.Config.SessionSecret)
```

### 2. Silent failure returns empty string for security-critical config
**File:** `internal/config/config.go:633-636`

```go
if err != nil {
    log.Error().Err(err).Msg("Could not read file")
    return ""  // Empty string → 32 null-byte encryption key!
}
```

When file read fails, returning empty string causes `GetEncryptionKey()` to create a **predictable null-byte key**, making all encrypted passwords trivially decryptable.

**Comparison with autobrr:** autobrr uses `log.Fatalf()` to crash on read errors - the correct approach for security-critical configuration.

**Fix:** Fail hard on read errors:
```go
log.Fatal().Err(err).Str("path", value).Msg("Could not read secret file")
```

---

## Important Issues (3 found)

### 3. Path collision: secrets that match filesystem paths
**File:** `internal/config/config.go:628-640`

Using `os.Stat()` to auto-detect paths creates ambiguity. If a user's secret string happens to match an existing file (e.g., `dev`, `tmp`), the file contents are read instead.

**Comparison with autobrr:** autobrr uses explicit `_FILE` suffix check first, avoiding this ambiguity entirely.

**Fix:** Handle `SESSION_SECRET_FILE` explicitly before viper binding, don't auto-detect.

### 4. viper.BindEnv misuse - doesn't work as fallback chain
**File:** `internal/config/config.go:182`

```go
c.viper.BindEnv("sessionSecret", envPrefix+"SESSION_SECRET_FILE", envPrefix+"SESSION_SECRET")
```

Viper binds all env vars to the key but doesn't implement file-reading logic. The path string is passed through as-is.

**Fix:** Check `SESSION_SECRET_FILE` env var explicitly before viper binding.

### 5. TOCTOU race condition
**File:** `internal/config/config.go:628-632`

There's a window between `os.Stat` and `os.ReadFile` where the file could change. Remove the stat check and handle `os.IsNotExist` from ReadFile directly.

---

## Test Coverage Gaps

| Missing Test Case | Risk |
|-------------------|------|
| Empty string input | Could cause security issues |
| Directory path (stat succeeds, read fails) | Silent empty secret |
| File with only whitespace | Empty secret after trim |
| Viper precedence integration test | Feature contract undocumented |

---

## Recommended Approach

Consider aligning with autobrr's implementation pattern (see `autobrr/internal/config/config.go` lines 465-475):

```go
func (c *AppConfig) loadFromEnv() {
    // Handle SESSION_SECRET_FILE explicitly first
    if filePath := os.Getenv(envPrefix + "SESSION_SECRET_FILE"); filePath != "" {
        content, err := os.ReadFile(filePath)
        if err != nil {
            log.Fatal().Err(err).Str("path", filePath).Msg("Could not read SESSION_SECRET_FILE")
        }
        c.viper.Set("sessionSecret", strings.TrimSpace(string(content)))
    } else {
        c.viper.BindEnv("sessionSecret", envPrefix+"SESSION_SECRET")
    }
    // ... rest of loadFromEnv
}
```

This approach:
- Explicitly checks `_FILE` variant first (no auto-detection ambiguity)
- Fails hard if file is specified but unreadable
- Sets the resolved value directly in viper
- Works at config load time, not just config generation

---

## Strengths

- Great feature idea that improves Docker/systemd secrets support
- Clean table-driven test structure  
- Documentation updated in README

undefined-landmark:
Thanks for the review. I think I have the implementation down now. Will write some tests and push the changes soon.

undefined-landmark:
Pushed some new changes, mainly implementing the logic from your review @s0up4200. Appreciate any new input.

While implementing I read that viper follows the following logic:

> Viper uses the following precedence for merging:
> 
>     explicit call to Set
>     flags
>     environment variables
>     config files
>     external key/value stores
>     defaults

With this PR we do break this a bit, as the _FILE env var will override flags, if I understand things correctly. Any ideas how to go about this in a better way? 

As far as I can tell there is no flag for SESSION_SECRET, so it is not a blocker in that regard.

undefined-landmark:
Thanks for taking the time to review! Feel free to ping me if anything needs to be fixed/extended wrt this change.

PR #662, opened 2025-12-01
fix(qbittorrent): cache tracker health counts in background

KyleSanderson:
## Summary by CodeRabbit

* **New Features**
  * Per-instance background tracker-health caching with start/stop and read APIs; UI now surfaces tracker health from the cache.
  * Client lifecycle automatically starts/stops background health refresh when instances are created, removed, or closed.

* **Refactor**
  * Tracker enrichment is gated by client capability and moved to background refreshes to reduce request latency.
  * Reuses pre-fetched tracker data and cache to avoid redundant hydration and improve performance.

### DIFF
--- a/internal/qbittorrent/pool.go
+++ b/internal/qbittorrent/pool.go
@@ -260,11 +260,12 @@ func (cp *ClientPool) createClientWithTimeout(ctx context.Context, instanceID in
 		// Don't fail client creation for sync manager issues
 	}
 
-	// Start background tracker health refresh if SyncManager is set
+	// Start background tracker health refresh if SyncManager is set and pool isn't closed
 	cp.mu.RLock()
 	sm := cp.syncManager
+	closed := cp.closed
 	cp.mu.RUnlock()
-	if sm != nil {
+	if sm != nil && !closed {
 		sm.StartTrackerHealthRefresh(instanceID)
 	}
 


s0up4200:
### Code review

Found 2 issues:

1. **Race condition between client creation and removal can cause goroutine leak** - In `createClientWithTimeout`, the pool lock is released (line 266) before calling `StartTrackerHealthRefresh` (line 268). Concurrently, `RemoveClient` can execute `StopTrackerHealthRefresh` between these lines. This allows `StartTrackerHealthRefresh` to be called after `StopTrackerHealthRefresh` for the same instance, creating a goroutine that runs indefinitely for a removed client.

https://github.com/autobrr/qui/blob/98a31843a3bac3d453dbe9202bac7f69b009175a/internal/qbittorrent/pool.go#L263-L269

2. **Goroutine leak when ClientPool is closed** - The `Close()` method iterates through clients and deletes them but never calls `StopTrackerHealthRefresh()` for each instance. This means all background tracker health refresh goroutines continue running after pool shutdown. Compare with `RemoveClient()` which correctly stops the refresh loop.

https://github.com/autobrr/qui/blob/98a31843a3bac3d453dbe9202bac7f69b009175a/internal/qbittorrent/pool.go#L374-L377

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

Issue #676, opened 2025-12-02
[FR] Add "Free space" to bottom details bar of the instance page


ewenjo:
Show the free space left on an instance on its dedicated page, only thing I'm missing from qbit webui.

bolach:
+1

ewenjo:
@s0up4200 @finevan  Can the `serverState.free_space_on_disk > 0` checks be removed (from both places)? It's useful to see that the disk is full instead of just hiding it.

finevan:
This matches the behavior on the dashboard page for consistency. I don’t know the history of the original component, but if the behavior is to be changed, the two should match.

ewenjo:
Figured you just matched the behavior yeah, so lets wait for Soup. I don't see any reason to hide it as `0` is a valid value from the API.

likecreep:
Can we also show this on mobile page?  Like mobile web page and mobile  PWA.

PR #691, opened 2025-12-04
feat(web): Torrent list details bar shows free space

finevan:
This addresses #676

Screenshot 2025-12-04 at 1 37 07 PM

## Summary by CodeRabbit

* **New Features**
  * Added a free disk space indicator in the torrent table header/footer showing available storage with a hard drive icon and a "Free Space" tooltip.
* **Bug Fixes**
  * External IPv4/IPv6 addresses now reflect the computed server state so displayed addresses match the current server status.

### DIFF
--- a/web/src/components/torrents/TorrentTableOptimized.tsx
+++ b/web/src/components/torrents/TorrentTableOptimized.tsx
@@ -2805,19 +2805,6 @@ export const TorrentTableOptimized = memo(function TorrentTableOptimized({
           </div>
 
           <div className="flex flex-wrap items-center justify-end gap-2 text-xs">
-            {effectiveServerState?.free_space_on_disk !== undefined && effectiveServerState.free_space_on_disk > 0 && (
-              <div className="flex items-center gap-2 pr-2 border-r last:border-r-0 last:pr-0">
-                <Tooltip>
-                  <TooltipTrigger asChild>
-                    <span className="flex items-center h-6 px-2 text-xs text-muted-foreground">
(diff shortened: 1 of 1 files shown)


web/src/components/torrents/TorrentTableOptimized.tsx
```diff
@@ -2804,6 +2805,19 @@ export const TorrentTableOptimized = memo(function TorrentTableOptimized({
           </div>
 
           <div className="flex flex-wrap items-center justify-end gap-2 text-xs">
+            {serverState?.free_space_on_disk !== undefined && serverState.free_space_on_disk > 0 && (
```
s0up4200: See if you can make use of `effectiveServerState` instead of calling `serverState` directly. `effectiveServerState` is a memoized version that handles instance changes and provides cached values during loading states. The existing code at lines 2916-2921 does use `serverState` directly for IP addresses, so there's inconsistency in the existing codebase too, feel free to fix the IP as well with this if you want.

finevan: Makes sense 👍
I've updated both free space and the ip address components, they seem to work locally

s0up4200:
CleanShot 2025-12-04 at 22 28 43@2x
Should we place it here instead? 🤔 That way it wont get pushed around by varying upload/dowload speeds 👍

finevan:
I like it -- I'm more of a backend dev and admittedly don't have much UI sense :-)

Updated!

PR #837, opened 2025-12-22
Fix/premium theme resets to default after idle time

thesecretlifeofabunny:
## Front-End: 
- Removed theme change on API error, kept fallback themes for safety like on login.
- Increased refetch from 30s to 1hour to reduce cpu cycles usage on theme validation.
- Created const for time variable instead of comments.
- Added return type to ThemeValidator.tsx

## Back-End:
- Renamed API function to reflect function usage.

## Comments:
If this approach isn't apt. My other thought was to keep track of how many isError counts occurred in a row and reset back to default after 5 or so occurrences in a row. (To keep original intention of those who implemented the license checks and outcomes). Otherwise I'm open to other suggestions on approaches that I should take or investigate.

## Summary by CodeRabbit

* **Improvements**
  * Simplified theme validation so users retain access to all themes when license data is temporarily unavailable.
  * Increased and standardized license caching and refresh intervals (less frequent checks, more consistent timing).
  * Made periodic background validation timing clearer and more maintainable.

* **Bug Fixes**
  * Stored premium themes are automatically reverted to a default when premium access is lost.


s0up4200:
HI there! Thanks for the PR! I was actually working on this locally myself and just pushed that to #844 which also has some performance improvements to loading/switching themes.

PR #851, opened 2025-12-23
fix(ui): incomplete torrents aren't "Completed: 1969-12-31"

finevan:
## Before
Before a download has completed, the UI displays an incorrect completion timestamp just prior to the unix epoch in both the torrent column view, as well as the torrent info panel due to a `-1` sentinel value from qbittorrent.
Screenshot 2025-12-23 at 3 50 55 PM

## After
Screenshot 2025-12-23 at 3 50 33 PM

-----
I considered updating the model to make the field nullable, however this element is piped through directly from qbittorrent and there's not enough reason to change it.

QBittorrent uses a sentinel value of `-1` when no completion date is available.

## Summary by CodeRabbit

* **Bug Fixes**
  * Completed date handling improved — list and details views now treat “-1” (and other missing values) as “no completion” and only show a completed timestamp when a valid completion date exists, preventing misleading or incorrect completion info from appearing.

### DIFF
--- a/web/src/components/torrents/TorrentDetailsPanel.tsx
+++ b/web/src/components/torrents/TorrentDetailsPanel.tsx
@@ -999,10 +999,12 @@ export const TorrentDetailsPanel = memo(function TorrentDetailsPanel({ instanceI
                               <p className="text-xs text-muted-foreground">Added</p>
                               <p className="text-sm">{formatTimestamp(properties.addition_date)}</p>
                             </div>
-                            <div className="space-y-1">
-                              <p className="text-xs text-muted-foreground">Completed</p>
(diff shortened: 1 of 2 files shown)


web/src/components/torrents/TorrentTableColumns.tsx
```diff
@@ -814,7 +814,7 @@ export const createColumns = (
     header: "Completed On",
     cell: ({ row }) => {
       const completionOn = row.original.completion_on
-      if (!completionOn || completionOn === 0) {
+      if (!completionOn || completionOn === -1) {
```
finevan, line 817: note `!completionOn` already handles the zero case. One could argue zero is a valid date in the general case, however the bittorrent protocol was first released in ~2001 so we're good.

PR #858, opened 2025-12-25
feat(docker): add PUID/PGID support

s0up4200:
## Description

This adds opt-in `PUID`/`PGID` support to the Docker image. When both variables are set, the entrypoint creates a matching user and group, fixes ownership of any non-matching files in `/config`, and drops privileges using `su-exec` before starting qui. If they are not set, the container continues to run as root.

The documentation now includes a Permissions section that covers `user:` as the standard Docker approach and `PUID`/`PGID` as the alternative for automatically repairing `/config` ownership, along with guidance for a one-time data-directory chown when migrating from root. Since #2086, `UMASK` is handled directly by the qui binary, so the entrypoint leaves it untouched.

## How has this been tested?

I tested the entrypoint in `alpine:3.23` containers across several scenarios: running as root by default, dropping privileges via `PUID`/`PGID`, fail-fast input validation, handling a correctly-owned `/config` containing root-owned files, handling dangling and outward-pointing symlinks, and handling an existing uid. CI does not build or run the image.

I also ran end-to-end tests with the built image against a live qBittorrent container and a synthetic `/data` volume. Running as uid 1000, qui repaired a root-owned `0600` file inside `/config` on startup, connected with `hasLocalFilesystemAccess`, and located and deleted an orphan file from `/data` as the unprivileged user. Testing within the same mounts verified that `UMASK=002` creates group-writable directories, cross-directory hardlinks work within `/data`, and a root-owned base directory blocks link creation as described in the documentation.

## Checklist


H2OKing89:
This would be $\color{#008000}{\underline{\textbf{VERY!}}}$ helpful. This would fix the folders from being owned by `root` on unraid using  **hardlink cross-seed** <3

DaCeige:
+1.  Looking for this feature also.

dolokolo:
+1, also waiting for this to be merged.

PR #474, opened 2025-10-18
feat(metrics): add all time metrics

rare-magma:
This adds all time metrics to the prometheus metrics exposed by qui:
```
# HELP qbittorrent_alltime_download_bytes Total downloaded data in bytes by instance
# TYPE qbittorrent_alltime_download_bytes counter
qbittorrent_alltime_download_bytes{instance_id="1",instance_name="t"} 10
# HELP qbittorrent_alltime_upload_bytes Total uploaded data in bytes by instance
# TYPE qbittorrent_alltime_upload_bytes counter
qbittorrent_alltime_upload_bytes{instance_id="1",instance_name="t"} 10

...

# HELP qbittorrent_session_download_bytes Total downloaded data in bytes per session by instance
# TYPE qbittorrent_session_download_bytes counter
qbittorrent_session_download_bytes{instance_id="1",instance_name="t"} 10
# HELP qbittorrent_session_upload_bytes Total uploaded data in bytes per session by instance
# TYPE qbittorrent_session_upload_bytes counter
qbittorrent_session_upload_bytes{instance_id="1",instance_name="t"} 10
```

Related issue: https://github.com/autobrr/qui/issues/424

PR #892, opened 2025-12-30
feat(metrics)!: Remove redundant transfer-rate prometheus metrics

finevan:
This may be somewhat controversial, so I'll do my best to explain my rationale:

The existing prometheus metrics which describe per-session total bytes uploaded and downloaded are substantially more accurate.

_Why?_

The transfer rate provided by qbit is a moving average over _some_ duration, and that duration isn't tied in any way to the configurable polling rate that the prometheus metrics are scraped. At best this can be derrived from existing metrics (as the value is quite literally the derivative of the transferred bytes graph). It's possible to read exclusively peaks or valleys, and fail to get a sense of actual transfer rate.

Prometheus expects such data to be provided as counters (as qui already exposes as per-session uploaded and downloaded bytes), so that the timeseries database may calculate a rate. This guarantees all bytes transferred are taken into account, effectively matching the aformentioned sliding window to either the prometheus polling interval, or a larger interval requested in prometheus queries.

Given this, the rate metric is something of a foot-gun, and we might as well reduce the periodic API requests from prometheus scrapes to slightly reduce server load.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed instance speed metrics from collection.


finevan:
The speed metrics I'd like to remove were included initially with the introduction of prometheus support in #54, and the upload/download metrics for both the current session and all-time were contributed a couple months later by @rare-magma in #474.

s0up4200:
If we decide to merge this, `GetInstanceSpeeds()` becomes dead code and can be removed.

finevan:
Good call, I've removed it

PR #914, opened 2025-12-31
feat(crossseed): add custom category prefix toggle

jabloink:
let me know 👍

## Release Notes

* **New Features**
  * Added a new configuration option to Cross Seed automation settings that allows prefixing the original category when using custom categories. When enabled, categories are formatted as "{custom_category}/{original_category}", providing more granular organization.

* **Chores**
  * Updated database schema, API definitions, and web interface components to support the new category prefixing configuration.

### DIFF
--- a/internal/api/handlers/crossseed.go
+++ b/internal/api/handlers/crossseed.go
@@ -43,6 +43,7 @@ type automationSettingsRequest struct {
 	UseCrossCategorySuffix       bool     `json:"useCrossCategorySuffix"`
 	UseCustomCategory            bool     `json:"useCustomCategory"`
 	CustomCategory               string   `json:"customCategory"`
+	PrefixOriginalCategory       bool     `json:"prefixOriginalCategory"`
 	RunExternalProgramID         *int     `json:"runExternalProgramId"`
 	SkipRecheck                  bool     `json:"skipRecheck"`
 }
@@ -72,6 +73,7 @@ type automationSettingsPatchRequest struct {
 	UseCrossCategorySuffix         *bool       `json:"useCrossCategorySuffix,omitempty"`
 	UseCustomCategory              *bool       `json:"useCustomCategory,omitempty"`
 	CustomCategory                 *string     `json:"customCategory,omitempty"`
+	PrefixOriginalCategory         *bool       `json:"prefixOriginalCategory,omitempty"`
 	RunExternalProgramID           optionalInt `json:"runExternalProgramId"`
 	// Source-specific tagging
 	RSSAutomationTags    *[]string `json:"rssAutomationTags,omitempty"`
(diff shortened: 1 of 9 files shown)


Barcode-eng (commented):
Seems like a reasonable addition to me! PR is outdated though, and will need docs and migration number update.

A thought that occurred to me – does it make sense to combine this with the existing "Add .cross category suffix" option? So, user can decide if they want to add a custom category as a prefix or suffix. In the case of "Add .cross category suffix" users, it would set a custom category of ".cross" and set it to suffix. It _might_ be too big of a lift for this PR, but it would make this customizable enough to cover most use-cases.

Anyways, my second paragraph is more thinking out loud, and I don't think that has to block this PR.

jabloink:
> Seems like a reasonable addition to me! PR is outdated though, and will need docs and migration number update.
> 
> A thought that occurred to me –- does it make sense to combine this with the existing "Add .cross category suffix" option? So, user can decide if they want to add a custom category as a prefix or suffix. In the case of "Add .cross category suffix" users, it would set a custom category of ".cross" and set it to suffix. It _might_ be too big of a lift for this PR, but it would make this customizable enough to cover most use-cases.
> 
> Anyways, my second paragraph is more thinking out loud, and I don't think that has to block this PR.

This is how I had it in my original fork, [mentioned here](https://discord.com/channels/881212911849209957/881216126980341780/1445835529462616236) but decided to do it this way because of [mentions](https://discord.com/channels/881212911849209957/1412437936716124190/1455922389211811863) that allowing a suffix sub-category could be problematic so I choose to make it more restrictive, I don't mind changing back though

ill add in docs and bump the migration number when I get a chance 👍

Barcode-eng:
@jabloink Ahh, thanks for that context! Yeah, seems like something we should talk about in the future then, as to not block this PR.

s0up4200:
@jabloink can you look at the conflicts when you get the chance? 👍

jabloink:
> @jabloink can you look at the conflicts when you get the chance? 👍

Before I do anything, circling back to what @Barcode-eng mentioned and my original message, I do think it makes sense to move things under the current suffix option instead of custom category, it would allow more flexibility. Would rather do it now than ship a feature and change it later. I can make this PR a draft and make those changes if you'd like.

Would consist of changing suffix to affix and allow the selection of suffix or prefix and the ability to change the affix value.

s0up4200:
> > @jabloink can you look at the conflicts when you get the chance? 👍
> 
> 
> 
> Before I do anything, circling back to what @Barcode-eng mentioned and my original message, I do think it makes sense to move things under the current suffix option instead of custom category, it would allow more flexibility. Would rather do it now than ship a feature and change it later. I can make this PR a draft and make those changes if you'd like.
> 
> 
> 
> Would consist of changing suffix to affix and allow the selection of suffix or prefix and the ability to change the affix value.

Sure, no rush 👍

PR #929, opened 2026-01-01
feat(metrics): Collect event-driven Automation metrics

boxhock:
It doesn't seem like we're currently collecting event-driven metrics in this project, so I took some liberties in how to structurethat. I'm open to refactoring.

This will expose the following new metrics:

```
# Total number of automation rule runs
qui_automation_rule_run_total{instance_id, instance_name, rule_id, rule_name}

# Total number of torrents that matched the trackers in the rule
qui_automation_rule_run_torrents_matched_total{instance_id, instance_name, rule_id, rule_name}

# Total number of automation rule actions
qui_automation_rule_run_action_total{instance_id, instance_name, rule_id, rule_name, action}

# Total number of automation rule actions where conditions were met but the action was not performed
qui_automation_rule_run_action_not_performed_total{instance_id, instance_name, rule_id, rule_name, reason}
```

## Summary by CodeRabbit

* **New Features**
  * Added automation metrics: per-rule run counts, per-action outcomes, and split category metrics ("no match" vs "blocked"); metrics are wired into the metrics system and exposed when enabled.

* **Tests**
  * Added tests validating automation metrics emission and the automation metrics accessor.

* **Chores**
  * Service now accepts and uses an automation metrics collector to emit per-rule metrics after processing.


internal/metrics/collector/automation.go
```diff
@@ -0,0 +1,119 @@
...
+	RuleRunTagConditionNotMet               *prometheus.CounterVec
+	RuleRunTagSkippedMissingUnregisteredSet *prometheus.CounterVec
+	RuleRunCategoryApplied                  *prometheus.CounterVec
+	RuleRunCategoryConditionNotMetOrBlocked *prometheus.CounterVec
+	RuleRunDeleteApplied                    *prometheus.CounterVec
+	RuleRunDeleteConditionNotMet            *prometheus.CounterVec
+	RuleRunDeleteNotCompleted               *prometheus.CounterVec
+}
+
+var automationRuleRunLabels = []string{"instance_id", "rule_id", "rule_name"}
```
finevan: Most metrics that include `instance_id` also include `instance_name`, much like including `rule_name` here, I think the `instance_name` would help us make great, readable dashboards!

In most prometheus backends, each unique combination of labels+values should be effectively deduplicated to a single timeseries so constant tags aren't particularly expensive to report.

internal/services/automations/processor.go
```diff
@@ -63,6 +64,26 @@ type ruleRunStats struct {
...
+	metricsCollector.RuleRunPauseApplied.WithLabelValues(labels...).Add(float64(s.PauseApplied))
+	metricsCollector.RuleRunPauseConditionNotMet.WithLabelValues(labels...).Add(float64(s.PauseConditionNotMet))
+	metricsCollector.RuleRunTagConditionMet.WithLabelValues(labels...).Add(float64(s.TagConditionMet))
+	metricsCollector.RuleRunTagConditionNotMet.WithLabelValues(labels...).Add(float64(s.TagConditionNotMet))
+	metricsCollector.RuleRunTagSkippedMissingUnregisteredSet.WithLabelValues(labels...).Add(float64(s.TagSkippedMissingUnregisteredSet))
+	metricsCollector.RuleRunCategoryApplied.WithLabelValues(labels...).Add(float64(s.CategoryApplied))
+	metricsCollector.RuleRunCategoryConditionNotMetOrBlocked.WithLabelValues(labels...).Add(float64(s.CategoryConditionNotMetOrBlocked))
+	metricsCollector.RuleRunDeleteApplied.WithLabelValues(labels...).Add(float64(s.DeleteApplied))
+	metricsCollector.RuleRunDeleteConditionNotMet.WithLabelValues(labels...).Add(float64(s.DeleteConditionNotMet))
+	metricsCollector.RuleRunDeleteNotCompleted.WithLabelValues(labels...).Add(float64(s.DeleteNotCompleted))
```
finevan: I have a little code-smell here, but I admit it's largely because I haven't used automations and I don't have a clear sense of what we're trying to describe with these metrics.

`RuleRunMatchedTrackers` describes the number of torrents matched by a rule, based on the tracker criteria. I found the name a bit confusing because it's sort of describing `RuleRunTrackerConditionMet` (to match the naming scheme of other metrics).

Separate timeseries for each type of condition seem difficult to consume in a way that helps us understand the state of the system. Every sweep of the automation processor will increment a bunch of these for each torrent, most of which may have already been processed, and some of which won't be relevant because they don't match labels/categories.  Perhaps we're better off only describing the rule outcomes with metrics, rather than their individual conditions?

internal/metrics/collector/automation.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func NewAutomationCollector(r *prometheus.Registry) *AutomationCollector {
+	m := &AutomationCollector{
+		RuleRunTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "qui_automation_rule_run_total",
+			Help: "Total number of automation rule runs",
+		}, automationRuleRunLabels),
+		RuleRunMatchedTrackers: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Name: "qui_automation_rule_run_matched_trackers",
```
finevan: Prometheus recommends adding a `_total` suffix for accumulating counters.

https://prometheus.io/docs/practices/naming/
> Note that an accumulating count has total as a suffix, in addition to the unit if applicable

internal/metrics/collector/automation.go
boxhock: Agreed, it shouldn't change cardinality. I'll look to get that added!

internal/services/automations/processor.go
boxhock: So basically counting the outcome/action taken by the rule as a label? Something like `metric_name_tbd{outcome/action="speed_applied|speed_condition_not_met|..."}`?

As for `RuleRunMatchedTrackers`, upon closer inspection, I don't think this is counting trackers at all? Seems to be the number of torrents? If we go with the label approach above, I guess we don't even need this – could just `sum(increase())`

finevan: If you do opt for the outcome-based model, one option may be to reuse a single rule-applied metric name, with a tag describing what it did, something along these lines:
```
qui_automation_rule_applied{
  action:paused # or deleted, transfer_speed_limit, etc
  rule_id: identifier
  rule_name: name
  instance_id: identifier
  instance_name: name
}
```

This way we can construct a single dashboard which illustrates all actions using a single query. If new actions are added to qui, the query and dashboard needn't be updated.

Sorry for all the comments here, I get pretty excited about metrics/monitoring, and I think this is super cool!

finevan: > So basically counting the outcome/action taken by the rule as a label?

@boxhock Ha yes exactly! I think we came to the same conclusion at the same time :-)

> As for RuleRunMatchedTrackers, upon closer inspection, I don't think this is counting trackers at all? Seems to be the number of torrents? If we go with the label approach above, I guess we don't even need this – could just sum(increase())

Ya, it's a bit confusing, and I'm not sure it's worth reporting unless we have a specific use-case for it. I think `sum(increase())` is more useful. One edge-case may be if it's possible for a rule to have multiple outcomes, they'd be summed in that query, but I think that's reasonable.

boxhock: @finevan Appreciate your feedback here, definitely coming together a lot better now :)

I went with a `rule_run_total` counter + a `rule_run_action_total` where the action is a label. Please see if this looks better to you.

> One edge-case may be if it's possible for a rule to have multiple outcomes, they'd be summed in that query, but I think that's reasonable.

~@s0up4200 Any immediate plans to have a single automation with multiple actions? If not, then I'll keep it as-is. If multiple actions for a single torrent in a single rule might be a possibility in the future, I'll future-proof it and add a separate metric to track how many torrents were matched by the rule.~

internal/metrics/collector/automation.go
boxhock: Done!

internal/metrics/collector/automation.go
boxhock: Added! I thought I'd have to do a SQL query for it, but turned out to be much easier...

internal/services/automations/processor.go
finevan: I like it, this looks a lot cleaner!

I think we can remove the tag variants which include `_met` or `_not_met`, since those describe individual conditions for the automation rather than the outcome. When the automation isn't making changes, those would still pop up in our graphs, drawing attention away from the actions like pauses/deletions/etc.

boxhock: Follow-up on this:

> One edge-case may be if it's possible for a rule to have multiple outcomes, they'd be summed in that query, but I think that's reasonable.

I just realized multiple actions did get added now, so we will need a counter for how many torrents were matched by the rule.

@finevan I see your point. However, especially since a rule can have multiple actions now, it would be good to see how many torrents did not meet the criteria to be actioned. Do you think it would be reasonable to add another metric for conditions not met (with a label for the action it would take if conditions were met)?

We'd have something like this:

```
rule_run_total{rule_id...}
rule_run_torrents_matched{rule_id...}
rule_run_action_total{rule_id..., action="tag|category_applied|deleted"}
rule_run_condition_not_met_total{rule_id..., action="tag|category_applied|deleted"}
```

I'm open to better naming suggestions btw

finevan: > especially since a rule can have multiple actions now, it would be good to see how many torrents did not meet the criteria to be actioned. Do you think it would be reasonable to add another metric for conditions not met (with a label for the action it would take if conditions were met)?

I think this is a tricky problem, because the set of "not matched" torrents includes not only new additions, but also torrents that have already been matched and actioned by a rule (in fact it may be possible to create a cyclical set of rules). Furthermore, I think the tracker criteria is taken into account for the initial filter, and isn't captured by matched/not-matched.
Taking a high-level view, action outcomes mean something changed in the client that is perceivable, and once a state change occurs, it's set. However the rate of the matched/not-matched metrics incrementing isn't based on a state change, but the rate that automation rules are evaluated. If rules are evaluated very frequently, we'd see a pretty constant increase in the metric values, where the slope changes as conditions are met. I'm not sure how helpful that would be compared to application logs. I suppose what I'm getting at is: I think the metrics are a great way to illustrate what happened (what state changes resulted from the automation rule), and I suspect we're better off referring to the logs to understand _why_ the rules did or did not result in a state change.

I like to start with a use-case in mind for the metric, so I can determine whether or not a particular implementation allows me to solve my problem, and I'm not really sure how I would use the not-matched data given the caveats above. This isn't to say that it's not meaningful, just that I don't feel confident that I'd be able get useful information out of it myself.

> I'm open to better naming suggestions btw

Naming is [hard](https://martinfowler.com/bliki/TwoHardThings.html).

boxhock: Yeah, I think I've been a bit to broad in my inclusion of everything that was "not met". I think my initial use case for that is covered by what we have now + `rule_run_torrents_matched`.

I still see a use case in capturing instances of when things were not actioned when they probably should have been. In other words, conditions were met but still not actioned.

Looking some more at it, I think these reasons would qualify for my use case:
- ~Delete conditions were met, but the file was not deleted (`delete_not_completed`)~ <- _removed in #926_
- Category change was blocked (`category_condition_not_met_or_blocked`) <- _or maybe not, since it seems to include "not_met"_
- Tags were not actioned because "condition uses IS_UNREGISTERED but health data isn't available" (`tag_skipped_missing_unregistered_set`)

Do you agree it could be useful to keep these, and throw out the rest of the "not met" conditions? It does potentially increase on every rule evaluation despite potentially no state changes, but it does highlight potential issues that you might want to look into.

I don't need much convincing to not keep this, though. The first one is already gone before this PR is merged, so it would effectively only be a metric for a single reason for why the action was not performed. Could be useful in the future if more reasons are added.

boxhock: And just to justify `rule_run_torrents_matched{}` – I think this is useful in understanding how many torrents could potentially be actioned by the rule evaluation. It allows for monitoring some things like "how many of the torrents that match the tracker filtering was actioned", catching things like during my earlier testing of the Automations PR when 100% of torrents were actioned when that number should have been much lower.

I know it does increase a metric not tied to a state change, but I think my use case is useful enough to keep that one.

finevan: You raise a great point, we can classify some cases as "action applied" (for example, the category was changed)  and others as "action could not be applied" (for example, the torrent is a cross-seed and the category cannot be changed independently from the original).

I think the trouble is the data that we have right now doesn't distinguish between "the category should not be updated" and "the category would have been updated, but isn't allowed due to cross-seed". Perhaps it would make sense to split `ruleRunStats.CategoryConditionNotMetOrBlocked` into `CategoryConditionNotMet` + `CategoryConditionBlocked`, where the latter is what you'd like to report?

What do you think about using a separate metric based name for the "action could not be applied" metrics? That way we can create future-proof graphs for each without having to update dashboards when new types of failures are added?

boxhock: Makes perfect sense to me. Will update!

boxhock: Added `rule_run_action_not_performed_total{rule_id..., reason="category_blocked|tag_skipped_missing_unregistered_set"}`. I think the metrics we have now are pretty reasonable.

In case of any other feedback I'll look at it tomorrow/this weekend!

internal/services/automations/processor.go
```diff
@@ -57,11 +58,29 @@ type ruleRunStats struct {
...
+	metricsCollector.GetAutomationRuleRunTotal(rule.InstanceID, instanceName, rule.ID, rule.Name).Inc()
+	metricsCollector.GetAutomationRuleRunTorrentsMatchedTotal(rule.InstanceID, instanceName, rule.ID, rule.Name).Add(float64(s.MatchedTrackers))
+
+	actionPerformedMetric := metricsCollector.GetAutomationRuleRunActionTotal(rule.InstanceID, instanceName, rule.ID, rule.Name)
+	actionPerformedMetric.WithLabelValues("speed_applied").Add(float64(s.SpeedApplied))
+	actionPerformedMetric.WithLabelValues("share_applied").Add(float64(s.ShareApplied))
+	actionPerformedMetric.WithLabelValues("pause_applied").Add(float64(s.PauseApplied))
+	actionPerformedMetric.WithLabelValues("tag_condition_met").Add(float64(s.TagConditionMet))
+	actionPerformedMetric.WithLabelValues("category_applied").Add(float64(s.CategoryApplied))
+	actionPerformedMetric.WithLabelValues("delete_applied").Add(float64(s.DeleteApplied))
```
finevan: This is total bike-shedding, but while we're here I think we can make the metric names a bit easier to follow with something like this, because the `action` tag implies it was applied (I don't have strong feelings about these names in particular, goal is avoiding the `_applied` suffix and making the tags describe their state changes clearly)
```suggestion
	actionPerformedMetric.WithLabelValues("speed_limit").Add(float64(s.SpeedApplied))
	actionPerformedMetric.WithLabelValues("share_limit").Add(float64(s.ShareApplied))
	actionPerformedMetric.WithLabelValues("pause").Add(float64(s.PauseApplied))
	actionPerformedMetric.WithLabelValues("tag").Add(float64(s.TagConditionMet))
	actionPerformedMetric.WithLabelValues("category").Add(float64(s.CategoryApplied))
	actionPerformedMetric.WithLabelValues("delete").Add(float64(s.DeleteApplied))
```

internal/services/automations/processor.go
finevan: Looks awesome, thank you so much!

internal/services/automations/processor.go
boxhock: Updated!

boxhock: One thing I don't love with the metrics PR right now is that "tag" counts the torrents that were tagged, not how many tags were added or removed. I think that's a correct and consistent use of that metric, so I wonder if we should add a separate metric to count the individual tag adds/removes, e.g. `rule_run_tags_actioned{rule_id..., action="add|remove"}`. Or make a more generic metric to count individual changes like that, but I think it would only be tagging for now. What do you think?

finevan: Hmm, I think ideally the `action` tag would have values `tag_added` and `tag_removed` for tags, if the value indicates that an item was considered rather than updated I don't think it fits the "action performed" metric timeseries, since it is likely to distract from other actions by diluting the data.

One option is to not report the tag data for now, since I suspect adding/removing tags isn't the most interesting action, and we could consider adding support later. This is probably the approach I'd take because it avoids making this PR too large so it's easier to review, and once we've used the initial set of data we may be in a better position to decide how we want to handle the tag data.

boxhock: The value currently indicates if an item was updated, not considered. What it doesn't include is how many tag additions/removals there were for each item that was updated.

Agreed on the assessment that we can cover that in a future PR after we get some experience with the current set of metrics :)

internal/metrics/collector/automation.go
```diff
@@ -0,0 +1,85 @@
...
+			Subsystem: "automation",
+			Name:      "rule_run_total",
+			Help:      "Total number of automation rule runs",
+		}, []string{"instance_id", "instance_name", "rule_id", "rule_name"}),
+		RuleRunTorrentsMatchedTotal: prometheus.NewCounterVec(prometheus.CounterOpts{
+			Namespace: "qui",
+			Subsystem: "automation",
+			Name:      "rule_run_torrents_matched_total",
+			Help:      "Total number of torrents that matched the trackers in the rule",
+		}, []string{"instance_id", "instance_name", "rule_id", "rule_name"}),
```
finevan, line 29: Thoughts on updating this to a guage metric? This way it shows us the most recent number of torrents matched any time it's reported, and we don't have to work backwards from the slope. If the reporting interval is less frequent than the rule execution interval, it won't necessarily tell us exactly how many items were matched every run, however I think that's also true in many cases with an accumulating counter. and the gauge is easier to understand.

```suggestion
		RuleRunTorrentsMatchedTotal: prometheus.NewGaugeVec(prometheus.GaugeOpts{
			Namespace: "qui",
			Subsystem: "automation",
			Name:      "rule_run_torrents_matched",
			Help:      "Total number of torrents that matched the trackers in the rule",
		}, []string{"instance_id", "instance_name", "rule_id", "rule_name"}),
```

We'd need to update `processor.go`
```diff
- metricsCollector.GetAutomationRuleRunTorrentsMatchedTotal(rule.InstanceID, instanceName, rule.ID, rule.Name).Add(float64(s.MatchedTrackers))
+ metricsCollector.GetAutomationRuleRunTorrentsMatchedTotal(rule.InstanceID, instanceName, rule.ID, rule.Name).Set(float64(s.MatchedTrackers))
```

finevan (approved):
I'm really excited to try this out. Added one more minor thought, but I trust your judgement on whether or not it's worthwhile

Issue #931, opened 2026-01-01
Submenus for "Unregistered torrents" or "Errors" conditions in the dashboard menu.


Vodayaa:
**Is your feature request related to a problem? Please describe.**

On the dashboard menu, I like seeing the numeric summary on the condition of torrents. However, there is currently no option to use the overview to see specifics.

**Describe the solution you'd like**

Clicking on the “Unregistered torrents” or “Errors” could bring up a menu that shows the torrents reported under those categories. 

**Describe alternatives you've considered**

Filters within each qb instance could be used to achieve a similar end result.

**Additional context**

![image](https://github.com/user-attachments/assets/13817f0f-9cbe-4abe-bf61-186e22986643)

PR #937, opened 2026-01-02
feat(web): make dashboard error counts clickable

Gykes:
Makes the "Unregistered torrents", "Tracker Down", and "Errors" counts on dashboard instance cards clickable. Clicking navigates to the instance's torrent list with the appropriate status filter pre-applied, allowing users to quickly iSee what torrents are erroring.

The implementation uses `localStorage` to set the filter state before navigation, which aligns with how the existing filter persistence system works. A potential future enhancement would be to move to URL search params (e.g., `/instances/1?status=unregistered`), which would make filtered views shareable via URL and more consistent with React routing patterns.

For Devs: First PR here, let me know if I goofed up somewhere.

Fixes: https://github.com/autobrr/qui/issues/931

## Summary by CodeRabbit

* **Improvements**
  * Dashboard status rows (Unregistered torrents, Tracker Down, Errors) are now clickable and navigate to the corresponding filtered instance view.
  * Clicking a status row automatically applies the related filter so the destination view shows only matching items while counts and visuals are preserved.
  * Added full-width, hover and transition styling for clearer interactive feedback.

* **Bug Fixes**
  * Safer persistence: saved filters/read-write operations now handle errors gracefully to prevent crashes and logging issues.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ tracker-icons/*
 backups/*
 .gocache/*
 torrentfiles/*
+analyses/*
--- a/distrib/docker/Dockerfile
+++ b/distrib/docker/Dockerfile
@@ -72,7 +72,7 @@ ENV HOME="/config" \
     XDG_DATA_HOME="/config"
 
 # Install runtime dependencies
-RUN apk --no-cache add ca-certificates curl tzdata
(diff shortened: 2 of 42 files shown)


Gykes:
@coderabbitai generate docstrings

PR #1012, opened 2026-01-02
fix(web): use correct KiB/s units for speed limit preferences

Gykes:
When entering a speed limit (e.g., 15 MiB/s) in Instance Preferences, the displayed value would be incorrect (showing ~1.1 instead of 15).  The conversion functions assumed qBittorrent's API uses bytes/s, but according to the docs `https://github.com/qbittorrent/qBittorrent/wiki/WebUI-API-(qBittorrent-5.0)`, the `dl_limit`, `up_limit`, `alt_dl_limit`, and `alt_up_limit` preferences are in KiB/s.

This caused a 1024x error in both directions:
  - Saving: 15 MiB/s × 1,048,576 = 15,728,640 (sent as if bytes/s, interpreted as KiB/s)
  - Loading: value ÷ 1,048,576 instead of ÷ 1,024

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected speed limit unit conversion logic in the preferences form to accurately handle KiB/s to MiB/s conversions for display and API communication.


Barcode-eng:
I'm not sure if I'm mis-reading the PR description, or if there have been other changes since this PR, but the current behavior seems to be correct for me.

If I enter a speed limit of 15 MiB/s in qui, the speed limit of qbit is set to 15 MiB/s. Qui also displays a speed limit of 15 MiB/s.

If anything, it seems like the qbit docs are incorrect?

Gykes:
> I'm not sure if I'm mis-reading the PR description, or if there have been other changes since this PR, but the current behavior seems to be correct for me.
> 
> If I enter a speed limit of 15 MiB/s in qui, the speed limit of qbit is set to 15 MiB/s. Qui also displays a speed limit of 15 MiB/s.
> 
> If anything, it seems like the qbit docs are incorrect?

I think there have been some changes since this PR. I updated my develop and could not reproduce the issue I was having. 

Before: 
Go to your instance settings.
Type 15 in the download limit.
Would be shown as 1.1 and would need to manually go up to 15 using the up/down arrow.

After: It works as intended.

s0up4200:
> > I'm not sure if I'm mis-reading the PR description, or if there have been other changes since this PR, but the current behavior seems to be correct for me.
> > If I enter a speed limit of 15 MiB/s in qui, the speed limit of qbit is set to 15 MiB/s. Qui also displays a speed limit of 15 MiB/s.
> > If anything, it seems like the qbit docs are incorrect?
> 
> I think there have been some changes since this PR. I updated my develop and could not reproduce the issue I was having.
> 
> Before: Go to your instance settings. Type 15 in the download limit. Would be shown as 1.1 and would need to manually go up to 15 using the up/down arrow.
> 
> After: It works as intended.

Should we close this then? 👍 Thanks either way of course!

Gykes:
Absolutely :)

PR #1052, opened 2026-01-02
fix(web): prevent tracker nicknames from disappearing on edit

s0up4200:
## Summary
- Fix tracker rows disappearing when editing nicknames by using stable unique keys
- Handle case where only secondary domains have torrents (not the primary domain)

## Problem
Two issues caused trackers to vanish after editing customization nicknames:

1. **Key collisions**: Using `displayName` as both Map key and React key could cause rows to overwrite each other if a nickname matched another tracker's displayName or raw domain

2. **Primary domain not in stats**: If only secondary domains had torrents (not the primary domain), the customization row was never created because Pass 1 only created rows for primary domains

## Solution
- Add `rowKey` field to `ProcessedTrackerStats` with format `customization:<id>` or `domain:<domain>`
- Use `rowKey` as Map key instead of `displayName`
- Create customization row in Pass 2 when secondary domains have stats but primary doesn't
- Use `tracker.rowKey` as React key for mobile Card and desktop TableRow

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced stability of tracker information display in the dashboard with improved aggregation logic for tracking statistics, ensuring consistent and reliable rendering of tracker data across sessions.

### DIFF
--- a/web/src/pages/Dashboard.tsx
+++ b/web/src/pages/Dashboard.tsx
@@ -891,6 +891,7 @@ interface ProcessedTrackerStats extends TrackerTransferStats {
   displayName: string
   originalDomains: string[]
   customizationId?: number
+  rowKey: string
 }
 
 interface TrackerBreakdownCardProps {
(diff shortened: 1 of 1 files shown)


s0up4200:
superseded by https://github.com/autobrr/qui/pull/1120

PR #1073, opened 2026-01-04
feat(metrics): add tracker metrics

Winter:
Implements additional Prometheus metrics specifically for trackers.

Added the following new metrics:
| Metric | Description |
|-------------|-------------|
| `qbittorrent_tracker_torrents` | Number of torrents by tracker |
| `qbittorrent_tracker_uploaded_bytes` | Total uploaded data in bytes by tracker |
| `qbittorrent_tracker_downloaded_bytes` | Total downloaded data in bytes by tracker |
| `qbittorrent_tracker_total_size_bytes` | Total content size in bytes by tracker |

Related Discussions:
* #980 

## Summary by CodeRabbit

* **New Features**
  * Metrics now report per-tracker details: torrent counts, uploaded/downloaded bytes, and total size for each tracker.
  * Tracker customizations are applied so trackers are shown using configured display names and inclusion rules.

* **Tests**
  * Test suites updated for the expanded metrics output and collector/registry behavior.
  * Added unit tests covering tracker grouping and fallback logic for customized trackers.

### DIFF
--- a/internal/metrics/collector/torrent.go
+++ b/internal/metrics/collector/torrent.go
@@ -5,7 +5,7 @@ package collector
 
 import (
 	"context"
-	"slices"
+	"maps"
 	"strings"
 	"time"
 
@@ -299,31 +299,11 @@ func (c *TorrentCollector) Collect(ch chan<- prometheus.Metric) {
 			}
 
 			if counts.TrackerTransfers != nil {
-				trackerStats := map[string]*qbittorrent.TrackerTransferStats{}
-
-				for domain, stats := range counts.TrackerTransfers {
-					displayName, shouldInclude := resolveTrackerForStats(domain, trackerCustomizations)
-
-					if !shouldInclude {
+				trackerStats := groupTrackerTransfersForMetrics(counts.TrackerTransfers, trackerCustomizations)
+				for trackerName, stats := range trackerStats {
+					if strings.TrimSpace(trackerName) == "" {
 						continue
 					}
-
-					if existing, ok := trackerStats[displayName]; ok {
-						existing.Count += stats.Count
-						existing.Uploaded += stats.Uploaded
-						existing.Downloaded += stats.Downloaded
-						existing.TotalSize += stats.TotalSize
-					} else {
-						trackerStats[displayName] = &qbittorrent.TrackerTransferStats{
-							Count:      stats.Count,
-							Uploaded:   stats.Uploaded,
-							Downloaded: stats.Downloaded,
-							TotalSize:  stats.TotalSize,
-						}
-					}
-				}
-
-				for trackerName, stats := range trackerStats {
 					ch <- prometheus.MustNewConstMetric(
 						c.trackerTorrentsDesc,
 						prometheus.GaugeValue,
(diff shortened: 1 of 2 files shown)


internal/metrics/collector/torrent.go
```diff
@@ -115,6 +122,30 @@ func NewTorrentCollector(syncManager *qbittorrent.SyncManager, clientPool *qbitt
...
+			"Total downloaded data in bytes by tracker",
+			[]string{"instance_id", "instance_name", "tracker_name"},
+			nil,
+		),
+		trackerTotalSizeDesc: prometheus.NewDesc(
+			"qbittorrent_tracker_total_size_bytes",
+			"Total content size in bytes by tracker",
+			[]string{"instance_id", "instance_name", "tracker_name"},
+			nil,
+		),
```
Barcode-eng, line 136: Would it make more sense to just apply the "tracker_name" label to the existing metrics, instead of creating new? Otherwise I am afraid we end up with duplicate metrics.

You can always `sum()` the existing metrics to combine different labels.

Winter: Wouldn't that cause issues with any existing dashboards?

The only concern I have is none of the existing metrics apply appropriately to tracker specific data; which metric do we stick `tracker_name` on for uploaded & downloaded?

`qbittorrent_alltime_*` is an absolute count that only increases, same with `qbittorrent_session_*` (resets, of course) whereas tracker specific data can increase and decrease.

`qbittorrent_tracker_torrents` can definitely be folded into labels on `qbittorrent_torrents_*`

As for `qbittorrent_tracker_totalsize` that can probably go, you're right it can just be `sum`'d.

Barcode-eng: > Wouldn't that cause issues with any existing dashboards?

It _shouldn't_. Worst case it just adds another series for each tracker in their dashboards, which can easily be solved by `sum`ing it.

> The only concern I have is none of the existing metrics apply appropriately to tracker specific data; which metric do we stick tracker_name on for uploaded & downloaded?
> 
> qbittorrent_alltime_* is an absolute count that only increases, same with qbittorrent_session_* (resets, of course) whereas tracker specific data can increase and decrease.

Ah, is there a possibility that the tracker stats don't match exactly with the current all-time or session stats? If not, I'd just add the label to whatever the tracker stats are (all-time or session). If the sum of the tracker stats don't equate with any of the existing metrics, then yeah I think new metrics are justified (but I think we should also name it appropriately by all-time or session).

Winter: All of the tracker specific data represents the present state of torrents list.

For example; if you have torrents A, B, and C with each having the same tracker "Example", and each torrent was 100 MB in file size. 

A `qbittorrent_tracker_downloaded_bytes` is created with a label "Example" with a value of 300 MB. If you remove A, leaving B and C `qbittorrent_tracker_downloaded_bytes` now reports 200 MB.

I do agree with the deduplicating of data, with totalsize equaling upload + download, and placing additional labels on the status related metrics.

But I believe the `qbittorrent_tracker_downloaded_bytes` and `qbittorrent_tracker_uploaded_bytes` are sufficiently justified as we cannot `sum` to reach any value.

With respect to the `session` and `alltime` naming for trackers; `alltime` doesn't make sense as it's just present state and I'd need to look into how `session` could be calculated.

Winter: I wrote these responses when I was tired, and looking back, most of them don't make a terrible amount of sense.

After playing around with different implementations, I believe we should keep the present implementation as is. 

My justification is as follows: 

1. `qbittorrent_tracker_downloaded_bytes` and `qbittorrent_tracker_uploaded_bytes` don't represent alltime or session related metrics.
2. `qbittorrent_tracker_total_size_bytes` is important as `qbittorrent_tracker_downloaded_bytes` will be inaccurate for cross-seeded torrents as they usually have 0 bytes downloaded.
3. Splitting `qbittorrent_tracker_torrents` into labels onto the status metrics (`qbittorrent_torrents_*`) creates a lot of unnecessary complexity for users.

Barcode-eng: That makes sense to me @Winter. Too bad qbit doesn't just split these metrics out natively, I think it would have been much better to have a combined metric with labels for detail (obviously nothing you can do about).

Thanks for the explanation! And thanks for implementing this, I am looking forward to using these metrics 😊

internal/metrics/collector/torrent.go
```diff
@@ -115,6 +124,30 @@ func NewTorrentCollector(syncManager *qbittorrent.SyncManager, clientPool *qbitt
 			[]string{"instance_id", "instance_name", "type"},
 			nil,
 		),
+		trackerTorrentsDesc: prometheus.NewDesc(
+			"qbittorrent_tracker_torrents",
+			"Number of torrents by tracker",
+			[]string{"instance_id", "instance_name", "tracker_name"},
```
finevan, line 116: Do you think it would make sense to include `tracker_id` as a tag here as well? The tracker display name could change over time, and a stable identifier may be helpful. It would provide symmetry with instance_id+instance_name as well

internal/metrics/collector/torrent.go
```diff
@@ -267,6 +314,66 @@ func (c *TorrentCollector) Collect(ch chan<- prometheus.Metric) {
...
+						c.trackerTorrentsDesc,
+						prometheus.GaugeValue,
+						float64(stats.Count),
+						instanceIDStr,
+						instanceName,
+						trackerName,
+					)
+					ch <- prometheus.MustNewConstMetric(
+						c.trackerUploadedDesc,
+						prometheus.GaugeValue,
```
finevan, line 317: Might be worth leaving a comment here why we're using a gauge instead of a counter.

I'd toyed with using a counter for more efficient rate plotting, but it's tricky because we're summing upload data across torrents that are currently in the system, so we "lose" data when torrents are removed. We could track torrent data between iterations, and keep a sum per-tracker of impact from torrents that are no longer reported, but that might be more complicated than it's worth.

internal/metrics/collector/torrent.go
```diff
@@ -267,6 +314,66 @@ func (c *TorrentCollector) Collect(ch chan<- prometheus.Metric) {
...
+						float64(stats.Uploaded),
```
finevan, line 318: On large, long-lived instances above ~10pb, should we worry about losing precision from float64 conversion? I wonder if it's worth reporting the current-session stats to use for transfer-rate calculation rather than (or in addition to) lifetime total?

finevan (commented):
This looks awesome and I'm excited to try it out!

I've added a couple comments, but I'm not affiliated with the project so please don't feel obligated to address them!

s0up4200:
@Winter do you want to handle the merge conflicts or should I do it when I get time to review it?

Winter:
> @Winter do you want to handle the merge conflicts or should I do it when I get time to review it?

I can handle the merge conflicts next week -- just away on a wedding trip at the moment, returning today!

@finevan I really appreciate your feedback, I'll be able to address them next week ❤️

s0up4200:
Should we use gauges or counters for upload/download?

s0up4200 (approved):

PR #1237, opened 2026-01-13
fix(crossseed): support v2 hashes for recheck

s0up4200:
Potential fix for #1220

qBittorrent indexes v2-only torrents by their SHA256 infohash, but cross-seed only computed the v1 SHA1 hash when parsing .torrent data. Follow-up recheck/resume actions would then miss the just-added torrent, leaving it in a missing files state until a manual recheck.

- Compute v1 + v2 hashes when parsing torrents (prefer v2 for v2-only)
- Pass both hashes through cross-seed recheck calls
- Resolve InfohashV1/V2 variants in BulkAction and dedupe canonical hashes

## Summary by CodeRabbit

* **New Features**
  * Full dual-hash (V1/V2) support: torrents with V2 info-hash are detected and propagated throughout cross-seed and add/recheck flows.

* **Bug Fixes**
  * Variant-aware hash resolution for bulk actions with duplicate suppression and targeted retry for newly-added torrents.
  * Recheck/resume and cache handling now consider both hash variants to reduce missed matches.

* **Tests**
  * Added and updated tests covering hybrid V1/V2 resolution and adjusted cross-seed scenarios.

### DIFF
--- a/internal/services/crossseed/service.go
+++ b/internal/services/crossseed/service.go
@@ -3053,7 +3053,22 @@ func (s *Service) processCrossSeedCandidate(
 	}
 
 	// Check if torrent already exists
-	existingTorrent, exists, err := s.syncManager.HasTorrentByAnyHash(ctx, candidate.InstanceID, []string{torrentHash, torrentHashV2})
+	hashes := make([]string, 0, 2)
+	seenHashes := make(map[string]struct{}, 2)
+	for _, hash := range []string{torrentHash, torrentHashV2} {
+		trimmed := strings.TrimSpace(hash)
+		if trimmed == "" {
+			continue
+		}
+		canonical := strings.ToLower(trimmed)
+		if _, ok := seenHashes[canonical]; ok {
+			continue
+		}
+		seenHashes[canonical] = struct{}{}
+		hashes = append(hashes, trimmed)
+	}
+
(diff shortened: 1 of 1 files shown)


internal/services/crossseed/parsing.go
```diff
@@ -492,33 +493,37 @@ func ParseTorrentName(torrentBytes []byte) (name string, hash string, err error)
...
 func ParseTorrentMetadata(torrentBytes []byte) (name string, hash string, files qbt.TorrentFiles, err error) {
-	name, hash, files, _, err = ParseTorrentMetadataWithInfo(torrentBytes)
+	name, hash, _, files, _, err = ParseTorrentMetadataWithInfo(torrentBytes)
 	return name, hash, files, err
 }
 
 // ParseTorrentMetadataWithInfo extracts comprehensive metadata from torrent bytes,
 // including the raw metainfo.Info for piece-level operations.
-func ParseTorrentMetadataWithInfo(torrentBytes []byte) (name, hash string, files qbt.TorrentFiles, info *metainfo.Info, err error) {
+func ParseTorrentMetadataWithInfo(torrentBytes []byte) (name, hashV1, hashV2 string, files qbt.TorrentFiles, info *metainfo.Info, err error) {
```
Barcode-eng: At this point, would it be better if we just return a struct? Thinking in terms of maintainability and readability

PR #1251, opened 2026-01-16
feat(automation): add execute external program automation action

0rkag:
## Summary
  - Adds a new automation action type that allows executing external programs when workflow conditions are met
  - Refactors external program execution logic into a reusable `externalprograms` package
  - Updates the workflow dialog UI to support configuring external program actions

  ## Changes
  - New `externalprograms/exec.go` with centralized execution logic
  - Enhanced argument parsing and templating support
  - Workflow UI now allows selecting and configuring external programs as actions
  - Refactored existing consumers (crossseed, API handlers) to use the new shared executor

  ## Test plan
  - [ ] Create a workflow with an external program action
  - [ ] Verify program executes when workflow conditions trigger
  - [ ] Test argument templating with torrent variables
  - [ ] Confirm existing external program functionality still works

## Summary by CodeRabbit

* **New Features**
  * Execute external programs from automation rules with conditions, rule metadata capture, and per-instance queuing.
  * Configurable executor with sync/async modes, optional output capture, argument templating, and allowlist enforcement.
  * UI: "Execute external program" action block, dropdown, validation, hydration, and workflow integration.
  * New activity events for program start and failure.

* **Tests**
  * Extensive unit and integration tests covering executor, handlers, automations, stores, and allowlist behavior.

* **Chores**
  * Added two entries to .gitignore.

### DIFF
--- a/internal/services/automations/external_program_integration_test.go
+++ b/internal/services/automations/external_program_integration_test.go
@@ -412,7 +412,7 @@ func TestIntegration_ExternalProgram_StoreOperations(t *testing.T) {
 
 	// Verify deletion
 	_, err = store.GetByID(ctx, program.ID)
-	assert.Equal(t, models.ErrExternalProgramNotFound, err)
+	assert.ErrorIs(t, err, models.ErrExternalProgramNotFound)
 
 	// List should be empty
 	programs, err = store.List(ctx)
--- a/web/src/components/instances/preferences/WorkflowsOverview.tsx
+++ b/web/src/components/instances/preferences/WorkflowsOverview.tsx
@@ -166,6 +166,10 @@ function formatAction(action: AutomationActivity["action"]): string {
       return "Share"
     case "paused":
       return "Pause"
+    case "external_program_started":
+      return "Started"
+    case "external_program_failed":
+      return "Failed"
     default:
       return action
   }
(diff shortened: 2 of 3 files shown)


internal/externalprograms/exec.go
```diff
@@ -0,0 +1,132 @@
...
+			// Unix/Linux: Build command string and spawn in a terminal
+			// Use shellquote library to properly escape for Unix shells
+			allArgs := append([]string{program.Path}, args...)
+			fullCmd := shellquote.Join(allArgs...)
+			// Try to find an available terminal emulator and spawn the command in it
+			cmd = createTerminalCommand(fullCmd)
+		} else {
+			// Launch directly without terminal (for GUI apps or background processes)
+			cmd = exec.Command(program.Path, args...)
+		}
```
Barcode-eng: Do we care that we are running these commands without any context/timeout? Just thinking of possible footguns here... If the user has an automation that runs every 1 minute, and it launches an external program that executes for longer than 1 minute, the user might get some unintended behavior.

Should we add a timeout equal to the automation interval, or add configurable execution timeouts for the external programs?

0rkag: Yeah i was wondering that too - just refactored the current code for executing stuff a bit - also just running it in a goroutine to get it to execute async is a bit weird...
In addition to that the `UseTerminal` flag is pretty shaky as well. But yeah I didn't wanna rewrite half of the codebase for my first commits as that can piss people off :)

What do you think should I work on this PR more and implement some improvements (at least some stuff that I would find useful / good engineering practices?

Barcode-eng: I should preface this by saying I'm just a nobody, so don't want you to spend lots of time just because I have opinions 😅 So if @s0up4200 or others don't see a problem with this, we should be fine tackling this in the future.

But with that said, I usually find it very useful to pass contexts wherever appropriate. Contexts in this case can be useful when shutting down qui, allowing us to give external programs a chance to finish execution before completely shutting down. There is a `ctx` in the Automations service loop, that I think would be appropriate to pass down to the external program execution, and I don't think that would be very difficult either.

With a context being passed down, it should be trivial to add the automation interval as a timeout to the execution context.

Lastly, regardless if you decide to add this or not, we should update the documentation to explain this feature that you've added (and any caveats, such as execution time vs automation interval) 😊

0rkag:
Superseeded by #1284

PR #1260, opened 2026-01-16
docs(automations): clarify tag behavior

jabloink:
already have a couple people ask questions about this, not sure if this makes it clear enough if anyone has input

## Summary by CodeRabbit

* **Documentation**
  * Clarified Tag action behavior in automations: tags are added when conditions match and removed when they don’t; modes limit which of those behaviors occur but do not alter the action when conditions match. Notably, in remove mode tags are removed on mismatch and are not added on match.

### DIFF
--- a/documentation/docs/features/automations.md
+++ b/documentation/docs/features/automations.md
@@ -218,7 +218,7 @@ Add or remove tags from torrents.
 | `remove` | Only remove from non-matches |
 
 :::note
-Tags are only ever added when conditions match and removed when conditions don't match. Mode limits which of these behaviors occur and does not change the action when conditions match. Even when using `remove` mode, tags will be removed when the torrent does **not** match the conditions; what it won't do is add tags to torrents that do match.
+Mode does not change the way torrents are flagged, meaning, even with `mode: remove`, tags will be removed if the torrent does **NOT** match the conditions. `mode: remove` simply means that tags will not be added to torrents that do match.
 :::
 
 Options:


documentation/docs/features/automations.md
```diff
@@ -217,6 +217,10 @@ Add or remove tags from torrents.
 | `add` | Only add to matches |
 | `remove` | Only remove from non-matches |
 
+:::note
+Tags are only ever added when conditions match and removed when conditions don't match. Mode limits which of these behaviors occur and does not change the action when conditions match. Even when using `remove` mode, tags will be removed when the torrent does **not** match the conditions; what it won't do is add tags to torrents that do match.
```
Barcode-eng: This might be a small nit, but...

> Mode limits which of these behaviors occur and does not change the action when conditions match.

This doesn't seem correct, in that Mode _does_ change what action to take when conditions match (`full` will add tags on matches, but `remove` will not add tags on matches). It might just be me being stupid, but I think the current text is a bit ambiguous.

This is the note taken from [tqm](https://github.com/autobrr/tqm), which I wonder might be a bit more clear (from L5)?

```
        # Mode is optional and defaults to "full" if not specified
        # "mode: full" means tag will be added to
        # torrent if matched and removed from torrent if not
        # use `add` or `remove` to only add/remove respectivly
        # NOTE: Mode does not change the way torrents are flagged,
        # meaning, even with "mode: remove",
        # tags will be removed if the torrent does NOT match the conditions.
        # "mode: remove" simply means that tags will not be added
        # to torrents that do match.
```

jabloink: I can see what you're saying, am trying to convey that setting the mode to remove doesn't change the action when a condition matches which is what most people think happens. 

I didn't want to copy and paste it at the time, but I do like tqms description... I will update it to that unless anyone comes up with something better. 👍

PR #1272, opened 2026-01-18
fix(crossseed): prevent automation search stalls

s0up4200:
- Bound completion-priority scheduler wait so one rate-limited/down indexer can’t stall the whole job.
- Respect ctx cancellation in cross-seed search waits.
- Fix UI apply toast when per-instance results are omitted.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate callback invocations and ensured per-search cancellation and consistent timeout/error handling.

* **Improvements**
  * Enforced a 30-second cap for completion-priority searches to avoid indefinite waits.
  * Enhanced cross-seed result notifications with clearer, contextual toasts (success/warning/error), pluralization, and handling for results that complete without detailed information.

* **Tests**
  * Updated scheduler tests to assert the new completion-priority max-wait behavior.

### DIFF
--- a/internal/services/crossseed/service.go
+++ b/internal/services/crossseed/service.go
@@ -5291,10 +5291,10 @@ func (s *Service) SearchTorrentMatches(ctx context.Context, instanceID int, hash
 	case err := <-errCh:
 		return nil, wrapCrossSeedSearchError(err)
 	case <-waitCtx.Done():
-		if ctx.Err() != nil {
-			return nil, wrapCrossSeedSearchError(ctx.Err())
+		if errors.Is(waitCtx.Err(), context.DeadlineExceeded) {
+			return nil, wrapCrossSeedSearchError(errors.New("search timed out"))
 		}
-		return nil, wrapCrossSeedSearchError(errors.New("search timed out"))
+		return nil, wrapCrossSeedSearchError(waitCtx.Err())
 	}
 
 	searchResults := searchResp.Results


internal/services/crossseed/service.go
```diff
@@ -5277,6 +5280,8 @@ func (s *Service) SearchTorrentMatches(ctx context.Context, instanceID int, hash
 		// continue
 	case err := <-errCh:
 		return nil, wrapCrossSeedSearchError(err)
+	case <-ctx.Done():
+		return nil, wrapCrossSeedSearchError(ctx.Err())
 	case <-time.After(5 * time.Minute):
 		return nil, wrapCrossSeedSearchError(errors.New("search timed out"))
```
Barcode-eng: Not necessarily a blocker for this PR, and I see the Search() function applies a fallback timeout, but this seems like a good place to move this timeout into the context. This helps making the timeout explicit and ensuring Search() doesn't continue if the caller/this function has already returned early, if they're different.

```suggestion
	case <-ctx.Done():
		return nil, wrapCrossSeedSearchError(ctx.Err())
```

Barcode-eng (approved):

internal/services/crossseed/service.go
```diff
@@ -5277,7 +5290,10 @@ func (s *Service) SearchTorrentMatches(ctx context.Context, instanceID int, hash
 		// continue
 	case err := <-errCh:
 		return nil, wrapCrossSeedSearchError(err)
-	case <-time.After(5 * time.Minute):
+	case <-waitCtx.Done():
+		if ctx.Err() != nil {
+			return nil, wrapCrossSeedSearchError(ctx.Err())
+		}
 		return nil, wrapCrossSeedSearchError(errors.New("search timed out"))
```
Barcode-eng:
```suggestion
		if errors.Is(waitCtx.Err(), context.DeadlineExceeded) {
			return nil, wrapCrossSeedSearchError(errors.New("search timed out"))
		}
		return nil, wrapCrossSeedSearchError(waitCtx.Err())
```

Barcode-eng (changes requested):

Barcode-eng (approved):

PR #1286, opened 2026-01-21
feat(automations): add tooltips to disabled conditions

jabloink:
image

## Summary by CodeRabbit

* **New Features**
  * Query builder and workflow UI now show unavailable fields and state options as disabled (with explanatory tooltips) instead of hiding them, so users see why options are unavailable.
  * Workflow conditions reflect instance capability checks, dynamically disabling options based on system capabilities for clearer UX.

### DIFF
--- a/web/src/components/query-builder/DisabledOption.tsx
+++ b/web/src/components/query-builder/DisabledOption.tsx
@@ -50,7 +50,9 @@ export function DisabledOption({ children, reason, className }: DisabledOptionPr
   return (
     <Tooltip>
       <TooltipTrigger asChild>
-        {child}
+         <div>
+           {child}
+         </div>
       </TooltipTrigger>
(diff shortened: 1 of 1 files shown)


web/src/components/query-builder/DisabledOption.tsx
```diff
@@ -13,19 +20,37 @@ interface DisabledOptionProps {
...
-          className={cn(
-            "relative flex cursor-not-allowed select-none items-center rounded-sm text-sm opacity-50 outline-none",
-            className
-          )}
-          role="option"
-          aria-disabled="true"
-        >
-          {children}
-        </div>
+        {child}
```
jabloink: noticed tooltips stopped working on disabled conditions after this change, wrapping in a div seems to allow hover events to fire again

 ```suggestion
          <div>
            {child}
          </div>
  ```

PR #1289, opened 2026-01-22
feat(crossseed): Multiple hard/reflink dirs

rybertm:
Allows multiple, comma separated base directories for hard/reflink. It picks the first path that is able to create the link (same filesystem).

## Summary by CodeRabbit

* **New Features**
  * Hardlink and Reflink modes now accept multiple base directories (comma-separated) and automatically choose an appropriate directory on the same filesystem as the source.

* **Documentation**
  * Guides and examples updated for comma-separated base directory paths and single vs. multi-filesystem setups.

* **UI Updates**
  * Settings label, placeholder, and helper text revised to reflect multi-directory support and clearer examples.

* **Bug Fixes**
  * Improved error messaging when no matching base directory is found.

### DIFF
--- a/internal/services/crossseed/hardlink_mode_test.go
+++ b/internal/services/crossseed/hardlink_mode_test.go
@@ -5,6 +5,7 @@ package crossseed
 
 import (
 	"context"
+	"os"
 	"path/filepath"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


rybertm:
This tries to emulate, at least in theory, what cross-seed does. 

Some alternative approaches:

1- Allow the base directory to be relative. Initially I figured it could be relative to the original torrent's path, but open to suggestions
2- Have some way of setting a different path per category, defaulting to the base directory, which could be either the current single, global base, the currently implemented multiple directories or the relative one suggested above

The simplest would be the approach already implemented in this PR, but I'd say n°2 is the most flexible. Let me know what you guys think

zisso1:
> 1- Allow the base directory to be relative. Initially I figured it could be relative to the original torrent's path, but open to suggestions

@rybertm I think this has huge potential and should be fairly easy to implement: if `dir` doesn't start with `/`, try to combine it with the `sourcePath`.

Setting up a base dir like `cross-seed` would just work for most setups, even when categories are stored in different file systems.

rybertm:
@s0up4200 any input on this? Don't want to change too much in case you guys have something else planned.

s0up4200:
> @s0up4200 any input on this? Don't want to change too much in case you guys have something else planned.

Will try to check it later today 👍

PR #1250, opened 2026-01-16
fix(crossseed): correct match type for episode-from-pack cross-seeding

neelmehta247:
In findBestCandidateMatch, parameters are intentionally swapped when calling getMatchTypeWithReason. This caused returned match types to have inverted semantics - "partial-contains" was returned when an episode file was found in a season pack, but this type had priority 0 and was rejected.

The rejection caused cross-seed attempts to fail with misleading errors like "No matching torrents found" or size mismatches comparing the full pack size to a single episode.

Fix by swapping the partial match types after the call to maintain correct semantics from the caller's perspective:
- "partial-contains" -> "partial-in-pack" (episode in pack, accepted)
- "partial-in-pack" -> "partial-contains" (pack from episode, rejected)

Fixes #1138

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed scoring for partial candidate matches to improve selection accuracy.
  * Ensured match-type handling is consistent when promoting or rejecting candidates, preventing mismatches between evaluation and final assignment.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #1291, opened 2026-01-22
fix(crossseed): valid partial matches being rejected

rybertm:
When testing the cross-seed I found some cases where some candidates appeared to be valid, but were being rejected.

Such a case is listed in the new test.

Let me know if the fix is too permissive, as you guys have dealt with torrents for longer than me.

## Summary by CodeRabbit

* **Tests**
  * Renamed/updated a test to verify matching a single-file source against a multi-file candidate pack, ensuring a match is found and match type recorded.

* **Improvements**
  * In regular mode, partial-contains matches that would leave extra files are now skipped and flagged as requiring hardlink/reflink mode.

* **UI**
  * Dialog shows "Skipped - Requires Hardlink/Reflink Mode" for these skipped matches.


s0up4200 (changes requested):
I think we should only allow partial-contains to apply in regular (reuse) mode when it’s a clean “folder wrapper” with no extra files, because when we’re converting movie/movie.mkv onto an existing rootless /downloads/movie.mkv we do it via contentLayout=NoSubfolder, and any extra files in the incoming torrent (sample/nfo/srr) would then download into /downloads (or /downloads/Sample/...) instead of staying contained under a movie/ folder. That’s messy at best and can cause collisions across torrents.

So I’d add a small guard: if the chosen match is partial-contains, the existing torrent is rootless, the incoming torrent has a top-level folder, and hasExtraSourceFiles(...) is true, then skip in regular mode with a message like “requires hardlink/reflink mode”; but still allow it when hardlink or reflink is enabled (since those modes can preserve layout without spraying extras into the base directory).

neelmehta247:
@s0up4200 i believe this fixes the same underlying condition as the one in #1250.

@rybertm does that PR fix this issue for you?

rybertm:
> @s0up4200 i believe this fixes the same underlying condition as the one in #1250.
> 
> @rybertm does that PR fix this issue for you?

I saw this before applying my fix, but sadly it does not.

rybertm:
> I think we should only allow partial-contains to apply in regular (reuse) mode when it’s a clean “folder wrapper” with no extra files, because when we’re converting movie/movie.mkv onto an existing rootless /downloads/movie.mkv we do it via contentLayout=NoSubfolder, and any extra files in the incoming torrent (sample/nfo/srr) would then download into /downloads (or /downloads/Sample/...) instead of staying contained under a movie/ folder. That’s messy at best and can cause collisions across torrents.
> 
> So I’d add a small guard: if the chosen match is partial-contains, the existing torrent is rootless, the incoming torrent has a top-level folder, and hasExtraSourceFiles(...) is true, then skip in regular mode with a message like “requires hardlink/reflink mode”; but still allow it when hardlink or reflink is enabled (since those modes can preserve layout without spraying extras into the base directory).

That makes sense. I will update the PR with your suggestions later.

bayanoj330-dev (dismissed):

rybertm:
@s0up4200 I reverted the original fix since another commit addressed it and added the check you mentioned. Let me know if it's what you proposed.

rybertm:
Fixed conflicts, ready for review

PR #1349, opened 2026-01-28
feat(cross-seed): add completion indexer targeting

### LINKED ISSUE #877: feat(crossseed): Target indexer available on "Auto-search on completion" as well
**Is your feature request related to a problem? Please describe.**
In the RSS Automation we can pick what Target indexers we like to use

**Describe the solution you'd like**
Have the same option in "Auto-search on completion", as it currently only have cats/tags

This would allow us to pick per client which indexers to include on completion.

s0up4200:
## Summary
- add per-instance completion indexer selection (empty = all enabled)
- persist indexerIds for completion settings + API/schema updates
- update completion UI to select indexers and avoid loading-state warning
- document completion indexer targeting in cross-seed overview

## Testing
- make lint (frontend lint fails due to pre-existing issues in unrelated files)
- make test

Fixes #877

## Release Notes

* **New Features**
  * Added per-instance indexer targeting for cross-seed completion searches, allowing users to restrict automatic searches to specific Torznab indexers instead of using all enabled indexers.
  * Updated completion preferences interface with indexer selection controls.


winterberryteastories:
Does this include selecting indexers for the "dir scan" as well or should I perhaps open a separate issue for that?

s0up4200:
> Does this include selecting indexers for the "dir scan" as well or should I perhaps open a separate issue for that?

Nope. I think there is a separate feature request for it already though.

winterberryteastories:
You're right looks like https://github.com/autobrr/qui/discussions/1331 is basically what I was hinting at.

PR #1371, opened 2026-01-30
feat(notifications): add shoutrrr and notifiarr

s0up4200:
## Summary
- Add Shoutrrr notification targets w/ expanded event coverage (torrents, backups/scans, cross-seed, automations)
- Discord embeds + Notifiarr fields for structured messages and metadata
- Notifications UI grouped by category with accordion + per-group select/clear
- Align shoutrrr dep (v0.13.1), remove local replace, update docs

## Summary by CodeRabbit

* **New Features**
  * Full notifications system (Discord, Notifiarr, Slack, Telegram, Gotify, etc.) with test/send flows and rich payloads
  * Settings → Notifications UI to create/edit/delete targets, select/group event types, and run tests
  * Notifications emitted for torrent completion, backups (success/failure), dir/ orphan scans, cross-seed runs, and automations
  * Public HTTP endpoints to list events, manage targets, and test delivery

* **Database**
  * Migration adds notifications table with updated_at trigger

* **Documentation**
  * Added Notifications feature guide with setup and event descriptions


heathlarsen:
This will be very helpful, excited for it!

Will this also be supported in the Orphan Scan feature so the report isn't only visible from Qui?
It would be very helpful to know we need to run the cleanup if we're too nervous to turn on auto-cleanup.

ppkhoa:
Can you make Discord notification (or notifications in general) content customizable? As of the latest `develop` docker image, this is the notification from qui via Discord:

image

I also check notification via smartwatch and all I see is `Torrent Added` and filename, which is not very clear which instance and the status, either just a cross-seed added (progress 1.00 or 100%) or a new torrent, had to pull out the phone and open Discord to check. Instead of catering to my specific use case, maybe give the users option to customize what the notification look like?

Also, another note, if progress is 1.00, meaning it's already completed => it's a cross-seed, I don't need the other fields like ETA, ratio, amount left, etc..., not useful for notification, which only display static info. Just `Instance`, `Filename`, `Tracker` and `Tags` are sufficient since those info are static. Not sure if other users agree, just my opinion

PR #1431, opened 2026-02-08
fix(web): auto-append slash on path autocomplete selection

### LINKED ISSUE #856: [FR] Auto-add / with path autocomplete dropdown
**Is your feature request related to a problem? Please describe.**
Path do not have full discovery/completion of lower folder without / at the end

**Describe the solution you'd like**
Automatically add / when user clicks with mouse on the suggested folder by QUI, enabling more navigation

**Describe alternatives you've considered**
None

**Additional context**
Adding /  would assume the user want to navigate lower into sub-folder

While this FR would help for navigating it do no help with everything
Example user « stuck » when he write with the keyboard full path without / at the end 

/home/test/folder
Do not get any discovery, user have to manually add the slash at the end /

Thanks a lot

nitrobass24:
## Summary
- When selecting a directory from the autocomplete dropdown, a trailing `/` is now automatically appended
- The suggestion dropdown stays open after selection so subdirectory entries load immediately
- This enables continuous folder navigation without the user needing to manually type `/`

Fixes #856

## Test plan
- [ ] Open the add torrent dialog with path autocomplete enabled
- [ ] Type a partial path and select a folder from the dropdown
- [ ] Verify `/` is appended and subdirectory suggestions appear immediately
- [ ] Verify keyboard selection (Enter/Tab) also appends `/` and shows subdirectories
- [ ] Verify typing a full path without trailing `/` shows the matching folder as a suggestion

## Summary by CodeRabbit

* **Bug Fixes**
  * Autocomplete now normalizes selected paths, adding an OS-appropriate trailing separator (backslash for Windows-style paths, slash otherwise) so both the input and outward selection receive a consistently formed path.
  * Suggestions are no longer dismissed on selection, keeping the list available to reduce accidental loss of focus.


nitrobass24:
Was cleaning up branches on my fork and accidentally removed some I shouldnt have. I will also update these with the base branch to be consistent with the 14.1 release.

s0up4200:
@nitrobass24 what happens if the path is a file?

nitrobass24:
Great question and not something I considered. That said based on my understanding of the project I dont think thats possible. 

When a user hits the drop down, usePathAutocomplete extracts the parent directory and passes it to useDirectoryContent, which calls api.getDirectoryContent(instanceId, path) — a GET to /instances/{instanceID}/getDirectoryContent?dirPath=....

The handler in torrents.go passes this through to SyncManager.GetDirectoryContentCtx(), which calls client.GetDirectoryContentCtx() from the go-qbittorrent library.

 go-qbittorrent library: In methods.go, GetDirectoryContentCtx calls the qBittorrent WebAPI endpoint app/getDirectoryContent with mode=dirs hardcoded:
  opts := map[string]string{
      "dirPath":      dirPath,
      "withMetadata": strconv.FormatBool(withMetadata),
      "mode":         "dirs",
  }
  resp, err := c.getCtx(ctx, "app/getDirectoryContent", opts)

 The mode=dirs parameter tells qBittorrent to only return directories, never files. So every entry in the suggestion dropdown is guaranteed to be a directory, making the unconditional / append safe.

Let me know if my understanding is correct or if you would like changes anyways let me know and ill come up with a safer approach.

PR #1551, opened 2026-03-04
fix(api): restrict CORS to explicit allowlist

s0up4200:
This hardens CORS by removing permissive origin reflection with credentials. CORS is now off by default and is only enabled when an explicit allowlist is configured.

The change adds `corsAllowedOrigins` to `config.toml` and `QUI__CORS_ALLOWED_ORIGINS` for environment-based configuration. Configured origins are normalized and validated as explicit origins only, with wildcard, path, query, fragment, and userinfo values rejected. IDN hosts are canonicalized, default ports are stripped, and invalid startup configuration now fails fast. On live reload, an invalid CORS update keeps the previously valid allowlist.

The API CORS coverage was updated to reflect deny-by-default behavior and allowlist-based preflight and actual request handling, and configuration documentation under `documentation/docs` was updated to match the new behavior.

## Summary by CodeRabbit

* **New Features**
  * CORS is disabled by default; enable by setting QUI__CORS_ALLOWED_ORIGINS with explicit origins (http(s)://host[:port]).

* **Behavior**
  * CORS middleware only activates when origins are configured.
  * Server validates and normalizes configured origins at startup and aborts on invalid values.
  * Dynamic reloads reject invalid CORS changes and preserve the prior valid setting.

* **Documentation**
  * Expanded configuration docs, examples, and SSO-proxy guidance.

* **Tests**
  * Broadened coverage for CORS validation, normalization, and runtime behavior.

* **Chores**
  * Removed CI test job from release workflow.


documentation/docs/advanced/sso-proxy-cors.md
```diff
@@ -18,4 +18,14 @@ When qui is behind an SSO proxy (Cloudflare Access, Pangolin, etc.), expired ses
 - Configure CORS **on the SSO proxy** (not in qui) for the auth endpoints.
 - Allow credentials and handle `OPTIONS` preflight when required.
 
+## Optional qui allowlist
+
+If your browser frontend must call qui from a different trusted origin, set an explicit allowlist:
```
ppfeister: Just a comment on the verbiage -

It doesn't matter if the hostname is different than what the server sees -- it just matters that if I visit `example.com` and `example.com` itself then tries to load `getqui.com` on the userside, that `example.com` be added to this configuration item as the loaded website that's side-loading from qui. So users can use several domains without ever adding them here, they just need to add them if _another webpage is loading data from qui on the user's behalf_ (without using its own authentication).

If that's the intent of this description, then we're gtg

internal/api/server.go
```diff
@@ -277,16 +277,18 @@ func (s *Server) Handler() (*chi.Mux, error) {
...
+		corsMiddleware := cors.New(cors.Options{
+			AllowCredentials: true,
+			AllowedOrigins:   s.config.Config.CORSAllowedOrigins,
+			AllowedMethods:   []string{"HEAD", "OPTIONS", "GET", "POST", "PUT", "PATCH", "DELETE"},
+			AllowedHeaders:   []string{"Accept", "Authorization", "Content-Type", "X-API-Key", "X-Requested-With"},
+			MaxAge:           300,
+			Debug:            false,
+		})
+		r.Use(corsMiddleware.Handler)
+	}
```
ppfeister, line 291: It _may_ be worth breaking this out into two components -
1. Allowed Origins
2. Allowed Origins w/Credentials

Allowed Origins allows other applications to fetch data from qui using xhr/js/etc following normal authentication patterns, whereas Allowed w/Credentials allows the application to use the user's existing session. This breakup allows for the administrator to decide what level of access to grant to the outside application, rather than all-or-nothing.

**That being said...**

Since qui is a single-user application and any connections would likely end up having the same level of access as the sole user, and with consideration of the limited use this may actually see in production, I would just call it a security-aware design choice rather than a hard requirement. Personally, I would probably just push this version for simplicity's sake and keep that consideration at mind.

Breaking it out into two components, as long as we're aware of the implication, may be an over-engineered solution for this particular application.

internal/config/config_test.go
```diff
@@ -387,40 +387,124 @@ func TestApplyDynamicChangesNotifiesOnValidAuthDisabledReload(t *testing.T) {
...
+	}
+
+	var listenerCalls int32
+	cfg.RegisterReloadListener(func(_ *domain.Config) {
+		atomic.AddInt32(&listenerCalls, 1)
+	})
+
+	previous := authReloadSettings{
+		authDisabled:               false,
+		iAcknowledgeThisIsABadIdea: false,
```
ppfeister, line 448: _Love this attribute name_

PR #1670, opened 2026-03-29
fix(automations): hardlink signature grouping

aulterego:
The automations "hardlink signature" group doesn't work as I expect it to. The description says it will "group torrents that share the same physical files via hardlinks", but if I make an automation to tag all hardlink signature groups, nothing is tagged. I would expect all of my torrents that share hardlinked files to be tagged.

It's possible that I have a misunderstanding about how the grouping is supposed to work—and from searching discord other users seem to as well. But assuming I'm not wrong, I think the issue was the following:

`ruleUsesHardlinkSignatureGrouping` only checked action-level GroupIDs (Delete, Category, Move) and the DefaultGroupID. It never scanned condition trees, so `IS_GROUPED`/`GROUP_SIZE` conditions referencing `hardlink_signature` (e.g. in tag actions) silently evaluated to false because `HardlinkSignatureByHash` was never populated.

Additionally, torrents with any outside hardlinks were excluded from signature grouping entirely, preventing grouping for the common case where files are also hardlinked to a media library.

Fix both issues:
- Scan all condition trees (via conditionTreesForRule) for IS_GROUPED/GROUP_SIZE referencing hardlink_signature
- Allow rules without an explicit Grouping config to use built-in group IDs like hardlink_signature
- Include torrents with outside links in signature grouping; the scope is still tracked in ScopeByHash for actions to check independently

I've tested this locally and it fixes my issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * Broad hardlink grouping now includes all torrents with hardlinks and consistently prunes singleton groups.
  * Hardlink expansion (delete/space calculations) now only counts delete-safe duplicates fully inside the client.
  * Rule evaluation properly detects hardlink-signature grouping even in nested or missing grouping configurations.

* **Tests**
  * Added and updated tests covering grouping detection, delete-safe signature behavior, dedupe logic, and context setup.

### DIFF
--- a/internal/services/automations/evaluator.go
+++ b/internal/services/automations/evaluator.go
@@ -65,8 +65,11 @@ type EvalContext struct {
 	// FilesToClear is a map of cross-seed keys to the amount of disk space that will be cleared by the "free space" condition, ensuring we don't double count cross-seeds (current active source)
 	FilesToClear map[crossSeedKey]struct{}
 	// HardlinkSignatureByHash maps torrent hash to its hardlink signature (sorted file IDs joined with ";").
-	// Only populated when includeHardlinks is enabled for FREE_SPACE rules.
+	// Used for hardlink_signature grouping and grouped-condition evaluation.
 	HardlinkSignatureByHash map[string]string
+	// DeleteSafeHardlinkSignatureByHash maps torrent hash to its hardlink signature for torrents whose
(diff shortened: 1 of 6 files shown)


s0up4200:
@aulterego I did a few changes, but your fix should still work, if you want to test

aulterego:
Cool, yep still seems to work!

PR #1695, opened 2026-03-31
fix(dirscan): link plan size tolerance + partial season pack injection in link tree mode

s0up4200:
Fixes link plan failures for size-tolerance matches and adds support for partial season pack injection in hardlink/reflink mode.

**Size tolerance fix**: `buildLinkTreeMatchedFiles` was setting `ExistingFile.Size` to the searchee (disk) file size, but `BuildPlan` indexes by torrent file size. When sizes differ within tolerance the lookup missed. Changed to use `pair.TorrentFile.Size`.

**Partial link tree injection**: When a multi-file torrent partially matches (e.g. a season pack with only some episodes on disk), the link tree now gets built with matched files only. A new `DownloadMissingFiles` setting (default: on) controls whether these partial injections proceed. When enabled, the torrent is added paused, rechecked so qBit discovers existing data, then resumed per the user's `StartPaused` preference. When disabled, partial link tree injections are rejected with a descriptive error.

Aims to solve https://github.com/autobrr/qui/discussions/1397

## Release Notes

* **New Features**
  * Added "Download missing files" setting to Dir Scan configuration. When enabled (default), partial matches like season packs or partial releases can be injected in hardlink/reflink mode. When disabled, such partial matches are rejected.

* **Documentation**
  * Updated Dir Scan and hardlink mode documentation to explain the new "Download missing files" setting and its behavior for partial match handling.


CjE5:
I believe this causes an issue, all of the sudden I get lots of false positives. I get matches for movie torrents for completely unrelated releases. When I re-check all of them check at 0.

CjE5:
Confirmed by reverting the commit locally cc @s0up4200

PR #1731, opened 2026-04-06
fix(automations): include AutoManagement in delete standalone check

nitrobass24:
## Summary
- Adds AutoManagement to the `hasOtherAction` check in `validatePayload` so that Delete + AutoManagement is rejected as invalid

The delete-is-standalone validation already checks SpeedLimits, ShareLimits, Pause, Resume, Recheck, Reannounce, Tags, Category, Move, and ExternalProgram — but AutoManagement was missed when it was added in ba3d5d97.

#1726 fixed the other three validation functions (`conditionsUseField`, `conditionTreesForValidation`, `collectConditionRegexErrors`) but this one was still missing.

## Test plan
- [x] `go test -race -count=1 -v ./internal/api/handlers/...` passes

## Summary by CodeRabbit

* **Bug Fixes**
  * Automation validation updated: Auto Management options can no longer be combined with an enabled Delete action. This prevents conflicting automation setups and reduces risk of unintended deletions; users configuring automations will now see the validation enforced when attempting such combinations.


nitrobass24:
This is a follow on to PR #1726

s0up4200 (approved):

PR #1737, opened 2026-04-07
feat(auth): add PKCE support to OIDC implementation

oynqr:
This PR adds PKCE support to the OIDC implementation, which is by default required by some IDPs like kanidm. The used oauth2 library already has support for PKCE, it just had to be wired up. See https://github.com/autobrr/autobrr/pull/2421 for equivalent changes in autobrr.

I tested this using a production NixOS deployment of qui and the kanidm IDP, with PKCE both enabled and disabled on the IDP side.

## Summary by CodeRabbit

* **New Features**
  * Enhanced authentication with PKCE (S256) support when the identity provider advertises it.
  * Authorization flow now includes a PKCE verifier for improved security, persists it during login, and cleans it up after callback to ensure safe validation.
  * Better compatibility and more secure sign-in experience with providers that support PKCE.

### DIFF
--- a/internal/api/handlers/oidc.go
+++ b/internal/api/handlers/oidc.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


s0up4200 (approved):
LGTM!

PR #1742, opened 2026-04-07
fix(dirscan): tighten flexible matching and improve diagnostics

s0up4200:
Dir scan now requires exact per-file size matches while keeping size tolerance limited to total torrent candidate filtering. It also tightens ambiguous flexible single-file matches by requiring corroborating title or external ID evidence, including preserving IMDb/TVDb/TMDb search context when indexers omit those fields in results.

This prevents fallback title-search false positives from being injected as valid matches and adds clearer match and rejection logging, partial-match unmatched-file samples, and updated dir-scan docs and UI copy so the behavior is easier to understand.

## Summary by CodeRabbit

* **Documentation**
  * Clarified Dir Scan Match Mode semantics and Size Tolerance behavior.

* **New Features**
  * Flexible single-file matches now require corroborating title or external-ID evidence to be accepted.

* **Improvements**
  * Enforced exact per-file size matching across modes; size tolerance now only filters by total torrent size.
  * Search results now include per-search identifier metadata to support corroboration.

* **UI**
  * Updated help text for Match Mode and Size Tolerance controls.

### DIFF
--- a/documentation/docs/features/cross-seed/dir-scan.md
+++ b/documentation/docs/features/cross-seed/dir-scan.md
@@ -169,8 +169,8 @@ Open **Dir Scan > Settings**:
 
 | Setting | Description |
 |---------|-------------|
-| Match Mode | `Strict` matches by filename + size. `Flexible` matches by size only. |
-| Size Tolerance (%) | Allows small size differences when matching. |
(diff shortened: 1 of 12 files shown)


CjE5:
Looks like this might be fixing the issues in here https://github.com/autobrr/qui/pull/1695

Running locally I don't see any false positives anymore.

PR #1761, opened 2026-04-09
fix(orphanscan): ignore qBittorrent incomplete files

### LINKED ISSUE #1759: [FR] Allow to exclude file types instead of full path from the Orphan Scan
### Discussed in https://github.com/autobrr/qui/discussions/1755

<div type='discussions-op-text'>

<sup>Originally posted by **foux** April  9, 2026</sup>
Today we've got an issue with orphan scan : orphan scan will mark Qbit incomplete files (*.!qB) as orphaned, when they are, most of the time, obviously not.

It is possible to either ignore them globally, or by adding the possibility to ignore files by extension from the Orphan Scan.</div>

nitrobass24:
qui now ignores qBittorrent incomplete download files with the `.!qB` suffix during orphan scans. This keeps in-progress downloads out of orphan previews and deletion flow by treating them as built-in transient files alongside other default ignores.

Adds the suffix ignore in the orphan scan walker and covers it with a regression test. Closes #1759.

### DIFF
--- a/internal/api/handlers/orphan_scan.go
+++ b/internal/api/handlers/orphan_scan.go
@@ -68,7 +68,6 @@ type OrphanScanSettingsPayload struct {
 	MaxFilesPerRun      *int     `json:"maxFilesPerRun"`
 	AutoCleanupEnabled  *bool    `json:"autoCleanupEnabled"`
 	AutoCleanupMaxFiles *int     `json:"autoCleanupMaxFiles"`
-	IgnoreQBIncomplete  *bool    `json:"ignoreQbIncomplete"`
 }
 
 // GetSettings returns the orphan scan settings for an instance.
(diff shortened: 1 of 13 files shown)


s0up4200 (commented):
Why not simply extend the `ignoredOrphanFileNameSuffixes` here with `.!qB` and call it a day with that? Is there a good reason this should be optional? 

https://github.com/autobrr/qui/blob/develop/internal/services/orphanscan/walker.go#L39-L41

nitrobass24:
I guess I dont have a strong opinion, but was thinking these typically get created in paths where people do their active downloading, but they were existing in other paths e.g. /cross-seed/* they might want them cleaned up? 

Would be a much simpler implementation do just always have it. No database changes this way.

s0up4200:
> Would be a much simpler implementation do just always have it. No database changes this way.

Yeah I think thats the best approach until someone presents a real demand for something bigger 👍 That way we do a 1-line addition and optionally a regression test case for it.

PR #1801, opened 2026-04-22
feat(actions): include web-dist in release artifacts

turtletowerz:
Previous packages with web interfaces ([autobrr](https://github.com/autobrr/autobrr/blob/develop/.github/workflows/release.yml#L411-L418), [dashbrr](https://github.com/autobrr/dashbrr/blob/develop/.github/workflows/release.yml#L210-L219)) have included a `web-dist` in the release artifacts, which has been helpful in making packages for package managers. Currently qui builds the web distribution but doesn't upload the artifact.

Note that this has not been tested since I don't have the `RELEASE_SIGNING_PRIVATE_KEY` and couldn't get past the build stage.

## Summary by CodeRabbit

* **Chores**
  * The web distribution is now packaged as a compressed tarball and automatically produced by CI.
  * This tarball is included in published release artifacts so downloads provide a ready-made web distribution for easier portability and consumption.

[Review Change Stack](https://app.coderabbit.ai/change-stack/autobrr/qui/pull/1801)


.github/workflows/release.yml
```diff
@@ -165,6 +165,17 @@ jobs:
...
+        if: startsWith(github.ref, 'refs/tags/')
+        run: mkdir -p dist && tar czf dist/web-dist.tar.gz web/dist
+
+      - name: Upload web-dist to release assets
+        uses: softprops/action-gh-release@v2
+        if: startsWith(github.ref, 'refs/tags/')
+        with:
+          files: |
+            dist/web-dist.tar.gz
+
```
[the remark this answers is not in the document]
turtletowerz: that is intentional

.github/workflows/release.yml
zze0s, line 105: It already uploads web-dist here after it has built web?

turtletowerz: i don't see this in any of the artifacts on the release page, is there somewhere else i should be looking?

zze0s: My bad, these are only internal. The web-dist need to be included in the goreleaser file. See the one in autobrr as an example https://github.com/autobrr/autobrr/blob/develop/.goreleaser.yml#L80-L83

zze0s (approved):

PR #1881, opened 2026-05-14
fix(web): keep showing torrent column filter icons on touch devices

martylukyy:
This PR fixes a regression caused by changes in #1869 where wider viewpoints (especially for tablets)
were not able to access the filter icons anymore.

The frontend now checks if the primary input type is touch and still shows the icons if that is the case.

## Summary by CodeRabbit

* **Style**
  * Improved filter icon visibility in table headers. The filter icon now intelligently appears based on hover capability and active filter status, showing on hover when applicable.

[Review Change Stack](https://app.coderabbit.ai/change-stack/autobrr/qui/pull/1881)


nuxencs (approved):
tested locally, lgtm

PR #1912, opened 2026-05-19
fix(crossseed): harden filesystem fallback and post-add rechecks

### LINKED ISSUE #1589: Preserve tracker-isolated destination roots across partial hardlink/reflink fallback and reuse chains
## Summary

When a cross-seed is added using a tracker-isolated hardlink/reflink destination preset, the intended destination root is currently only honored if the hardlink/reflink path fully succeeds for that add.

If a later cross-seed matches against that newly created cross-link tree and hardlink/reflink creation falls back to regular mode, the add path is recomputed from the matched torrent's SavePath instead of preserving the incoming torrent's own tracker-isolated destination root.

This causes multiple cross-seeds that should live in separate tracker-specific directories to be added against the same existing cross-link folder.

## Current behavior

A typical sequence looks like this:

1. A cross-seed is added in hardlink/reflink mode and materialized under a managed tracker-specific directory.
2. A later cross-seed matches against that newly added torrent rather than the original source torrent.
3. Hardlink/reflink creation only partially succeeds, or fails after the system has already determined that partial reuse is otherwise acceptable.
4. The code falls back to regular mode.
5. Regular mode uses the matched torrent's SavePath, which now points at the earlier tracker-specific cross-link directory.
6. The new cross-seed is therefore added into the previous cross-seed's directory instead of its own tracker-specific directory.

The result is that tracker isolation is lost even though the configured directory organization says these torrents should be kept separate.

## Why this is incorrect

The current behavior couples save-path selection to the transfer mode that ultimately gets used.

That is the wrong boundary.

The intended destination root is a storage policy decision for the incoming torrent. Whether files are pre-created via hardlinks, pre-created via reflinks, or left absent until the user later starts the torrent in qBittorrent is a separate execution detail.

If partial hardlinking/reflinking has already been deemed safe enough to proceed, then preserving the incoming torrent's tracker-specific destination root is the more correct behavior:

- Piece-boundary checks already decide when partial reuse is unsafe.
- Size mismatch tolerance already decides when partial reuse should be rejected.
- Files that were not linked/reflinked do not exist yet, so there is no need to collapse the torrent into another tracker's folder just to accommodate them.
- Those missing files will only be created later through explicit user action when the torrent is started.

So once the add is allowed, the torrent should remain rooted in its own managed tracker-specific directory even if only a subset of files were materialized up front.

## Impact

- Tracker-isolated directory organization becomes unreliable in reuse chains.
- Later cross-seeds can inherit the wrong managed destination root.
- Multiple logically separate cross-seeds end up sharing one tracker-specific folder.
- This makes on-disk layout harder to reason about and breaks the expectation that directory organization reflects the incoming torrent's tracker grouping.

## Stronger fix

Instead of deriving the destination root separately inside hardlink mode, reflink mode, and regular fallback, compute the incoming torrent's managed destination root once and carry it through the entire add flow.

Suggested approach:

1. Determine the intended managed destination root for the incoming torrent before selecting hardlink/reflink/regular execution.
2. Treat that destination root as authoritative for the add, independent of which execution mode ultimately succeeds.
3. Use the same root for:
   - hardlink plan creation
   - reflink plan creation
   - fallback regular-mode savepath
   - any later alignment/rename logic that assumes where the torrent should live
4. If a torrent is being added into a managed cross-link tree, do not replace that root with the matched torrent's raw SavePath just because the code path changed modes.
5. Preserve tracker isolation whenever partial materialization has succeeded or when the torrent is otherwise allowed to proceed under existing reuse-safety checks.

Conceptually, the system should distinguish between:

- intended destination root for the incoming torrent
- matched source location used to discover reusable bytes

Those are not the same thing and should not be conflated.

## Expected behavior

Once a cross-seed qualifies for tracker-isolated storage, it should always be added against its own managed tracker-specific root, even if:

- it matched against another already-added cross-seed,
- only a subset of files were hardlinked/reflinked,
- or the add path falls back from hardlink/reflink mode to regular mode.

Missing files can remain absent until the user explicitly starts the torrent.

Audionut:
This branch hardens cross-seed search/apply behavior, with most of the work focused on making match selection, duplicate handling, and post-add recheck/resume flows more deterministic.

- Adds explicit cross-seed search result statuses (`added`, `skipped`, `failed`) and updates the API/UI so skipped results are no longer mixed together with real failures.
- Adds layout-aware content prefiltering for existing local torrents before search results are applied. This helps pre-searching by narrowing indexer choices and identifying already-seeded content earlier, so qui can avoid spending work on results that are known to be invalid for the selected source. When a later search result maps back to a prefiltered duplicate, qui can report a clear `size_mismatch` or `content_mismatch` instead of downloading/applying it and failing generically.
- Adds ranked cross-seed add plans. Instead of taking the first plausible match, qui now scores candidates by match strength and layout cost: exact/stronger matches win, then plans needing less root/file rename alignment win, then better folder-shape/file-count choices break ties. This should reduce unnecessary renames/rechecks and pick the candidate most likely to add cleanly.
- Preserves explicit `0%` size tolerance overrides, so strict matching can be requested without being mistaken for “use default tolerance”.
- Improves hardlink/reflink fallback safety by distinguishing materialized vs unmaterialized files, enforcing piece-boundary checks for risky fallback cases, and skipping below-threshold partial materializations before adding.
- Makes post-add recheck/resume more reliable by waiting for qBittorrent to expose newly added torrents, avoiding recheck while `checking_resume_data` is still active, retrying resume, and confirming the torrent stays running after files-checked stop conditions settle.
- Tightens Jackett scheduler accounting so long-running/canceled searches do not incorrectly free rate-limit timing too early.
- Improves Windows filesystem handling for hardlink/reflink decisions by resolving symlinked paths and comparing volume identity rather than relying only on path roots.
- Updates docs around cross-seed statuses, recheck requirements, hardlink/reflink fallback behavior, and bad-hash/missing-file troubleshooting.
- Adds broad regression coverage for add-plan ranking, duplicate filtering, content prefiltering, recheck/resume behavior, scheduler behavior, dirscan injection, and Windows filesystem helpers.

## Summary by CodeRabbit

* **New Features**
  * Per-request size-mismatch tolerance (including explicit zero), explicit per-result status (added/skipped/failed), and support for dual infohash fields surfaced in UI/API.
* **Bug Fixes**
  * More reliable post-add recheck/resume with bounded retries, readiness checks, and correct behavior on timeouts/cancellation.
* **Refactor**
  * Improved resume/recheck state tracking, matching/plan selection, and unified retry/rate-limit behavior.
* **Documentation**
  * Clarified fallback, recheck and resume semantics for hardlink/reflink/link modes.
* **Tests**
  * Vastly expanded coverage for retries, tolerances, thresholds, resume and fallback flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/autobrr/qui/pull/1912?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nitrobass24:
Found 2 issues:

1. `bulkActionSyncRetry` now derives its context from the caller (`ctx`) instead of `context.Background()`. Commit 0cb682c9 introduced `context.Background()` explicitly to prevent caller cancellation from aborting this critical sync retry. With the post-add path using 12 attempts at 250ms intervals (3s minimum sleep + sync latency), the shared 5s `bulkActionSyncRetryTimeout` applied via `context.WithTimeout(ctx, ...)` can be further shortened by an already-expiring caller context (e.g., HTTP handler timeout), silently aborting retries before all 12 attempts complete.

https://github.com/autobrr/qui/blob/a1a6efb32ca176fec525c20c58607ae8fdf60011/internal/qbittorrent/sync_manager.go#L70-L75

https://github.com/autobrr/qui/blob/a1a6efb32ca176fec525c20c58607ae8fdf60011/internal/qbittorrent/sync_manager.go#L1984-L1993

2. When `linkFallbackRequiresFullRecheck` is true and `alignmentSucceeded` is false, `needsRecheck` evaluates to false. The torrent was already added in paused+stopped state (L4701-4702) but no recheck is queued and no resume is scheduled, leaving it stuck paused indefinitely with no log explaining why. Alignment failure is possible when `alignCrossSeedContentPaths` fails on the fallback path.

https://github.com/autobrr/qui/blob/a1a6efb32ca176fec525c20c58607ae8fdf60011/internal/services/crossseed/service.go#L4690-L4702

https://github.com/autobrr/qui/blob/a1a6efb32ca176fec525c20c58607ae8fdf60011/internal/services/crossseed/service.go#L4966-L4970

Audionut:
~~This invalidates the per torrent logs from recent runs history.~~

~~Given that recent runs only covers 10 past runs, I'm not sure if adding a small decoding helper is worth the code.~~

edit: added a small helper. leaving the breaking change flag just in case anyone is getting result history via the api.

PR #1914, opened 2026-05-19
feat(fsops): add Backend interface, local implementation, and pool resolver

nitrobass24:
## Summary

- `internal/fsops/backend.go` — `Backend` interface (12 methods) abstracting filesystem operations for local and future remote backends
- `internal/fsops/types.go` — value types (`FileInfo`, `LstatInfo` with degraded-identity `FileIDErr`, `WalkEntry`, `WalkOptions`, `StatfsResult`, etc.)
- `internal/fsops/errors.go` — sentinel errors (`ErrNoFilesystemAccess`)
- `internal/fsops/noop.go` — noop backend for instances without filesystem access
- `internal/fsops/pool.go` — pool resolver (instance ID → local or noop backend)
- `internal/fsops/local/` — local backend implementation + platform-specific Statfs
- `docs/remote-backend-design.md` — SSH/SFTP-native remote backend design (supersedes #1913's deployable-agent approach); linked from `docs/architecture.md`

Surface is intentionally minimal: every method has a caller in the callsite migration (#1915). Batch/diagnostic methods dropped per review return with the remote backend PR that consumes them.

No callsite changes — services still use `os.*` directly. This is the foundation that subsequent PRs will wire into services.

## Test plan

- [x] `go build ./...` and `GOOS=windows go build ./internal/fsops/...` — clean
- [x] `go test -race -count=1 ./internal/fsops/...` — all pass
- [x] `make precommit` — clean

## Summary by CodeRabbit

* **New Features**
  * Adds local filesystem access with per-instance backend selection and unavailable-access handling.
  * Supports metadata inspection, directory traversal, disk-space reporting, filesystem capability detection, and hardlink/reflink tree operations.
* **Bug Fixes / Diagnostics**
  * Provides clearer errors when filesystem access is unavailable or operations are canceled.
* **Documentation**
  * Documents filesystem backend architecture and planned remote access support.
* **Tests**
  * Adds comprehensive coverage for filesystem operations, traversal, tree handling, pooling, and cancellation.


internal/fsops/backend.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// --- Diagnostic ---
+
+	// Info returns metadata about this backend (kind, version, capabilities).
+	Info(ctx context.Context) (*BackendInfo, error)
+
+	// HealthCheck returns nil if the backend is operational, or an error
+	// describing the problem.
+	HealthCheck(ctx context.Context) error
+}
```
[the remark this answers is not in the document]
nitrobass24: The design doc explicitly documents this as a deliberate exception to the ≤5 method rule. The interface represents a single abstraction (filesystem operations for one instance) and every implementation must satisfy all methods. Splitting into composed sub-interfaces would add complexity without enabling callers to depend on narrower types -- services need the full Backend, not fragments.

internal/fsops/local/local.go
```diff
@@ -0,0 +1,328 @@
...
+				return fs.SkipAll
+			}
+
+			if opts.MaxEntries > 0 && count >= opts.MaxEntries {
+				return fs.SkipAll
+			}
+
+			// Skip hidden files/dirs if requested.
+			name := d.Name()
+			if opts.SkipHidden && len(name) > 0 && name[0] == '.' && path != root {
```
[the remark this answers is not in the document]
nitrobass24: filepath.WalkDir never passes a nil fs.DirEntry to the callback. The d parameter is nil only when the root itself can't be stat'd, in which case Go passes the error as walkErr and d is the root entry (not nil). Adding a nil check for something the stdlib guarantees won't happen is the kind of defensive code AGENTS.md says to avoid.

internal/fsops/pool.go
```diff
@@ -0,0 +1,56 @@
...
+	instance, err := p.instanceStore.Get(ctx, instanceID)
+	if err != nil {
+		return nil, fmt.Errorf("load instance %d: %w", instanceID, err)
+	}
+	if instance == nil {
+		return nil, fmt.Errorf("instance %d not found", instanceID)
+	}
+
+	if instance.HasLocalFilesystemAccess {
+		return p.local, nil
```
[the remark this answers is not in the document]
nitrobass24: fixed in [85117b5](https://github.com/autobrr/qui/pull/1914/commits/85117b5a07078023fade01248325501ed8db6887)

internal/fsops/noop.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/autobrr/qui/pkg/hardlinktree"
+)
+
+// noopBackend implements Backend by returning ErrNoFilesystemAccess for every
+// operation. Used for instances that have no filesystem access configured.
+// Context cancellation takes precedence over ErrNoFilesystemAccess so callers
+// get the correct error when a request is cancelled.
+type noopBackend struct{}
+
+var _ Backend = noopBackend{}
```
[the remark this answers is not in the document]
nitrobass24: Deliberate exception, per the design doc and the earlier round of this same suggestion on `backend.go`: `Backend` is a single abstraction boundary — it defines the exact operation surface a remote (SSH-backed) implementation must provide, and every implementation (local, noop, remote) must satisfy all of it. Splitting it into composed domain interfaces would thread interface-set plumbing through the pool and every consumer in the dependent callsite PRs, with no behavioral gain and no partial implementation to justify the seams. If one ever becomes real (e.g., a read-only remote backend), that's the point to split without breaking callers.

internal/fsops/local/local.go
```diff
@@ -0,0 +1,374 @@
...
+					entry.Mode = fi.Mode()
+					entry.IsDir = fi.IsDir()
+					entry.IsSymlink = fi.Mode()&os.ModeSymlink != 0
+
+					if (opts.WantFileID || opts.WantNlinks) && fi.Mode().IsRegular() {
+						fid, nlinks, fidErr := hardlink.GetFileID(fi, path)
+						if fidErr != nil {
+							// The caller asked for identity data; an entry without it
+							// must read as unreadable, not as a file with no links.
+							entry.Err = fidErr
```
com6056: A `GetFileID` failure here surfaces as `entry.Err`, and the consumers in the callsite branch treat any non-permission entry error as fatal for the whole walk, so a Windows-only identity failure (ACL denial, delete-pending file) now kills a scan that previously survived it, with no test covering that path. Same overload in `osFileInfoToLstat` (the return at 367), where a `FileID` failure fails the whole `Lstat` even for callers that only want size and mtime. I'd split identity failure out: a `FileIDErr` field on `WalkEntry` plus a `WantFileID` opt-out on `Lstat`/`LstatBatch` like `WalkOptions` already has. An SFTP backend can't produce `FileID` at all, so the interface needs the graceful path regardless.

internal/fsops/local/local.go
```diff
@@ -0,0 +1,374 @@
...
+		return nil, err
+	}
+	created, err := reflinktree.Create(plan)
+	if err != nil {
+		return nil, err
+	}
+	return treeCreateResult(created, plan), nil
+}
+
+func (b *Backend) RemoveTree(ctx context.Context, created *fsops.TreeCreateResult) error {
```
com6056, line 239: `RemoveTree` removes only what the handle recorded, so a pre-existing plan root stays behind. The current callsites know that and prune it themselves with `backend.Remove`, but nothing here says callers own root pruning, worth a sentence so the next caller doesn't assume it's covered.

internal/fsops/types.go
```diff
@@ -0,0 +1,95 @@
...
+// StatfsResult holds filesystem space information.
+type StatfsResult struct {
+	BytesAvailable int64
+	BytesTotal     int64
+}
+
+// RemoveOptions controls the behavior of a Remove call.
+type RemoveOptions struct {
+	Recursive   bool
+	IgnorePaths []string
```
com6056: `IgnorePaths` has no producers and the local backend refuses it, only when `Recursive` is set though, so a non-recursive `Remove` silently ignores it (`local.go:281`). `RequestID` is written and read by nobody. I'd drop both until a backend needs them.

internal/fsops/backend.go
```diff
@@ -0,0 +1,99 @@
...
+	LstatBatch(ctx context.Context, paths []string) ([]*LstatInfo, []error, error)
+
+	// ReadDir returns directory entries. If maxEntries > 0 the result is
+	// truncated and the bool return is true.
+	ReadDir(ctx context.Context, path string, maxEntries int) ([]DirEntry, bool, error)
+
+	// WalkDir walks a directory tree and streams entries on the returned channel.
+	// The channel is closed when the walk completes, is cancelled via ctx, or
+	// hits an unrecoverable error. Callers must drain the channel.
+	WalkDir(ctx context.Context, root string, opts WalkOptions) (<-chan WalkEntry, error)
```
com6056, line 51: Normal completion, `MaxEntries` truncation, and ctx cancellation all just close the channel, so truncation is invisible to callers (`ReadDir` signals it explicitly). Worth adding a truncation signal. The "callers must drain" sentence above could also say "or cancel ctx", the callsite branch has two spots that drop the channel without doing either.

internal/fsops/backend.go
```diff
@@ -0,0 +1,99 @@
...
+	// (same device ID on Unix, same volume serial on Windows).
+	SameFilesystem(ctx context.Context, p1, p2 string) (bool, error)
+
+	// FileID returns the unique file identity and link count for path.
+	FileID(ctx context.Context, path string) (hardlink.FileID, uint64, error)
+
+	// --- Write (mutating) ---
+
+	// MkdirAll creates a directory and all parents. Equivalent to os.MkdirAll.
+	MkdirAll(ctx context.Context, path string, perm fs.FileMode) error
```
com6056, line 65: Add a sentence steering callers to `fsutil.ContentDirMode` / `LinkTreeBaseDirMode`. The callsite migration is exactly when someone will hand-type a `0o755` (#1704/#2086).

internal/fsops/types.go
```diff
@@ -0,0 +1,95 @@
...
+	Name      string
+	IsDir     bool
+	IsSymlink bool
+	Mode      fs.FileMode
+}
+
+// LstatInfo holds metadata from an Lstat call, including hardlink identity.
+type LstatInfo struct {
+	FileInfo
+	FileID hardlink.FileID
```
com6056, line 40: `hardlink.FileID` has different fields per platform (`Dev`/`Ino` on unix, `VolumeSerialNumber` plus a 16-byte `Identifier` on Windows), so `LstatInfo` can't serialize across a mixed-OS pair. Pick an opaque wire form now, while the interface is still unshipped and it's free.

com6056 (commented):
Merge is clean against current develop and the #2287 rollback adaptation is correct. Nothing below is architectural.

internal/fsops/local/local_test.go
```diff
@@ -0,0 +1,506 @@
...
+			cancel()
+			break
+		}
+	}
+	// Drain remaining entries (channel should close promptly).
+	for range ch {
+		count++
+	}
+	// Should have stopped well before 50+1 entries.
+	assert.Less(t, count, 52)
```
com6056: This assertion holds no matter what the walk does: the tree is 51 entries so `count < 52` always holds, and the 64-deep buffer lets the whole walk finish before the first receive. Assert `ctx.Err()` plus a count meaningfully below the total, or use a tree bigger than the buffer. Same shape at 261, `LessOrEqual(count, 3)` passes on zero entries. `SkippedExists` also has no coverage yet.

internal/fsops/noop.go
```diff
@@ -0,0 +1,80 @@
...
+func (noopBackend) Remove(ctx context.Context, _ string, _ RemoveOptions) error {
+	return noopErr(ctx)
+}
+func (noopBackend) HardlinkTree(ctx context.Context, _ *hardlinktree.TreePlan) (*TreeCreateResult, error) {
+	return nil, noopErr(ctx)
+}
+func (noopBackend) ReflinkTree(ctx context.Context, _ *hardlinktree.TreePlan) (*TreeCreateResult, error) {
+	return nil, noopErr(ctx)
+}
+func (noopBackend) RemoveTree(ctx context.Context, _ *TreeCreateResult) error {
```
com6056: The interface doc says a nil handle is safe and the local backend returns nil, but noop errors here, so a defensive `RemoveTree` call on an instance without filesystem access gets a spurious error. Returning nil for the nil handle (after the ctx check, so cancellation keeps precedence) fixes it, and `pool_test.go:160` asserts the current behavior so it needs updating with it.

s0up4200:
took a deep pass over the whole stack today. the core here is solid, error passthrough keeps errors.Is working and the rollback handle model lines up with #2287. but the contract needs a revision before this merges.

Lstat and WalkDir make a GetFileID failure fatal with no opt-out (local.go:361), thats what turns one bad file into a whole-scan failure in the callsite branch. identity needs to degrade, not abort.

ReadDir calls Info() on every entry just to fill Mode, and nothing in #1915 ever reads it. thats a per-entry lstat plus a new abort path develop didnt have, a directory with r but not x listed fine before and errors now.

after the full migration in #1915, StatBatch, LstatBatch, FileID, Info and HealthCheck have zero callers, same for ErrConnectionLost, ErrPathNotAllowed, BackendInfo, WalkOptions.MaxEntries and the RemoveOptions fields. drop them until a backend actually needs them, 17 methods is a lot of surface for one implementation.

the godoc also cites a design doc and a "Stage C" that dont exist in this repo, and docs/architecture.md needs an entry for a module every service now depends on.

internal/fsops/local/local.go
nitrobass24: 8d962b57 — identity failure now lands in a new LstatInfo.FileIDErr instead of entry.Err, on both WalkDir and Lstat, so it can't abort a scan. Since WalkEntry embeds LstatInfo, the one field covers both paths. I skipped the WantFileID opt-out on Lstat — the identity attempt is cheap locally and callers that don't care can just ignore FileIDErr; the SFTP backend gets its graceful path from the same field. All the FileID consumers on #1915 already guard FileID.IsZero(), so degradation needed no callsite changes.

internal/fsops/local/local.go
nitrobass24: Doc'd in 0cc22644 — the interface comment now says a pre-existing plan root is not removed and callers own pruning it.

internal/fsops/types.go
nitrobass24: Both dropped in 3b9d69b6, along with the rest of the caller-less surface soup flagged. RemoveOptions is just Recursive now, so the half-enforced combination is gone.

internal/fsops/backend.go
nitrobass24: dropped in 3b9d69b6 — with no cap, the only early close left is a cancellation the caller initiated, so there's no truncation to signal. Doc now reads "drain the channel or cancel ctx" (0cc22644), and you were right about the two spots on #1915 — dirscan's searchee scan and orphanscan's in-use check both returned mid-loop without either. Fixed there with cancel+drain (b7f4ddfb).

internal/fsops/backend.go
nitrobass24: the doc now points at fsutil.ContentDirMode/LinkTreeBaseDirMode with the #1704/#2086 references.

internal/fsops/types.go
nitrobass24: I think this is a bigger discussion. I added an open question in docs/remote-backend-design.md.

hardlink.FileID is shared with crossseed on develop, so the change ripples beyond fsops either way, and nothing serializes it until the remote backend exists.

internal/fsops/local/local_test.go
nitrobass24: 9c46f8d6 — the cancellation test now walks 150 files against the 64-entry buffer and bounds the post-cancel count below what an ignored cancel would produce. The MaxEntries test went away with the feature. SkippedExists is covered now too, including that RemoveTree of the idempotent second result leaves the first call's links alone.

internal/fsops/noop.go
nitrobass24: ad15417b — nil handle returns nil after the ctx check, and the pool test now asserts both directions (nil handle succeeds, non-nil still gets ErrNoFilesystemAccess).

nitrobass24:
Closed 1913 since the code isnt needed for now with the current approach and added docs/remote-backend-design.md to this PR

internal/fsops/types.go
nitrobass24: Coming back to this: you're right on the shape, we'll go opaque, but at remote-backend time rather than in this PR.

Two things decided it for me. First, packing into the existing struct doesn't actually work: unix Dev+Ino is 16 bytes, but a Windows remote's identity is a volume serial plus a 16-byte file ID — 24 bytes. There's no lossless packing in either direction, so the opaque form is the only representation that covers a mixed-OS pair at all. Second, the "free now" part doesn't work, because hardlink.FileID is already on develop (crossseed, orphanscan, dirscan, automations), so the ripple is the same size now or later. Since nothing serializes identity until the remote backend exists, soup's drop-until-needed standard picks the timing.

The migration stays cheap whenever it lands: consumers only do ==, map keys, and IsZero(), all of which survive a comparable fixed-size form ([24]byte-ish) unchanged — Bytes() basically previews it, dirscan already keys on it. The churn is the per-platform constructors and test literals.

I updated docs/remote-backend-design.md to record this as a decision instead of an open question.

docs/remote-backend-design.md
```diff
@@ -0,0 +1,171 @@
...
+cross-instance comparisons must check same-backend first.
+
+## Exec Conventions
+
+- `LC_ALL=C` and NUL separators everywhere; torrent filenames can contain
+  nearly anything except NUL and `/`.
+- Probe GNU vs BSD userland at connect (`find -printf` and `stat -c` are
+  GNU-only); degrade the affected ops per-tool on BSD remotes.
+- Every exec carries a timeout and honors ctx cancellation; output is
+  size-capped.
```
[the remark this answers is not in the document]
nitrobass24: Addressed the specifiable half in 61cc6a1f — new "Path and Command Safety" section: torrent/API-derived paths validated under the same boundary rules the local side enforces (no leading separator, no `..`, no drive-letter/UNC where relative is required; absolute paths only as instance-level roots), exec lines built with strict quoting and `--`-terminated options for `find`/`stat`/`xargs`/`rm`/`cp`, never raw-path interpolation into a shell string.

The implement-and-test half has no target yet — this PR contains no SFTP or exec code; that's the remote-backend PR. The section puts the obligation there explicitly, including the test matrix you listed (traversal payloads, option-like filenames like `-rf`, shell metacharacters). For the local path that does exist today, boundary validation is already in #1915 (`buildFullPath` rejects POSIX and Windows escapes).

nitrobass24:
@com6056 - Can you rereview and close out resolved items? 
@Audionut  - Can you run your review scripts on this please?

Audionut:
## Findings

- **High — host-dependent path semantics.** [`internal/fsops/backend.go`](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/internal/fsops/backend.go#L14-L22) leaves path manipulation to host `filepath`, while [the remote-backend design](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/docs/remote-backend-design.md#L97-L107) requires slash-delimited remote paths. On Windows, `filepath.IsAbs("/data")` is false, so Windows-hosted qui cannot safely operate Unix remotes. Define host-independent backend path semantics before callsite migration.
- **Medium — cancellation cannot interrupt mutations or rollback them.** [`internal/fsops/local/local.go`](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/internal/fsops/local/local.go#L203-L243) checks cancellation only before recursive removal and tree operations. Mid-operation cancellation continues mutations; reusing the cancelled context for `RemoveTree` then prevents rollback.
- **Medium — portable error semantics cover only `Stat`.** [`internal/fsops/backend.go`](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/internal/fsops/backend.go#L24-L47) standardizes `fs.ErrNotExist` only for `Stat`. Remote callers also depend on portable not-found and permission errors from `Lstat`, `WalkDir`, and `Remove`; define mappings and contract tests.
- **Medium — SFTP-only hardlink conflict handling is undefined.** [`docs/remote-backend-design.md`](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/docs/remote-backend-design.md#L30-L38) advertises hardlink-tree creation despite unavailable inode/nlink identity. Existing targets need verification to distinguish an idempotent same-link from conflicting content.
- **Low — oversized interface.** [`internal/fsops/backend.go`](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/internal/fsops/backend.go#L22-L79) has 12 methods, violating the repo's ≤5-method rule. Split consumer-specific interfaces.

## Checks

Targeted race tests, `go build ./...`, diff check, and Linux/macOS/FreeBSD cross-builds passed. `make lint` found zero branch issues but failed on the unchanged [`pkg/sharedextents/sharedextents_windows_test.go`](https://github.com/autobrr/qui/blob/0a5dff03c685ea5d7d4f7794413cd39f0e1dc56c/pkg/sharedextents/sharedextents_windows_test.go#L145), which still uses `reflinktree.Create`'s old signature.

internal/fsops/local/local.go
com6056: `FileIDErr` is exactly the split I was after, and having `Stat` and `Lstat` degrade the same way rather than only an opt-out is a better answer than what I suggested. One follow-on about who consumes it in the summary comment.

internal/fsops/local/local.go
com6056: That's the sentence, and putting it right after the #2282 rationale makes the narrow scope read as deliberate instead of like an oversight.

internal/fsops/types.go
com6056: Both gone, and `RemoveOptions` being just `Recursive` is much easier to reason about. Separate thought for later, not blocking: `WalkOptions.IgnorePaths` is in a similar spot now, orphanscan still runs its own `isIgnoredPath` on every file entry with case folding and a separator boundary, so the exact `path == ignored` match here is a weaker duplicate of it.

internal/fsops/backend.go
com6056: Dropping `MaxEntries` makes the truncation signal moot, which beats adding one. The "or cancel ctx" wording landed too, and both spots I found now cancel the walk and drain it in a defer, so those are closed. The orphanscan walker at `walker.go:217` is the odd one out, it drains but never cancels, so it doesn't leak, it just runs the walk to completion on an early return. Fine as is, a `context.WithCancel` next time someone's in there.

internal/fsops/backend.go
com6056: The `fsutil.ContentDirMode` / `fsutil.LinkTreeBaseDirMode` pointer is there with the issue numbers attached, so anyone hand-typing `0o755` during the migration has to ignore the doc to do it.

internal/fsops/types.go
com6056: Opaque comparable fixed size is the right call. One correction to the doc's rationale for deferring plus a full encoding proposal are in the summary comment, so they don't get lost when this thread resolves.

internal/fsops/local/local_test.go
com6056: 150 files against the 64 buffer does it. I flipped off both the `ctx.Err()` check and the `<-ctx.Done()` send guard locally and the test fails with 151 not less than 100, so it's really asserting something now. `SkippedExists` got covered too, with the assertion that removing the second empty result leaves the first call's link alone.

internal/fsops/noop.go
com6056: Nil returns nil after the ctx check, and having `pool_test.go` assert both the nil and non-nil case keeps it that way. `TestNoopBackend_AllMethodsError` is a slightly odd name for it now, but not worth a commit of its own.

com6056:
Re-review done, all eight of my threads check out and I've replied on each. Resolving is gated to you or a maintainer, so close them out whenever. I ran the round through build, vet, the fsops tests, and windows plus darwin cross-builds on the head and it's all clean, and the new cancellation test really does fail if you disable the guards, so it earns its keep.

One code item, and it's small either way: `Stat`'s followed-target identity has zero consumers. `dirscan/fileid_index.go:87` and `scanSingleFile` on #1915 still call `Lstat`, which is exactly the symlink regression soup flagged there, and those are the only two `Lstat` callsites whose semantics moved vs develop (the rest mirror `os.Lstat`), so wiring those two to `Stat` closes it completely, the API and its test are already sitting there. The alternative is making identity opt-in on `Stat` the way `WalkOptions.WantFileID` works, because right now every `Stat` caller is an existence or `IsDir` check, and on Windows the missing-files scan pays a `CreateFile` per file for a `FileID` nobody reads. `GetFileID` opens a handle for two `GetFileInformationByHandleEx` calls, and develop's `missing_files.go` was a bare `os.Stat` that never went near the fileid path, so it's new cost rather than the cost develop already paid.

On `docs/remote-backend-design.md`, the architecture and the tier model match what we agreed and the op mapping reads accurately against the slimmed interface. Nothing below is a merge gate, these are doc lines I'd add or decide now while it's cheap, from the security side I signed up for:

- Symlink policy for destructive remote ops. Path validation stops `..` but a symlinked directory bypasses it entirely: an SFTP-tier recursive delete that walks with `stat` instead of `lstat` descends the link and deletes outside the tree, while the exec tier's `rm -rf --` doesn't descend, so the two tiers would also diverge. Commit to `lstat`-driven recursion, never descend a link, remove the link itself only, and state the tier-parity requirement.
- Host-key change after pinning. First contact is handled well, but a later mismatch is the case that matters: fail closed, no automatic re-pin, show both fingerprints, require the same explicit confirmation as first contact, and never fall back to TOFU because the pin is missing.
- The pinned fingerprint should get the same AAD binding as the private key when the schema lands in #1917. The fingerprint decides who you talk to, and as a plain column a DB write can downgrade the pin, which makes the key encryption moot.
- Path semantics, +1 to Audionut's first finding: the doc requires slash-delimited remote paths while `backend.go:17` leaves path manipulation to host `filepath` in service code, and on a Windows host `filepath.IsAbs("/data")` is false and `Join` inserts backslashes. That's the one real disagreement between the doc and the shipped interface, and it's cheapest to decide before the callsite migration bakes it in.
- Small one: the frontend section wants a degraded-mode indicator on SFTP-only instances, but capabilities are probe-per-connect and deliberately not persisted, so nothing says where the indicator reads from between SSH tests. The pool owning live capabilities behind an instance-status endpoint would close it.
- On the wire-form section: the doc defers the opaque type because nothing serializes identity yet, but dirscan already persists it. `dir_scan_files.file_id` is a BLOB of `FileID.Bytes()` with a unique index on `(directory_id, file_id)`, and rename detection matches on it, so the encoding change needs a migration or a version byte whenever it lands, which is an argument for pinning the shape now. I'd make it tagged, something like `struct { kind uint8; raw [24]byte }` with kind 0 for no identity, 1 for unix `dev`+`ino` big endian, 2 for a Windows volume serial plus the 16-byte file id, 3 for a remote-provided blob. The tag is the hard-to-retrofit part, untagged bytes let a unix `dev=1,ino=2` collide with a Windows identity in the same space, and it keeps `IsZero()` meaning "no identity" rather than "a backend handed us zeros". 24 is the floor not the target, and `Bytes()` should return the tagged form too, since dirscan keys its index on `string(FileID.Bytes())` and that key has to stay collision-free across a mixed pair. Related, the same-backend comparison guard is prose-only right now, and `==` across two hosts compiles, runs, and is only wrong in production, so I'd put the scope in the value or send comparisons through a helper that takes it.

The doc ships with the PR and can take amendments after, so none of that holds up the merge. The security ones are my lane anyway, so rather than leaving them as homework I'll post suggested edits on the doc file for the symlink policy, the host-key-change behavior, the fingerprint AAD note, and the FileID encoding, and you can accept or rework them. The path-semantics one I'm leaving as a question since that's a contract call for you and soup, not wording.

docs/remote-backend-design.md
```diff
@@ -0,0 +1,194 @@
...
+  under the same boundary rules the local backends enforce: slash-delimited,
+  no leading separator, no `..` component, no drive-letter or UNC form where
+  a relative path is required. Absolute remote paths are permitted only as
+  instance-level roots (e.g. save paths reported by qBittorrent); torrent
+  file names always join beneath one.
+- Exec command lines never interpolate raw paths into a shell string:
+  arguments are strictly quoted, and options are terminated with `--`
+  before any path for `find`, `stat`, `xargs`, `rm`, and `cp`.
+- The remote-backend PR carries tests for traversal payloads, option-like
+  filenames (`-rf`), and shell metacharacters in paths.
```
com6056: The symlink rule, the honest version: the exec and local tiers are race-safe, SFTP v3 can't be, so it says so and bounds it.

```suggestion
- Symlink policy for destructive ops: recursion never follows a link, on
  any tier. A symlinked directory is never descended; the link itself is
  removed, never the target's contents. `rm -rf --` and the local
  backend's `os.RemoveAll` both traverse `openat`-style, so they refuse
  links and are immune to a mid-walk swap. SFTP v3 gets only the first
  half: no `openat`, no `O_NOFOLLOW`, every path re-resolved server-side,
  so a swapped ancestor directory can redirect an unlink between the walk
  and the remove. That race is unfixable over v3; bound it by
  re-`lstat`ing immediately before each unlink. `RemoveTree` is a path
  list, not a walk, and takes the same shape per entry: unlink only,
  directories only when empty.
- The never-follow rule covers identity as well as traversal: identity
  feeding a destructive decision comes from `lstat`, so a planted symlink
  cannot attribute another file's identity to a path inside the tree.
- The remote-backend PR carries tests for traversal payloads, option-like
  filenames (`-rf`), shell metacharacters in paths, and symlinked
  directories inside a tree marked for recursive removal.
```

docs/remote-backend-design.md
```diff
@@ -0,0 +1,194 @@
...
+## Security
+
+- Dedicated SSH key per instance; never the user's personal key.
+- Private key stored AES-GCM encrypted with AAD binding (instance id +
+  field), same `sessionSecret` pattern as existing credential encryption.
+- Host key verification is TOFU with explicit confirmation: the first-seen
+  key is held ephemeral and surfaced as a fingerprint via the ssh-test
+  flow; it is persisted and enforced only after the user confirms it (or
+  it matches a preconfigured fingerprint). No connection is trusted for
+  real operations before that. `InsecureIgnoreHostKey` is forbidden.
```
com6056, line 212: The half of TOFU that bites later, plus the pin lifecycle so `ssh-test` and credential deletion can't quietly re-pin.

```suggestion
  real operations before that. `InsecureIgnoreHostKey` is forbidden.
- What gets pinned is the marshaled public key and its algorithm, not a
  display string; later connects constrain `HostKeyAlgorithms` to the
  pinned type, so a key-type change is a mismatch, never a negotiation
  accident. Fingerprints render as `SHA256:` for humans only.
- A host-key change after pinning fails closed: no automatic re-pin, and
  no fallback to TOFU if the stored pin is missing or unreadable. The
  mismatch surfaces both fingerprints and both key types behind a
  confirmation deliberately heavier than first contact, one that names
  interception as a possible cause and points at out-of-band
  verification. A legitimate re-key and an interception look identical to
  qui, so the user makes that call, never the code. On a background
  reconnect nobody is there to prompt, so a mismatch parks the instance
  in a needs-reconfirmation state and fails every fsop until a human
  clears it.
- The pin belongs to the host, not the credential: deleting SSH
  credentials keeps the pin, `ssh-test` against a pinned instance routes
  into the mismatch flow rather than first contact, and editing host or
  port drops the pin deliberately and requires fresh confirmation. Tests
  cover mismatch-fails-closed, no re-pin via `ssh-test`, and key-type
  change.
```

docs/remote-backend-design.md
```diff
@@ -0,0 +1,194 @@
...
+  arguments are strictly quoted, and options are terminated with `--`
+  before any path for `find`, `stat`, `xargs`, `rm`, and `cp`.
+- The remote-backend PR carries tests for traversal payloads, option-like
+  filenames (`-rf`), and shell metacharacters in paths.
+
+## Security
+
+- Dedicated SSH key per instance; never the user's personal key.
+- Private key stored AES-GCM encrypted with AAD binding (instance id +
+  field), same `sessionSecret` pattern as existing credential encryption.
```
com6056: Binding the pin the same way as the key, with host and port in the AAD so a redirect fails as tampering instead of a maybe-mismatch.

```suggestion
- Private key stored AES-GCM encrypted, keyed from `sessionSecret` like
  existing credential encryption; the AAD binding (instance id + field)
  is new here, today's credential stores pass no additional data. The
  pinned host key is stored under the same AEAD with host and port in its
  AAD as well: the pin fixes where the key gets used, and encrypting the
  key alone says nothing about that. A plaintext-column edit redirecting
  the instance then fails as a decryption error, which is unambiguous
  tampering, rather than as a host-key mismatch, which is not. The scope
  is deliberate: this defeats a DB-write attacker and cross-instance
  transplant, not a stolen data directory (`sessionSecret` sits beside
  `qui.db` by default) and not rollback to an older row set for the same
  instance, which needs state outside the DB. Tests cover
  tampered-ciphertext failing closed rather than falling back to TOFU.
```

docs/remote-backend-design.md
```diff
@@ -0,0 +1,194 @@
...
+`hardlink.FileID` is already shared on develop (crossseed, orphanscan,
+dirscan, automations), so the ripple is the same size now or later, and
+nothing serializes identity until the remote backend exists. Migration
+stays cheap: consumers only use `==`, map keys, and `IsZero()`, all of
+which survive a comparable fixed-size form unchanged (`Bytes()` previews
+it); the churn is per-platform constructors and test literals.
+
+Guard regardless of representation: FileID comparisons are only valid
+within one backend/host — identity bytes collide across machines, so
+cross-instance comparisons must check same-backend first.
```
com6056: Pins the encoding, and updates the serialization line: `dirscan` already persists identity as `dir_scan_files.file_id`, so this ships with a migration whenever it lands. Also draws the trust boundary for peer-supplied identity.

```suggestion
**Decision: `hardlink.FileID` becomes an opaque, tagged, comparable
fixed-size form, implemented in the remote-backend PR** (raised by com6056
on #1914). Opaque is the only viable shape: unix identity is 16 bytes,
Windows identity is up to 24, so no packing into the other platform's
struct is lossless in either direction, and the type is already shared
across develop's consumers, so the ripple is the same size now or later.
Tagged, because untagged bytes let a unix `dev=1,ino=2` collide with a
Windows identity in the same space: `struct { kind uint8; raw [24]byte }`,
kind 0 for no identity, 1 for unix `dev`+`ino` big endian, 2 for a
Windows volume serial plus the 16-byte file id, 3 for a remote-provided
blob. Constructors zero the unused tail of `raw` so `==` stays
well-defined, and `IsZero()` then means "no identity" rather than "a
backend wrote zeros" (load-bearing: the hardlink index trusts any FileID
returned without error, so zeroes-as-identity would collapse torrents
into one delete-safe group). `Bytes()` returns the tagged form: dirscan
keys its FileID index on `string(FileID.Bytes())` and persists it as
`dir_scan_files.file_id` under a partial index on
`(directory_id, file_id)` matched by rename detection, so the encoding
change ships with a migration (or a read path accepting the legacy
widths) whenever it lands. Consumer churn is otherwise unchanged: `==`,
map keys, and `IsZero()` all survive; the churn is per-platform
constructors and test literals.

Guard regardless of representation: FileID comparisons are only valid
within one backend/host, and `==` still compiles cross-host, so
comparisons go through a helper that takes the backend scope (or the
scope rides in the value). Remote-sourced identity is always kind 3, even
when the remote is unix and the bytes are a `dev`/`ino`: it is parsed
from command output on a host qui does not control, and the tag keeps
peer-asserted identity type-distinct from kernel-attested identity. Peer
identity is advisory. It may suppress work (dedup accounting,
already-seeding skips) but never expands a destructive action:
automations' delete-expansion groups torrents by a FileID signature and
deletes the group with files, so on a remote backend that expansion
degrades to the no-identity behavior. A hostile remote can never reach a
worse outcome than the SFTP-only tier already produces with no identity
at all.
```

com6056 (commented):
The suggested edits I mentioned. Three are wording for decisions we already talked about, with the sharp edges from an adversarial pass written in (the SFTP tier's residual delete race, the pin lifecycle, what the AAD binding actually buys and what it doesn't). S4 also folds in one thing I hit reading dirscan: `dir_scan_files.file_id` already persists `FileID.Bytes()`, so the encoding change ships with a migration whenever it lands. All of this is still design-time, fsops has no consumers outside itself yet, so these are wording changes, not refactors. Accept, rework, or push back on any of them.

One question rather than a suggestion: the old #1917 schema allowed `ssh_auth_type = 'password'`, and a password gets transmitted to whatever host the pin admits, so if the reshape keeps it the security section needs to cover that, and if it drops it, saying key-only is cheaper.

docs/remote-backend-design.md
nitrobass24: Applied as suggested in c933f52f. Good catch on the dirscan persistence — the "nothing serializes it" premise was mine and it was wrong; the migration note and the advisory-identity trust boundary both read right.

docs/remote-backend-design.md
nitrobass24: Applied in c933f52f, including the honest SFTP v3 race bound.

docs/remote-backend-design.md
nitrobass24: Applied in c933f52f.

docs/remote-backend-design.md
nitrobass24: Applied in c933f52f.

nitrobass24:
@Audionut Thanks for running this — all six landed somewhere concrete:

- **Path semantics (High)**: agreed this is the real contract gap. The design doc now has a Path Domains section (0f1b9054): every path belongs to its backend's filesystem in native form — local speaks host paths so current callsites are correct by construction, remote speaks slash-delimited POSIX regardless of host OS, and host `filepath` never touches a remote path. The remote-backend PR introduces a path dialect for backend-owned manipulation; local's dialect is host `filepath`, so the callsite migration bakes nothing in that the remote breaks. @s0up4200 this one's a contract call — flagging for your sign-off.
- **Cancellation/rollback (Medium)**: the rollback half is fixed on #1915 — all rollback paths (dirscan inject, seasonpack, hardlink/reflink) run under `context.WithoutCancel` now. Mid-operation interruption of tree creates is deliberate: `pkg/hardlinktree` is ctx-free, bounded, and rolls itself back on failure.
- **Portable errors (Medium)**: contract documented on the interface (`errors.Is` with `fs.ErrNotExist`/`fs.ErrPermission` across all methods) with the local backend as reference under test (0f1b9054).
- **SFTP hardlink conflicts (Medium)**: doc'd — without identity, an existing target can't be proven the same link, so SFTP-only treats it as a conflict and fails the create, never an idempotent skip.
- **Interface size (Low)**: prior decision on record — declined with the design-doc exception, and the surface was already cut 17→12 this round.
- **Lint**: real catch, develop-inherited — the windows-tagged sharedextents test never got the #2287 signature change and nothing cross-vets `pkg/` on Windows. Fixed in c4a763d2.

nitrobass24:
@com6056 Stat's consumers are wired now — you were right. `fileid_index` and `scanSingleFile` call `Stat` on #1915 (d4cf3198), and the symlink test from the migration turned out to be asserting the regression as intended behavior, so it now asserts develop's (89c0fab0). 

On the cost fork I took wiring over opt-in: the Windows double-open on missing-files is real but bounded, and identity-opt-in is exactly the kind of flag we just finished stripping. If it stings in practice it can ride in with the remote backend, which needs per-op identity control anyway. 

All four doc suggestions applied as-is in c933f52f  the dirscan persistence catch corrected a factually wrong line of mine, appreciated. Threads resolved.

nitrobass24:
@s0up4200 — the stack has been through com's re-review (all threads resolved), Audionut's scripts, CodeRabbit re-runs, and a Fable adversarial multi-agent pass. Everything found is fixed. What's left is the set of calls that are yours to make. All of them live in `docs/remote-backend-design.md` on this PR, so a doc review covers the lot.

**Merge safety:** this PR changes no behavior on develop. `internal/fsops` has zero callers until #1915/#1916 wire it in — the diff is the new package, docs, a .gitignore line, and a test-only fix for the windows-tagged sharedextents test that's currently broken on develop (pre-#2287 signature).

**Needs your decision:**

1. **Direction sign-off** — the doc is the artifact for the thing I originally wanted your blessing on before committing another migration: no deployed agent, single SSH connection with SFTP subsystem + opt-in exec channels, capabilities probed per-connect. #1913 is closed but preserved as the fallback tier if this hits a performance wall.
2. **Path Domains contract** (§Path Domains) — backend paths live in the backend's native dialect; host `filepath` never touches a remote path; a path dialect lands with the remote PR. This constrains how #1915's callsites treat paths, so it should be settled before that merges. Raised by Audionut, seconded by com; it's the one open contract disagreement the reviews surfaced.
3. **Security/storage model** (§Security, §Schema) — pinned host key stored as marshaled public key + algorithm (not a fingerprint) under the same AES-GCM/`sessionSecret` pattern the credential stores use today, with AAD binding (instance+field, plus host+port on the pin) that the legacy columns don't have. It's a deliberate ratchet: new columns stronger than old, old untouched. Motivating case is Postgres deployments where the DB lives away from `sessionSecret`. This also fixes the #1917 scope — the Schema section is what that PR will build.

**Decided during review — veto if you disagree:**
- `FileID` becomes an opaque *tagged* form (kind byte + bytes) at remote-backend time; remote-parsed identity is its own kind and stays advisory — never expands a destructive action. dirscan already persists `FileID.Bytes()`, so the encoding lands with a migration.
- `SameFilesystem` with unusable fsids returns an explicit error, never a guess — both migrated callers already treat that as "don't hardlink."
- `StatBatch`/`LstatBatch` return with the remote-backend PR as their consumer (your drop-until-needed standard, applied in both directions).
- SFTP extensions are probed per-operation; missing extension = explicit unsupported error, not an assumed capability.

**After this merges:** #1915 is where your original eight findings were fixed (each in its own commit).

PR #1915, opened 2026-05-19
refactor(services): use fsops.Backend for all filesystem operations

nitrobass24:
## Summary

Replaces all direct `os.*`/`unix.*` filesystem calls across the services with `fsops.Backend` interface calls. The local backend delegates to the same `os.*` calls as before, and **behavior parity with develop is the bar** — three review rounds (s0up4200, com6056, Audionut, CodeRabbit, plus an adversarial multi-agent pass) verified it, and where the migration had drifted, this branch restores parity. Every service that touches torrent data on disk now resolves the owning instance's backend, so a future remote backend slots in with no feature-code changes.

### Services updated
- **automations**: `missing_files.go`, `free_space.go`, `hardlink_index.go` — `os.Stat`/`unix.Statfs`/`os.Lstat` → backend calls; `scanTorrentFiles` takes ctx + backend and reads FileID/Nlinks from `backend.Lstat`
- **qbittorrent**: `delete_cleanup.go`, `sync_manager.go` — managed-delete cleanup through the backend; the pool rides an `atomic.Value` matching the neighboring `filesManager`
- **dirscan**: `scanner.go`, `fileid_index.go`, `inject.go`, `align.go` — full walker + link-tree refactor to backend
- **orphanscan**: `walker.go`, `delete.go` — full walker + delete refactor to backend
- **crossseed**: `service.go`, `seasonpack.go` — hardlinked-copy detection, hardlink/reflink tree creation + rollback, `SupportsReflink`, season-pack link pipeline, `FindMatchingBaseDir`, `matchedFilesystemProbePath`

### Parity restorations (review-caught migration drift)
- orphan-scan alias dedup gate restored to develop's nlink==1 semantics (#1212 protection), with develop's test expectations
- symlinked torrent data follows the target again: dirscan's already-seeding index and root-level single-file searchees use `Stat` (followed identity), matching develop's `os.Stat`
- per-file stat failures no longer abort whole scans — WalkDir emits enumeration errors only, so consumers skip unreadable files like develop did
- Windows keeps refusing path-based free space (develop behavior); the free-space files are deduplicated to one per-platform function
- torrent file names carrying literal backslashes are rejected instead of rewritten into invented paths (`buildFullPath` and crossseed's `resolveLocalTorrentFile`), so a legal Linux filename can't read as "missing" and trigger a destructive rule

### Hardening for failure modes the backend abstraction introduces
- rollbacks survive cancellation: link-tree rollback paths run under `context.WithoutCancel`, and the backend that *created* a tree is threaded to its rollback (dirscan `linkBackend`, `seasonPackPlanBuild.backend`) so a failed re-resolve can't silently skip cleanup
- strict-mode cross-seed checks fail closed on backend outage: resolve failures feed `verificationErr`, so a delete dialog shows "failed to check" instead of a false "no cross-seeds"
- WalkDir consumers cancel + drain on early return (no leaked walk goroutines); orphan-scan delete resolves the backend once per batch with a nil-pool guard

### Tree operations follow the handle-based rollback model
`TreeCreateResult` records the files/dirs a create call actually made, and `RemoveTree` removes exactly that — never the whole plan, which could delete links shared with sibling torrents (discussion #2282). The `seasonPackLinkCreator` test hook is now `func(ctx, plan) (*fsops.TreeCreateResult, error)`.

### Deliberately not migrated
`sharedextents` (FIEMAP ioctl comparing physical extents) is host-bound by nature and stays a local-backend capability; the reflink-copy match upgrade treats `ErrUnsupported` as no-evidence and degrades cleanly.

### Wiring
`cmd/qui/main.go` creates one `fsops.Pool` (local backend) and passes it to all service constructors; `SyncManager`/`CrossSeedService` receive it via setters before serving.

### Test-only behavior notes
Tests that build services inline wire an `fsops.Pool`; dirscan alignment tests set `HasLocalFilesystemAccess` because stats now correctly refuse instances without filesystem access.

Depends on #1914 (fsops interface + local backend).
Supersedes #1916.

## Test plan

- [x] `go build ./...` and `GOOS=windows go vet ./internal/... ./pkg/...` — clean
- [x] `go test -race -count=1` — automations, dirscan, orphanscan, crossseed, qbittorrent, fsops — all pass
- [x] `make precommit` + `make build` — pass
- [x] CI: backend, web, postgres, migration-parity, Docker builds — green

### Live test

Run against a real instance on the `pr-1915` image (16,603 torrents, 33,114 files on disk, local filesystem access; plus a second instance with **no** filesystem access, which exercises the noop backend). Verified on two builds, the later one including this branch's develop merge; the automations/hardlink/`Statfs` probes reproduce identically on both.

**No error, warning, or panic attributable to this PR** appeared in the live log stream or in 192h of persisted history — 794 automation activities, 7 orphan-scan runs, 10 cross-seed runs, 5 dirscan runs, zero failures. Two mid-run cancellations exited cleanly with no leaked walks and no rollback warnings.

**34 of 48 migrated call sites were observed executing against real data.** Coverage by subsystem:

| Subsystem | Covered | Key evidence |
|---|---|---|
| automations | 3/4 | hardlink index over 16,603 torrents (`invalidPaths 0`); missing-files `checked=16581/16581`; `Statfs` returns per-path values and fails closed on a bad path |
| orphanscan | 8/10 | walk of 33,114 files across 47 roots; purpose-built fixture deleted via file, directory and disc-unit branches |
| dirscan | 10/14 | 5 scans incl. one that walked 1,378 files; two successful reflink injections rechecking to `progress: 1` |
| crossseed | 7/16 | 56 automated injections landing in the `by-tracker` link-tree layout while this build was live |
| qbittorrent | 2/2 | managed delete pruned the release dir and halted at the base dir |
| free space | 1/2 | path-source `Statfs` verified; Windows refusal needs a Windows host |

One results worth a reviewer's attention:

- **`RemoveTree` honours the `TreeCreateResult` model (discussion #2282).** A hardlink create failed partway; afterwards the release directory it made is gone while the pre-existing parent still holds 1,013 files. It removed exactly what it created and left the shared parent alone.

**Not covered live**, with reasons: 
 - season-pack (5 — needs a real pack whose episodes are already seeding individually); 
  - the four rollback paths (need a create that fails *after* the tree exists, i.e. fault injection); 
  - `processHardlinkMode` (needs hardlink mode plus an injection on a single mount); 
  - `shouldUseSearcheeDirectory` (needs regular mode or fallback enabled); Windows free-space refusal; and 
  - `discParentIsPureDiscRoot`, whose `tfm == nil` guard is unreachable in production on develop too. All are covered by the unit tests above.

<details>
<summary>Full call-site checklist (48)</summary>

Every migrated call site below. Checked = observed executing against real data on this build.

**`internal/services/automations/hardlink_index.go`**
- [x] `scanTorrentFiles` → `Lstat` (:256) — `automations: hardlink index built`, 16,603 torrents, scopeComputed 16,580, outsideLinks 1,195, **invalidPaths 0**, inaccessible 23, 15.3 s. `HARDLINK_SCOPE` preview returns 905 outside / 71 inside, so scope is being derived, not defaulted.
- [x] `scanOtherInstancesForDeficits` → `Lstat` (:1156) — **noop-backend branch only.** Ran against the no-filesystem-access instance; every call returned `ErrNoFilesystemAccess` and the index completed unpoisoned (`HARDLINK_SCOPE_CROSS` matched the same 905/71). The local-backend branch is untested — see gaps.

**`internal/services/automations/missing_files.go`**
- [x] `detectMissingFiles` → `Stat` (:61) — `missing files detection completed`, `checked=16581` of 16,581 completed torrents, i.e. every path resolved. `HAS_MISSING_FILES` preview: 18 true / 7,948 false; all 18 trace to priority-0 files never written to disk, which is develop's behavior too.

**`internal/services/crossseed/service.go`**
- [x] `forEachLocalTorrentFile` → `Lstat` (:1346) — three `local-matches?strict=true` calls returned 200 with `content_path` matches. Strict mode fails closed on any `sourceFilesErr`/`candidateFilesErr`/`verificationErr`, so a 200 here is proof the backed `Lstat` sweep completed.
- [ ] `processHardlinkMode` → `HardlinkTree` (:14760)
- [ ] `processHardlinkMode` → `RemoveTree` rollback under `context.WithoutCancel` (:14867)
- [x] `FindMatchingBaseDir` → `MkdirAll` (:15091) — see the 56-injection note below.
- [x] `FindMatchingBaseDir` → `SameFilesystem` (:15096) — same.
- [x] `matchedFilesystemProbePath` → `Stat` (:15120) — runs ahead of `FindMatchingBaseDir` to supply the source path (:14681, :15405), so it is on every one of those injections.
- [x] `materializeReflink` → `SupportsReflink` (:15219) — same.
- [x] `materializeReflink` → `ReflinkTree` (:15226) — **56 cross-seed injections** completed while this build was live (`webhook.qui` / `completion.qui` tagged, 2026-08-21 → 2026-08-22), every one landing in the `by-tracker` link-tree layout under a configured base dir rather than at a source-derived save path. With `useReflinks: true` and `fallbackToRegularMode: false`, a failure anywhere in this chain aborts the injection instead of copying — so 56 successes is proof the whole reflink path works.
- [ ] `processReflinkMode` → `RemoveTree` rollback (:15600)

**`internal/services/crossseed/seasonpack.go`**
- [ ] `seasonPackBaseDirMatchesAllSources` → `MkdirAll` (:647)
- [ ] `seasonPackBaseDirMatchesAllSources` → `SameFilesystem` (:652, :656)
- [ ] `nearestExistingParent` → `Stat` (:671)
- [ ] `rollbackSeasonPackTree` → `RemoveTree` (:1564)
- [ ] `rollbackSeasonPackTree` → `Remove` of the pack dir (:1569)

**`internal/services/orphanscan/walker.go`**
- [x] `walkScanRootWithUnitFilter` → `WalkDir` (:217) — manual run 223: `orphanscan: built file map`, 33,114 files across 47 roots in 64 s, no enumeration errors, 0 orphans. Consistent with the five prior scheduled runs.
- [x] `discParentIsSafeDiscRoot` → `ReadDir` (:521) — the `VIDEO_TS` fixture was reported as its **parent directory**, not as the file inside it, which only happens when disc-unit detection resolves the parent as a safe disc root.
- [ ] `discParentIsPureDiscRoot` → `ReadDir` (:493) — the `tfm == nil` fallback inside `discParentIsSafeDiscRoot`. `NewTorrentFileMap()` (service.go:1306) is the map's only construction site, always returns non-nil, and is threaded straight to the walker, so this branch is unreachable in production and covered only by unit tests. Pre-existing: develop has the same guard.

**`internal/services/orphanscan/delete.go`** — exercised with a purpose-built fixture under a live scan root: a plain 2 MB file, a symlink to it, and a `VIDEO_TS`-only folder (a "pure disc root", so the walker reports the parent as one unit). Manual scan → preview listed exactly the two expected units and nothing else → confirmed. Result: `filesDeleted 2`, `bytesReclaimed 4 MB`, no errors.
- [x] `safeDeleteTarget` → `Lstat` (:141) — the dispatcher for both units; routed the file and the directory down different branches.
- [x] `safeDeleteFile` → `Lstat` (:69), `Remove` (:80) — the plain file.
- [x] `safeDeleteDirectory` → `Remove{Recursive:true}` (:179) — the disc unit, removed as a whole.
- [x] `checkDirContainsInUseFile` → `WalkDir` + cancel/drain on early return (:92) — runs before every directory delete; passed with no in-use file found.
- [x] `safeDeleteEmptyDir` → `Remove` (:195) — attempted on the parent of the deleted file and correctly **refused**, because the symlink was still there. `foldersDeleted 0` is the "never recursive" guarantee holding.
- [ ] `safeDeleteSymlink` → `Remove` (:162) — **not reachable through a normal scan.** The walker skips symlinks (`entry.IsSymlink → continue`), exactly as develop does (`d.Type()&fs.ModeSymlink != 0 → return nil`), so a symlink never becomes an orphan record. This branch only fires when a path recorded as a file at scan time is a symlink by the time deletion runs. Parity is preserved; the branch is defensive.

**`internal/services/dirscan/`** — dirscan is disabled on this instance with zero configured directories, so the whole subsystem is unexercised.
- [x] `scanner.ScanDirectory` → `ReadDir` (scanner.go:92) — manual scan run 63 over a scan root, enumerated 1 searchee.
- [x] `scanner.scanSearcheeDir` → `WalkDir` (scanner.go:171) — same run, walked the searchee and discovered 3 files with FileID + link count populated.
- [x] `scanner.scanSingleFile` → `Stat` (scanner.go:231) — scanned a root holding a single bare media file. The resulting searchee is named with the extension stripped (`<name>.mkv` → `<name>`) and carries `size` straight from the `Stat` result, which is the signature of this branch; a directory searchee would take its name from the folder via `scanSearcheeDir`.
- [x] `scanner.isDiscLayoutRoot` → `ReadDir` (scanner.go:287) — called unconditionally from `scanSearcheeDir` (:167) before the walk; ran and correctly returned false for a non-disc release folder.
- [x] `align.searcheePathIsDir` → `Stat` (align.go:139) — called unconditionally at inject.go:245 to build the alignment plan, ahead of the add, so both successful injections ran it. (The alignment itself is scoped to regular-mode full matches and correctly did not apply to these link-tree adds.)
- [x] `fileid_index.addTorrentFilesToFileIDIndex` → `Stat` (fileid_index.go:89) — `dirscan: built FileID index`, 16,581 torrents / **33,037 FileIDs** resolved in 17.2 s, statErrors 37 (0.11%, consistent with the 18 known missing-file torrents). This is the `Stat`-not-`Lstat` parity restoration, and it worked end to end: all 3 scanned files came back `fileIDPresent: true` → `alreadySeeding: true`, i.e. the index matched them to live torrents.
- [ ] `inject.shouldUseSearcheeDirectory` → `Stat` (inject.go:672) — reachable in ordinary use, via `calculateSavePath` at inject.go:586 (regular mode) and :602 (fallback after a link failure). Untested only because the test instance runs reflinks with `fallbackToRegularMode: false`, so injection returns at :598 first. Needs an instance with both link modes off, or fallback enabled.
- [x] `inject.materializeLinkTree` → `MkdirAll` (inject.go:779) — dirscan run 66 (13 searchees / 1,378 files) matched one release and entered injection.
- [x] `inject.createLinkTree` → `SupportsReflink` (inject.go:922) — returned supported; the call must succeed for the reflink branch to continue.
- [x] `inject.createLinkTree` → `ReflinkTree` (inject.go:929) — **two successful dirscan injections**, each landing in the `by-tracker` layout under a configured base dir and rechecking to `progress: 1`, i.e. qBittorrent verified every file at the cloned paths. Notably this is the same source/destination pair that failed for `HardlinkTree` with `EXDEV`: `FICLONE` works on the shared superblock while `link()` refuses across two `vfsmount`s, so reflink mode crosses a bind-mount boundary that hardlink mode cannot.
- [x] `inject.createLinkTree` → `SameFilesystem` (inject.go:937) — returned true for a source and base dir that were two bind mounts of the same dataset (`st_dev` matches; the kernel still refuses the link because `do_linkat` compares `vfsmount`). The create then failed at `link()` with `EXDEV`, which is the designed path: this call is a pre-flight hint, and `HardlinkTree`'s `EXDEV` handling is the actual guard. `pkg/fsutil` is untouched by this PR.
- [x] `inject.createLinkTree` → `HardlinkTree` (inject.go:949) — exercised with the instance temporarily flipped to hardlink mode. Returned `EXDEV`, which the caller caught and reported as an actionable message naming both fixes; with `fallbackToRegularMode: false` it refused rather than silently copying 12 GB.
- [ ] `inject.rollbackLinkTree` → `RemoveTree` via the recorded `linkBackend` (inject.go:637) — not reached: this fires when an injection fails *after* the tree exists, whereas the observed failure was inside the create.
- [ ] `inject.rollbackLinkTree` → `Remove` of the root dir (inject.go:640) — same.

`HardlinkTree`'s own partial-failure rollback **was** exercised and behaved per the `TreeCreateResult` model (discussion #2282). It created `<baseDir>/<release>/`, failed on the first `link()`, and rolled back: scanning that release dir afterwards returns `no such file or directory`, while the pre-existing plan root `<baseDir>` still scans fine with 1,013 files. It removed exactly what it made and left the shared parent alone — which is the property that stops a rollback from deleting links belonging to sibling torrents.

**`internal/qbittorrent/delete_cleanup.go`** — verified with a deliberate managed delete: a disposable single-folder torrent (5 files, 6 MB, zero cross-seed siblings) sitting at `<baseDir>/<release folder>/`, with its parent temporarily added to the instance's `HardlinkBaseDir`.
- [x] `managedDeleteCleanupDir` → `Stat` (:70) — resolved the content path to a directory and selected it as the prune target.
- [x] `pruneEmptyManagedDeleteDirOnce` → `Remove` (:154) — after the delete, scanning the release folder path fails with `no such file or directory`, so the directory was removed; scanning the base dir still succeeds, so the upward walk **halted at the base dir** instead of continuing to prune. Exactly one directory removed, which is the intended loop bound. Nothing logs on the success path, hence the filesystem-level check.

**`internal/services/automations/free_space_unix.go`**
- [x] `GetFreeSpaceBytesForSource` → `Statfs` (:44) — set a rule's source to `path` and compared match counts at a fixed threshold: `/data` gives 985 (identical to the qBittorrent source, correct — the save paths live there), while `/config` and `/` give 2,324, so the backend is genuinely supplying the number. A path that does not exist fails closed with HTTP 500 and logs `automations: failed to get free space: ... statfs /nonexistent-probe-dir: no such file or directory` rather than treating the disk as empty.

**`internal/services/automations/free_space_windows.go`** — Windows refuses path-based free space by design; needs a Windows host.
- [ ] path-source refusal on Windows

</details>

## Summary by CodeRabbit

* **New Features**
  * Added filesystem-aware operations across scanning, synchronization, cross-seeding, automation, and cleanup workflows.
  * Free-space checks can evaluate configured filesystem paths on supported platforms.
  * Root-level media symlinks are recognized during directory scans.
  * Improved cross-seed searches with cached media IDs, alternate-title queries, and clearer candidate decision tracking.

* **Bug Fixes**
  * Improved handling of unavailable filesystems, missing paths, and uncertain metadata.
  * Strengthened path validation against traversal and unsafe paths.
  * Cleanup operations now continue safely after request cancellation and roll back partial changes.
  * Orphan cleanup now avoids deleting files when their status cannot be verified.


nitrobass24:
Splitting into smaller feature-scoped PRs per updated implementation plan.

internal/fsops/backend.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// --- Diagnostic ---
+
+	// Info returns metadata about this backend (kind, version, capabilities).
+	Info(ctx context.Context) (*BackendInfo, error)
+
+	// HealthCheck returns nil if the backend is operational, or an error
+	// describing the problem.
+	HealthCheck(ctx context.Context) error
+}
```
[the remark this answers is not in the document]
nitrobass24: Deliberate exception, same as the resolved thread on #1914: `Backend` is one abstraction boundary — the surface a remote implementation must provide — and splitting it would thread composed-interface plumbing through the pool and every consumer with no partial implementation to justify it.

internal/fsops/local/local_test.go
```diff
@@ -0,0 +1,501 @@
...
+	dir := t.TempDir()
+
+	fi, err := b.Stat(context.Background(), dir)
+	require.NoError(t, err)
+	assert.True(t, fi.IsDir)
+}
+
+func TestStat_NotFound(t *testing.T) {
+	b := newBackend()
+	_, err := b.Stat(context.Background(), "/nonexistent/path/xyz")
```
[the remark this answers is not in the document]
nitrobass24: Fixed in cfecf93e — built under `t.TempDir()` now.

internal/fsops/local/local.go
```diff
@@ -0,0 +1,328 @@
...
+					entry.IsDir = fi.IsDir()
+					entry.IsSymlink = fi.Mode()&os.ModeSymlink != 0
+
+					if (opts.WantFileID || opts.WantNlinks) && fi.Mode().IsRegular() {
+						fid, nlinks, fidErr := hardlink.GetFileID(fi, path)
+						if fidErr == nil {
+							entry.FileID = fid
+							entry.Nlinks = nlinks
+						}
+					}
```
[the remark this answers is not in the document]
nitrobass24: Fixed in cfecf93e — when the caller asks for identity data, a failed lookup now sets `entry.Err` instead of yielding a zero FileID.

internal/fsops/local/local.go
```diff
@@ -0,0 +1,328 @@
...
+
+func (b *Backend) Remove(ctx context.Context, path string, opts fsops.RemoveOptions) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	if opts.Recursive {
+		return os.RemoveAll(path)
+	}
+	return os.Remove(path)
+}
```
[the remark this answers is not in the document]
nitrobass24: Already handled: `Remove` fails explicitly with "recursive remove with IgnorePaths is not supported by the local backend" rather than silently dropping the option.

internal/fsops/local/local.go
```diff
@@ -0,0 +1,328 @@
...
+		return &fsops.TreeCreateResult{RolledBack: true}, err
+	}
+	return &fsops.TreeCreateResult{Created: len(plan.Files)}, nil
+}
+
+func (b *Backend) RemoveTree(ctx context.Context, plan *hardlinktree.TreePlan) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+	return hardlinktree.Rollback(plan)
```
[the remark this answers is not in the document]
nitrobass24: Skipping: the underlying create is short-lived and rolls back internally on failure; aborting a link tree mid-flight would leave exactly the partial state the rollback design exists to prevent. Marked outdated since the handle-based rework.

internal/fsops/pool.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+}
+
+// GetBackend returns the appropriate Backend for the given instance ID.
+// Returns ErrNoFilesystemAccess wrapped in a noop backend for instances
+// without filesystem access configured.
+func (p *Pool) GetBackend(ctx context.Context, instanceID int) (Backend, error) {
+	instance, err := p.instanceStore.Get(ctx, instanceID)
+	if err != nil {
+		return nil, fmt.Errorf("load instance %d: %w", instanceID, err)
```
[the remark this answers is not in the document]
nitrobass24: Addressed on #1914: `NewPool` panics at construction on nil dependencies, so `GetBackend` cannot see them.

internal/services/automations/free_space_windows.go
```diff
@@ -83,8 +84,15 @@ func GetFreeSpaceBytesForSource(
...
-		return 0, errors.New("path-based free space source is not supported on Windows")
+		// Read free space via the backend (local or remote helper)
+		if backend == nil {
+			return 0, errors.New("backend is required for path-based free space source")
+		}
+		result, err := backend.Statfs(ctx, resolved.Path)
+		if err != nil {
+			return 0, fmt.Errorf("failed to get filesystem stats for %s: %w", resolved.Path, err)
+		}
+		return result.BytesAvailable, nil
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — the Windows implementation now trims and rejects blank paths with an explicit error, matching the Unix one. A configured path source with a blank path is a config error surfaced to the user rather than silently reading a different disk.

internal/services/automations/service.go
```diff
@@ -934,7 +937,13 @@ func (s *Service) setupFreeSpaceContext(ctx context.Context, instanceID int, rul
 		return nil
 	}
 
-	freeSpace, err := GetFreeSpaceBytesForSource(ctx, s.syncManager, instance, rule.FreeSpaceSource)
+	backend, err := s.backendPool.GetBackend(ctx, instanceID)
+	if err != nil {
+		log.Warn().Err(err).Int("instanceID", instanceID).Msg("automations: no backend for free space check")
+		return nil
+	}
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — backend resolution failure now fails FREE_SPACE setup instead of evaluating without context. The nil-pool guard is omitted: main wires the pool unconditionally.

internal/services/dirscan/service.go
```diff
@@ -741,7 +745,13 @@ func (s *Service) validateDirectory(ctx context.Context, directoryID int, runID
 // runScanPhase executes the directory scanning phase.
 // Returns the raw scan result and true if successful, or nil and false on failure.
 func (s *Service) runScanPhase(ctx context.Context, dir *models.DirScanDirectory, scanRoot string, runID int64, l *zerolog.Logger) (*ScanResult, map[string]string, bool) {
-	scanner := NewScanner()
+	backend, err := s.backendPool.GetBackend(ctx, dir.TargetInstanceID)
+	if err != nil {
+		l.Warn().Err(err).Msg("dirscan: no filesystem backend, skipping scan")
+		return nil, nil, false
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — both early returns now call `markRunFailed`, so the run reaches a terminal state.

internal/services/orphanscan/service.go
```diff
@@ -555,7 +558,13 @@ func (s *Service) executeScan(ctx context.Context, instanceID int, runID int64)
 			return
 		}
 
-		orphans, _, err := walkScanRoot(ctx, root, tfm, ignorePaths, gracePeriod, 0)
+		backend, backendErr := s.backendPool.GetBackend(ctx, instanceID)
+		if backendErr != nil {
+			log.Error().Err(backendErr).Int("instanceID", instanceID).Msg("orphanscan: failed to get backend")
+			s.failRun(ctx, runID, instanceID, fmt.Sprintf("failed to get backend: %v", backendErr))
+			return
```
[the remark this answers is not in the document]
nitrobass24: The scan entry point already nil-checks the pool and fails the run; deletion only runs against a completed scan, and production wiring always provides the pool, so the extra guards would defend an unreachable state.

internal/qbittorrent/sync_manager.go
```diff
@@ -1857,7 +1871,13 @@ func (sm *SyncManager) BulkAction(ctx context.Context, instanceID int, hashes []
...
 		// Invalidate caches for deleted torrents
 		if err == nil {
-			cleanupManagedDeleteTargets(managedDeleteCleanupTargets)
+			if sm.backendPool != nil {
+				if backend, backendErr := sm.backendPool.GetBackend(ctx, instanceID); backendErr == nil {
+					cleanupManagedDeleteTargets(ctx, managedDeleteCleanupTargets, backend)
+				} else {
+					log.Warn().Err(backendErr).Int("instanceID", instanceID).Msg("managed delete cleanup: failed to get backend")
+				}
+			}
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — the builder returns the backend it resolved and cleanup reuses it; a failed resolve now logs and skips explicitly.

documentation/design/ssh-helper-plan.md
```diff
@@ -0,0 +1,936 @@
...
+- Never log credentials or full key material. Mask SSH keys in logs.
+- Use `golang.org/x/crypto/ssh` directly — do not wrap the `ssh` CLI.
+- Host key verification must not default to insecure. TOFU is fine; `InsecureIgnoreHostKey` is not.
+- Goroutines spawned for connection handling need explicit lifecycle management with `sync.WaitGroup` or `errgroup`.
+
+**General coding standards:**
+- `gofmt`-clean, PascalCase exports, camelCase locals
+- Interfaces max 5 methods (the 17-method `Backend` is the documented exception)
+- Explicit error handling, no silent failures
+- `go test -race -count=3` for all test runs
```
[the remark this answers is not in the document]
nitrobass24: This file was committed by accident (local scratch swept up in a merge); removed from the PR in fc76b33e.

internal/services/automations/hardlink_index.go
```diff
@@ -608,7 +633,7 @@ func (s *Service) scanHashes(ctx context.Context, instanceID int, torrentByHash
 			scanned[hash] = &torrentFileInfo{savePath: torrentByHash[hash].SavePath}
 			continue
 		}
-		scanned[hash] = scanTorrentFiles(torrentByHash[hash], files)
+		scanned[hash] = scanTorrentFiles(ctx, backend, torrentByHash[hash], files)
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — cancellation now returns `nil, false`, falling back to the full build, which already refuses to cache on cancel.

internal/services/automations/hardlink_index.go
```diff
@@ -818,7 +854,7 @@ func isPathInsideBase(basePath, fullPath string) bool {
 	// Check if the relative path escapes the base:
 	// - ".." means direct parent traversal
 	// - Paths starting with "../" traverse upward
-	if rel == ".." || strings.HasPrefix(rel, ".."+string(os.PathSeparator)) {
+	if rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)) {
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — `buildFullPath` normalizes to slash form and rejects absolute, drive-qualified, UNC, and dot-dot names in either separator style before conversion; all three callers treat rejection as invalid. Test covers both forms.

internal/services/dirscan/scanner.go
```diff
@@ -85,39 +89,39 @@ func (s *Scanner) ScanDirectory(ctx context.Context, rootPath string) (*ScanResu
...
-	if entry.IsDir() {
-		s.processDirEntry(ctx, entry, entryPath, result)
-	} else if isMediaFile(entry.Name()) {
-		s.processFileEntry(entryPath, result)
+func (s *Scanner) processRootEntry(ctx context.Context, entry fsops.DirEntry, entryPath string, result *ScanResult) {
+	if entry.IsDir {
+		s.processDirEntry(ctx, entryPath, entry.Name, result)
+	} else if isMediaFile(entry.Name) {
+		s.processFileEntry(ctx, entryPath, result)
 	}
```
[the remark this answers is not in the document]
nitrobass24: Fixed in e1bb3073 — root-level symlinks are skipped like the directory walk, with a test.

internal/api/handlers/crossseed_apply_handler_test.go
```diff
@@ -231,7 +232,9 @@ func TestSeasonPackApply_Returns500ForFailedApplyResponse(t *testing.T) {
 			SeasonPackCoverageThreshold: 1,
 		}, nil
 	})
-	setServiceField(t, svc, "seasonPackLinkCreator", func(*hardlinktree.TreePlan) (*hardlinktree.Created, error) { return &hardlinktree.Created{}, nil })
+	setServiceField(t, svc, "seasonPackLinkCreator", func(context.Context, *hardlinktree.TreePlan) (*fsops.TreeCreateResult, error) {
+		return &fsops.TreeCreateResult{}, nil
+	})
```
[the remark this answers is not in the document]
nitrobass24: Confirmed — and it went deeper than the missing pool: the test's fake `/media/...` source path failed base-dir selection, so the 500 had been coming from assembly all along, even before this PR. Fixed in 5d6f6e5a: the service gets a local backend pool, the episode exists on disk with the pack file's exact size on the same filesystem as the hardlink base dir, the pack is single-file (an unmatched extra would trip the piece-boundary safety check before the add), and the test now asserts `AddTorrent` was called exactly once so the 500 provably comes from the add failure.

internal/services/crossseed/service.go
```diff
@@ -1059,14 +1091,27 @@ func (s *Service) localLinkedMatchType(
 		return ""
 	}
 
-	if candidateSharesSourceFileID(sourceIDs, candidate.SavePath, candidateFiles) {
+	candidateBackend, err := s.getBackendForInstance(matchCtx.ctx, candidateInstance.ID)
+	if err != nil {
```
com6056, line 1135: This is the one I'd want fixed before merge. Returning `""` on a backend failure feeds none of the fields the fail-closed check at 828 reads (`verificationErr`, `sourceFilesErr`, `candidateFilesErr`), so in strict mode a backend outage reads as "no cross-seeds found" in the delete dialog instead of failing the check. A stat failure only loses one file's evidence, this branch discards all evidence for the candidate. `getSourceFileIDs` at 1031 and the source-backend branch at 1108 return the same way.

internal/services/dirscan/scanner.go
```diff
@@ -155,118 +159,78 @@ func (s *Scanner) processFileEntry(entryPath string, result *ScanResult) {
...
-func (s *Scanner) walkDirEntry(ctx context.Context, path string, d fs.DirEntry, walkErr error, searchee *Searchee) error {
-	if walkErr != nil {
-		if os.IsPermission(walkErr) {
-			return nil
+	for entry := range ch {
+		if entry.Err != nil {
+			if errors.Is(entry.Err, fs.ErrPermission) {
+				continue
+			}
+			return nil, fmt.Errorf("walk entry %s: %w", entry.Path, entry.Err)
```
com6056, line 193: Develop skipped a file it couldn't stat (`addFileToSearchee` returned nil on an `Info` error and `getFileIDSafe` swallowed identity failures), but those now arrive as `entry.Err` and abort the whole directory searchee here. The early return also leaves the channel undrained, so the producer blocks until the scan ctx dies. I'd demote per-file errors back to a skip, and pair the drain with a ctx cancel so it doesn't walk the rest of the tree just to throw it away.

internal/services/orphanscan/delete.go
```diff
@@ -87,24 +89,28 @@ func validateDeleteTarget(scanRoot, target string) error {
...
+			if errors.Is(entry.Err, fs.ErrNotExist) {
+				continue
+			}
+			return fmt.Errorf("walk entry: %w", entry.Err)
+		}
+		if entry.IsDir {
+			continue
+		}
+		if err := checkFileInUse(entry.Path, tfm); err != nil {
+			return err
```
com6056, line 129: Same leak as `scanner.go:184`: returning from inside the range leaves the walk goroutine blocked until the run's ctx dies, once per directory that still has an in-use file. Same fix, drain plus cancel.

internal/services/automations/service.go
```diff
@@ -5195,18 +5222,25 @@ func buildTrackerDisplayNameMap(customizations []*models.TrackerCustomization) m
...
-
-	cleaned := filepath.Clean(normalizedFile)
-	if filepath.IsAbs(cleaned) {
-		return cleaned
+// buildFullPath converts a torrent-internal (slash-delimited) file name into a
+// local filesystem path under basePath. Torrent metadata is untrusted on every
+// OS, so names that are absolute, drive-qualified, UNC, or escape basePath via
+// ".." are rejected in both POSIX and Windows form (ok=false).
+func buildFullPath(basePath, fileName string) (string, bool) {
+	slashName := strings.ReplaceAll(fileName, `\`, "/")
```
com6056: Not new to this branch (it mirrors develop's `resolveLocalTorrentFile`), but worth deciding on: backslash is a legal filename byte on Linux and qBittorrent sends slash-delimited paths, so a legal name like `AC\DC - Back In Black.mkv` comes back `ok=true` with a rewritten nested path, and missing-files then flags the torrent as missing files that exist. Rejecting the name like the `ok=false` path does for drive prefixes would satisfy the AGENTS.md escape rules without inventing paths, silently rewriting is the part that creates false positives. Probably wants the same fix as the crossseed copy in a follow-up.

internal/services/dirscan/inject.go
```diff
@@ -597,23 +618,23 @@ func (i *Injector) logLinkTreeFallback(instance *models.Instance, err error) {
...
 
-// rollbackLinkTree removes what the link-tree creator recorded in the handle,
+// rollbackLinkTree removes what the link-tree creator recorded in the result,
 // never the whole plan: target paths can be shared with an earlier successful
 // injection for the same release (discussion #2282). The root dir is removed
 // only when empty.
-func (i *Injector) rollbackLinkTree(created *hardlinktree.Created, rootDir string) {
-	if created == nil || rootDir == "" {
+func (i *Injector) rollbackLinkTree(ctx context.Context, created *fsops.TreeCreateResult, rootDir string, backend fsops.Backend) {
+	if created == nil || rootDir == "" || backend == nil {
```
com6056, line 631: The nil case is reachable: `materializeLinkTree` resolves its own backend at 763, so a transient store error on the resolve at 232 leaves `linkCreated` non-nil with a nil backend here and the tree survives rollback. Threading the backend `materializeLinkTree` already resolved out through `prepareInjection` closes it, same shape as the `sync_manager` cleanup-targets fix.

internal/services/dirscan/fileid_index.go
```diff
@@ -77,20 +81,19 @@ func collectCompletedTorrentSavePaths(torrents []qbittorrent.CrossInstanceTorren
 	return hashes, savePaths
 }
 
-func addTorrentFilesToFileIDIndex(index map[string]string, hash, savePath string, files qbt.TorrentFiles) (statErrors int) {
+func addTorrentFilesToFileIDIndex(ctx context.Context, index map[string]string, hash, savePath string, files qbt.TorrentFiles, backend fsops.Backend) (statErrors int) {
 	for _, file := range files {
 		absPath := filepath.Join(savePath, filepath.FromSlash(file.Name))
-		fi, err := os.Stat(absPath)
+		info, err := backend.Lstat(ctx, absPath)
```
com6056: `os.Stat` became `backend.Lstat`, so a symlinked torrent file now indexes the link, `FileID` stays zero (`osFileInfoToLstat` only fills regular files), and symlink-farm setups lose already-seeding detection. The interface can't express Stat-with-FileID today (`Backend.FileID` Lstats too), so counting these as skipped would at least make the degradation visible until the interface grows a follow-symlinks option.

internal/qbittorrent/sync_manager.go
```diff
@@ -476,6 +485,11 @@ func (sm *SyncManager) SetFilesManager(fm FilesManager) {
 	sm.filesManager.Store(fm)
 }
 
+// SetBackendPool sets the filesystem backend pool for managed delete cleanup.
+func (sm *SyncManager) SetBackendPool(pool backendPoolGetter) {
+	sm.backendPool = pool
```
com6056: No race today since this runs in `main` before the server starts, just consistency: `filesManager` beside it uses `atomic.Value` and crossseed holds this same pool in an `atomic.Value`. Constructor injection would be simpler still if the ordering allows it.

com6056 (commented):
The merge kept develop's summer features intact everywhere I checked, including the hardlink index and the season-pack rollback.

Two things worth stating in the PR description: root-level symlinked media files are no longer searchees (deliberate and tested, but user-visible for symlink-farm scan roots), and path-based free space is in a halfway state on Windows, `free_space_windows.go` now goes through `backend.Statfs` but `handlers/automations.go` still returns 400 on Windows, `capabilities.go` still advertises it as unsupported, and the workflow dialog still disables it. Either finish the enablement or keep the refusal consistent.

No converted callsite has a test through a real `Pool` with filesystem access off, one per service would pin the noop degradation behavior. And `isIgnoredOrphanDirName` in `orphanscan/walker.go` is dead now that the walker delegates to `WalkOptions.IgnoreDirNames`, the unused lint will flag it.

One small hardening: in `orphanscan/delete.go`, a ctx cancel mid-walk closes the channel with no error entry, so `checkDirContainsInUseFile` can return nil after an incomplete check, and `backend.Remove`'s own ctx pre-check is all that prevents a delete on an unverified directory. An explicit `ctx.Err()` check after the loop makes that deliberate.

internal/services/orphanscan/service.go
```diff
@@ -891,7 +904,13 @@ func (s *Service) executeDeletion(ctx context.Context, instanceID int, runID int
 			continue
 		}
 
-		disp, err := safeDeleteTarget(scanRoot, f.FilePath, tfm, ignorePaths)
+		deleteBackend, backendErr := s.backendPool.GetBackend(ctx, instanceID)
```
com6056: The scan path resolves the backend per scan root and guards `s.backendPool == nil` first. This loop calls `GetBackend` per deleted file with no guard (960 in the empty-dir cleanup loop does the same), so a nil pool panics here instead of failing the run. Hoist to once per batch and add the guard.

s0up4200:
this one isnt mergeable yet, the zero behavioral change claim doesnt hold in a few places that matter.

the big two: a per-file stat/FileID error develop skipped now discards the whole searchee in dirscan (scanner.go:180) and the whole scan roots results in orphanscan (walker.go:238). orphan scan feeds deletes, so silently losing a root is not ok.

rollback is broken under cancellation: inject.go:625 and the crossseed/seasonpack rollback paths pass the live run ctx into RemoveTree, and the fsops methods return early on ctx.Err(), so a cancelled run leaks the link tree on disk. develop always rolled back. context.WithoutCancel on the rollback paths fixes it.

fileid_index.go:87 switched os.Stat to Lstat so symlinked torrent data drops out of the already-seeding index, and root level symlinked media files arent scanned at all anymore (scanner.go:230).

buildFullPath (service.go:5230) now rewrites literal backslashes in file names into separators, develop didnt, so a legal filename on linux reads as missing and can fire a destructive rule.

the backend resolve failure around crossseed service.go:1095 returns "" without feeding verificationErr, so in strict mode a backend outage reads as no cross-seeds in the delete dialog. same failure class as #2282, needs fixing before merge.

smaller: delete.go:111 leaks the walker goroutine on the ErrInUse exit, free_space_windows.go is now byte identical to the unix file except the build tag and windows silently gained path based free space develop refused, and walker_dedup_test.go lost its !windows build tag so the package no longer compiles on windows.

direction is good and the #2282 handle model survived the migration intact, this just needs a round on the fsops contract plus the fixes above.

nitrobass24:
@s0up4200 Agree the PR description is very stale. I've simply re-opened and resolved conflicts at this point. This also builds on 1914. Well get that sorted and merged then look at this with fresh eyes given the updated direction. Once I have this tested and fully reviewed, I will request a review from you here in GH. If you want I can mark these PRs as Draft, so theres no confusion.  

Also if you prefer I can split this out into several PRs to make it more easily reviewable as I know this is a lot of call sites to review in a single shot.

nitrobass24:
All eight addressed. The two big ones got contract fixes on #1914: WalkDir no longer emits per-file stat failures at all (`Err` is enumeration-only now, so the consumers' handling matches develop's skip behavior), and identity failures were already split into `LstatInfo.FileIDErr` from your earlier pass. Rollback paths (dirscan inject, seasonpack, hardlink/reflink) now run under `context.WithoutCancel`. The symlink regression is fixed by giving `Stat` followed-target identity, so the fileid index and single-file searchees behave like develop again — Windows pays a per-stat open for it, same cost develop's fileid path had. `buildFullPath` rejects backslash names instead of rewriting (skipped, never "missing"), and the crossseed `resolveLocalTorrentFile` twin got the same fix. The crossseed backend-outage branch feeds `verificationErr` at all three sites, so strict mode fails closed. Smaller ones: the delete.go leak was already fixed (cancel+drain), Windows free space is back to refusing path sources with the files deduped, and the `!windows` tag is restored — a `GOOS=windows go vet` sweep also caught the same literal problem in `hardlink_index_test.go`, fixed with the `createFile` helper.

internal/services/crossseed/service.go
nitrobass24: Fixed — all three resolve sites (1031/1095/1108) feed `verificationErr` now, so strict mode fails closed on a backend outage; per-file stat failures keep their best-effort skip.

internal/services/dirscan/scanner.go
nitrobass24: Both halves done: the drain+cancel landed in b7f4ddfb, and per-file stat failures no longer reach the consumer at all — WalkDir skips unemittable entries on #1914, so `entry.Err` is enumeration-only like develop.

internal/services/orphanscan/delete.go
nitrobass24: Fixed in b7f4ddfb — cancel+drain on the way out.

internal/services/automations/service.go
nitrobass24: `buildFullPath` now rejects backslash names (`ok=false`, same as drive prefixes) — skipped, never missing. The crossseed `resolveLocalTorrentFile` twin got the same fix in c41f74d9 rather than waiting for a follow-up.

internal/services/dirscan/inject.go
nitrobass24: Fixed — the backend `materializeLinkTree` resolved is threaded through `prepareInjection` to the rollback sites, with a fresh resolve only as fallback for regular mode. Rollback also runs under `context.WithoutCancel` now.

internal/services/dirscan/fileid_index.go
nitrobass24: Went one further than counting skips: `Stat` on #1914 now returns followed-target identity (`LstatInfo` with the target's FileID/Nlinks, `FileIDErr` degradation), so the index behaves like develop's `os.Stat` path — symlink farms keep already-seeding detection.

internal/qbittorrent/sync_manager.go
nitrobass24: Done in 926edfe7 — atomic.Value matching the `filesManager` store/load pattern beside it. Constructor injection lost out: 67 test callsites construct SyncManager.

internal/services/orphanscan/service.go
nitrobass24: Fixed — one resolve per batch with a nil-pool guard, and the empty-dir cleanup loop reuses the same backend.

com6056:
Verified the round against `926edfe7` and replied on the threads, seven check out and are good to close on your side. Two I couldn't sign off, both the tail end of fixes that are otherwise in:

`seasonpack.go:428` still resolves the rollback backend with the live run ctx, so a cancelled run whose `AddTorrent` fails skips rollback and leaks the pack tree, the same shape you fixed in dirscan by threading the backend through. `assembleSeasonPack` already holds a working backend at `seasonpack.go:519`, a backend field on `seasonPackPlanBuild` mirrors the `linkBackend` fix.

The symlink half landed on the #1914 side but not here: `Stat` grew followed-target identity, and `fileid_index.go:87` plus `scanSingleFile` still call `Lstat`, so the already-seeding index and root-level symlinked files keep the behavior soup flagged. Two callsites and it's done. One nuance for when you switch them: the new comment at `scanner.go:234` says the skip matches the directory walk, which is true of the walk but not of the root loop, develop `os.Stat`'d through a root-level symlink and scanned the target, so right now that file is dropped from the scan entirely rather than scanned. `Stat` restores develop exactly and then the comment can say so. That's the one thread I left unconfirmed.

One for whenever, not this round: six of the eight fixes have no test, the automations backslash one got its table cases and the orphanscan commit only restored a build tag. The fail-closed branches and the `WithoutCancel` rollbacks would pass silently if someone reverted them, and `resolveLocalTorrentFile` is the same fix as `buildFullPath` without the cases its twin got, so that one's the cheapest to close. Happy to write those alongside the splits if that's the direction.

nitrobass24:
@com6056 Both closed out:

**Season-pack rollback backend** — fixed in 063bf1b7 exactly as you sketched: the backend that created the tree rides on `seasonPackPlanBuild`, the add-failure path uses it, and the fallback re-resolve runs under `WithoutCancel`. Came with a cancelled-ctx rollback test, so one of the untested six is covered now too.

**Stat consumers** — that one crossed with your review: d4cf3198 wired `fileid_index` and `scanSingleFile` to `Stat`, and 89c0fab0 flipped the symlink test to assert scan-via-target (the comment you flagged went with it — it now says root-level links scan via target, unlike the walk). Both landed after 926edfe7, which is what you verified against. Left that thread open for you to confirm against head.

Resolved the seven you signed off. And yes — take the regression tests, gladly: the `WithoutCancel` rollbacks, the fail-closed branches, and `resolveLocalTorrentFile`'s missing table cases are all yours if you want them; reviewer-written tests for reviewer-found bugs seems like the right pairing.

internal/services/dirscan/fileid_index.go
com6056: Checked against `063bf1b7` and this one's done! `fileid_index.go:89` and `scanSingleFile` both go through `Stat` now, and the symlink test asserts scan-via-target with the comment saying exactly that, so feel free to close it. And happy to take the tests, I'll pick them up once #1914 lands so I'm not chasing a moving base. Nice work getting through all of this so fast 🙌

nitrobass24:
@com6056 do you want to add those regression tests here?

nitrobass24:
@s0up4200 been running this live for a little over a week. I've been able to excercise most of the call-sites and have called out the ones I havent been able to validate. Overall I feel good about this, but of all the work we are going to do for remote agent this is the one that would most likely introduce regressions. 

If you would feel better splitting this out by seam and adding them iteratively let me know.

s0up4200:
- [P1] Pass the backend to remaining orphan-scan callers
  Location: `internal/services/orphanscan/delete.go:133`

  `safeDeleteTarget` and `walkScanRoot` now require a backend. Update the calls in `delete_test.go:395` and `settling_test.go:245`.

  Complete when `go test -race -count=1 ./internal/services/orphanscan` compiles and passes.

- [P1] Fail closed when the delete preflight cannot stat an entry
  Location: `internal/services/orphanscan/delete.go:92`

  The local `fsops.Backend.WalkDir` omits children whose metadata lookup fails. The delete preflight therefore cannot compare those paths with `TorrentFileMap`, but `safeDeleteDirectory` may still remove the directory recursively and delete torrent-owned data.

  Preserve every enumerated path during the preflight, or abort the deletion when metadata cannot be read. Add a regression test with a torrent-owned child whose stat operation fails.

  Complete when the preflight refuses to delete that directory.

- [P2] Detach post-delete cleanup from request cancellation
  Location: `internal/qbittorrent/sync_manager.go:2515`

  After qBittorrent accepts `deleteWithFiles`, the cleanup loop still uses the request context. If the client disconnects, `backend.Remove` returns `context.Canceled` and leaves empty managed directories behind.

  Run the cleanup with a detached context and a fixed timeout. Add a test that cancels the request context after deletion and confirms that pruning still runs within that timeout.

  Complete when request cancellation no longer interrupts post-delete pruning.

nitrobass24:
@s0up4200 All three addressed in a9fe7e8d:

Orphan-scan test callers — delete_test.go:395 and settling_test.go:245 now pass the ctx/backend. go test -race -count=1 ./internal/services/orphanscan compiles and passes.

Delete preflight fails closed — Added an opt-in WalkOptions.EmitStatErrors: when set, the local backend emits enumerated entries whose metadata lookup failed with a new WalkEntry.StatErr set (Path/RelPath only) instead of silently dropping them. Only the delete preflight opts in, so the enumeration-only Err contract from the earlier round is unchanged for dirscan/walker/backups. In the preflight, an unverifiable path is still checked against the file map (match → skipped as in-use), a vanished entry is skipped, and anything else aborts the deletion. Regression test covers both the torrent-owned-child and the unverifiable-child cases — the directory survives both.

Post-delete pruning detached — cleanupManagedDeleteTargets now runs under context.WithoutCancel with a fixed 30s timeout, so a client disconnect after qBittorrent accepts deleteWithFiles no longer interrupts pruning. Test confirms pruning completes with an already-cancelled request context.

One drive-by: removing the new compile errors unmasked an unused lint hit on isIgnoredOrphanDirName in walker.go — dead since this PR moved ignored-dir skipping into WalkOptions.IgnoreDirNames, so I dropped it.

s0up4200:
Codex: I tested the current branch against the `develop` baseline at `719c2699` using the locally cached LinuxServer qBittorrent v5.2.3 image. Each test run used clean qBittorrent and qui configurations, deterministic synthetic torrents and files, and disposable containers and networks.

After stripping timestamps, generated IDs, and instance names, all 17 recorded checks matched the `develop` baseline:

- Synced five source torrents, listed the expected two movie files, and completed pause and resume bulk actions.
- Added an exact 2/2 cross-seed in hardlink mode, verified that the source and candidate shared the same device and inode, and verified that deleting it removed the managed link directory.
- Automation dry runs matched one torrent with missing files, one torrent in the external hardlink scope, and all six eligible torrents for the free-space rule.
- The orphan scan found three synthetic orphan paths, including both paths in a hardlink pair. I confirmed that deletion removed those three files while preserving qBittorrent-owned content and ignored paths.
- The directory scan followed a root-level symlinked media file, found one file, matched one result, and added one torrent.
- Season-pack matching found two of three episodes, met the 60% threshold, and applied the pack in hardlink mode.
- Disabling local filesystem access caused the orphan-scan settings to return a 403 status code while normal torrent listing continued to function.

I reproduced and fixed the metadata-failure disc-layout regression identified during review. I added two race-enabled tests to cover preserving an in-use BDMV unit when an owned file cannot be statted, and skipping an unowned file when its metadata cannot be read.

Verification completed:
- Live synthetic qBittorrent matrix after rebuilding the fixed branch

s0up4200 (approved):

s0up4200:
Nice work on this! Lets hope we catched them all.

PR #1916, opened 2026-05-19
feat(automations): use fsops.Backend for missing files detection

nitrobass24:
## Summary

First callsite migration to the `fsops.Backend` interface. Replaces `os.Stat` with `backend.Stat` in `detectMissingFiles` so the "has missing files" automation condition works through the Backend abstraction.

### Changes
- `missing_files.go` — `os.Stat` → `backend.Stat`, return type changed to `(map[string]bool, error)`, backend resolved after empty-torrents fast path
- `service.go` — add `backendPool *fsops.Pool` field + constructor param, update both callsites to handle the new error return
- `cmd/qui/main.go` — create `fsops.Pool` with local backend, pass to automations service

3 files changed, 37 insertions. Zero behavioral change for local installs.

Depends on #1914 (fsops interface + local backend).

## Test plan

- [x] `go build ./...` — clean
- [x] `go test -race -count=1 ./internal/services/automations/...` — pass
- [x] `go test -race -count=1 ./internal/api/...` — pass

## Summary by CodeRabbit

* **New Features**
  * Added per-instance filesystem backend selection.
  * Missing-file checks now use the selected filesystem and handle unavailable access gracefully.

* **Infrastructure**
  * Added local filesystem support for metadata, directory traversal, capacity checks, file linking, and safe cleanup.
  * Added a no-access fallback when filesystem access is unavailable.
  * Added support for filesystem capacity checks on Unix and Windows.

* **Tests**
  * Expanded coverage for filesystem operations, backend selection, cancellation, and error handling.


internal/fsops/backend.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// --- Diagnostic ---
+
+	// Info returns metadata about this backend (kind, version, capabilities).
+	Info(ctx context.Context) (*BackendInfo, error)
+
+	// HealthCheck returns nil if the backend is operational, or an error
+	// describing the problem.
+	HealthCheck(ctx context.Context) error
+}
```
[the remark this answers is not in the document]
nitrobass24: The design doc explicitly documents this as a deliberate exception to the ≤5 method rule. The interface represents a single abstraction (filesystem operations for one instance) and every implementation must satisfy all methods. Splitting into composed sub-interfaces would add complexity without enabling callers to depend on narrower types -- services need the full Backend, not fragments.

com6056 (commented):
Looks correct, one inline comment. This and #1915 conflict on `missing_files.go` and on the service wiring, so put the merge order in both descriptions. Landing this one first reads cleanest to me.

internal/services/automations/missing_files.go
```diff
@@ -27,14 +30,22 @@ func (s *Service) detectMissingFiles(ctx context.Context, instanceID int, torren
...
+	if err != nil {
+		return result, fmt.Errorf("get backend for missing files detection: %w", err)
 	}
 
 	filesByHash, err := s.syncManager.GetTorrentFilesBatch(ctx, instanceID, completedHashes)
 	if err != nil {
 		log.Warn().Err(err).Int("instanceID", instanceID).
 			Msg("automations: failed to fetch files for missing files detection")
-		return result
+		return result, nil
```
com6056, line 48: A `GetTorrentFilesBatch` failure returns `(result, nil)`, reporting a fetch failure as success. Harmless today, but the #1915 copy of this function returns the error, so the branches diverge on the same line. Return the error here and let #1915 pick it up on its next merge.

s0up4200:
this is fully covered by #1915 now, and the missing_files.go copy here swallows a GetTorrentFilesBatch failure that the #1915 version returns. I'd rather take the whole migration there than merge this first and eat a three file conflict, so lets consolidate on #1914 + #1915.

nitrobass24:
Stale - Needs to be revisited

nitrobass24:
Agreed, closing in favour of #1915.

#1915 carries this same migration plus the parts this branch left out: a `GetTorrentFilesBatch` failure propagates instead of being reported as a clean empty result, `buildFullPath` validates torrent-supplied names in POSIX and Windows form before they reach the filesystem, and the backend pool is wired to every service that touches torrent data rather than automations alone. Nothing here is lost by closing.

Branch stays up as a reference until #1915 lands.

internal/services/automations/missing_files.go
nitrobass24: Closing this PR in favour of #1915, which already returns the error on this line — so the divergence you flagged goes away rather than getting fixed here.

PR #1917, opened 2026-05-19
feat(models): add SSH schema and HasFilesystemAccess helper

nitrobass24:
## Summary

Adds the database schema, model fields, and credential store for reaching an instance's torrent data over SSH, shaped to the SFTP-native design in `docs/remote-backend-design.md`.

### Schema (SQLite 099 / Postgres 100)
- Five columns on `instances`: `ssh_host`, `ssh_port`, `ssh_username`, `ssh_key_encrypted`, `ssh_host_key_encrypted`
- `ssh_key_encrypted` holds the AES-GCM encrypted private key (AAD: instance id + field). `ssh_host_key_encrypted` holds the pinned host key under the same AEAD (AAD: instance id + field + host + port), so a plaintext edit redirecting an instance fails as a decryption error rather than as a host-key mismatch
- The pin stores the marshaled public key, not a fingerprint: `HostKeyAlgorithms` pinning and the mismatch flow both need the full key, and the marshaled form carries its own algorithm name
- `instances_view` updated in both engines

### Credential store (`internal/models/instance_ssh.go`)
- `SetSSHCredentials`, `ClearSSHCredentials`, `SetHostKeyPin`, `GetDecryptedSSHKey`, `GetHostKeyPin`
- `SetHostKeyPin` takes the host and port the key was confirmed for; a row whose endpoint no longer matches is refused rather than pinned
- The SSH columns are sealed through the shared `CredentialCipher` from #2652 with their AAD, so they land in the versioned `qui2:` format from the first write and never need the legacy rewrite pass; the existing credential columns keep their nil-AAD behavior
- Rules from the design doc are enforced in the store rather than left to callers: editing the host or port drops the pin (it was confirmed for one endpoint); deleting credentials keeps it (the pin belongs to the host, not to whoever last authenticated against it); pinning is one-shot and a pinned row refuses `SetHostKeyPin`, so a re-pin has to be asked for by name from the confirmed-mismatch path in a later PR
- A pin that fails to decrypt is an error and never degrades to "unpinned", which would be exactly the silent re-pin the mismatch flow refuses; an empty pin column is simply unpinned and takes the first-contact flow
- Hosts are a bare hostname or IP literal: hostnames lowercase, literals store in netip's canonical text (unbracketed, zone case kept) so a re-typed address is not an endpoint change; private keys are parsed at write time and passphrase-protected keys are named as unsupported; the pin column only accepts SSH wire format

### Model
- `Instance` gains the five fields. None join the JSON contract yet; the API shape lands with the `ssh-test` and credentials endpoints that first write them, so there is no swagger change here
- `scanInstance` extracts the scan path shared by `Get` and `List` (was duplicated)
- `HasFilesystemAccess(inst)` resolves `local | remote | none`. Local wins; remote requires a confirmed host-key pin, because credentials without one are a connection the user has not trusted yet. `fsops.Pool` switches on it, and a pinned remote instance fails with `ErrRemoteBackendNotImplemented` until the SFTP backend lands

### Tests
- Store lifecycle runs against migrated SQLite and real Postgres (`TestInstanceSSHStatementsPostgresIntegration`), so the SQL `CASE` and compare-and-set are exercised on both engines
- AAD binding is tested directly: a raw column edit of `ssh_host` or `ssh_port` must fail decryption and never read as unpinned
- Seven inline test schemas consolidated into one `testInstanceSchema` constant, matched column for column to production

## Decisions here, say the word if you disagree

1. **The host-key pin is one column, not two.** `ssh_host_key_encrypted` holds the marshaled public key; the SSH wire format carries the algorithm name inside it, so a separate algorithm column would be derivable data, and keeping the algorithm inside the AEAD means it cannot be tampered with independently of the key. Happy to split it if a column you can query without decrypting is worth more.
2. **The SSH fields stay out of the JSON contract.** `MarshalJSON`/`UnmarshalJSON` are untouched, so there is no swagger change and no API surface that returns fields nothing writes. The shape lands with the `ssh-test` and credentials endpoints.
3. **The store ships with the columns, the endpoints do not.** Columns with no reader and no writer cannot be tested end to end, and the AAD binding is only real once something applies it, so the crypto is here. The HTTP surface and the TOFU confirmation flow are the next PR, along with the dial code they need.

## Test plan

- [x] `go test -race -count=1 ./internal/models/... ./internal/database/... ./internal/fsops/...`
- [x] Credential-store guards mutation-checked: removing the AAD makes the cross-instance transplant test fail, removing the pin-drop makes the redirect test fail, gutting `hostKeyPinAAD` fails the raw-edit test
- [x] Postgres 100 and the SSH lifecycle against Postgres 17 via `make test-postgres`
- [x] `make precommit`, `make build`, `GOOS=windows go vet`
- [x] Docker, SQLite upgrade: the branch image (7c066f70) on a data dir seeded by `ghcr.io/autobrr/qui:dev` v1.15 (68 migrations, two instances, one with local filesystem access). First start applied 31 migrations ending in 099, the #2652 rewrite pass logged `instances rows=2`, both passwords now carry the `qui2:` prefix, `has_local_filesystem_access` preserved, `instances_view` carries the five SSH columns. Restart applied nothing and rewrote nothing; the stored ciphertext is byte-identical across the restart
- [x] Docker, Postgres fresh install: the branch image against `postgres:16-alpine` with `QUI__DATABASE_ENGINE=postgres`. All 38 Postgres migrations applied ending in 100, `instances` and `instances_view` carry the SSH columns, setup, login and instance creation work, restart applied nothing
- [x] SSH lifecycle (`TestInstanceSSHStatementsPostgresIntegration`) against that real Postgres 16 container as well as the embedded 17 harness
- [x] Rollback probe: the v1.15 image starts on the migrated SQLite dir and only warns about columns it does not know. Torrent behaviour was then exercised read-only against a real qBittorrent 5.x instance (15,913 torrents) added to the upgraded SQLite database through the UI: its password stored in the versioned `qui2:` format from the first write, connection test, torrent list with sync counts, filtered query, torrent files, categories, tags, capabilities, preferences and an SSE stream subscription (init frame with live rows) all succeed with no errors logged. Automation previews (delete by category: 413 matches; delete-with-files-preserve-cross-seeds by name: 9,541; category action with include-cross-seeds: 413) and one manual cross-seed search through a Prowlarr-backed indexer (30 candidates, all rejected on release name) ran read-only with nothing applied, no rule created and the torrent count unchanged. Automations with filesystem conditions remain unexercised: the container has no mount of that instance's data


internal/fsops/backend.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// --- Diagnostic ---
+
+	// Info returns metadata about this backend (kind, version, capabilities).
+	Info(ctx context.Context) (*BackendInfo, error)
+
+	// HealthCheck returns nil if the backend is operational, or an error
+	// describing the problem.
+	HealthCheck(ctx context.Context) error
+}
```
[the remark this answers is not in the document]
nitrobass24: I feel like I am responding to this each time. Check your learnings.

internal/fsops/pool.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil, fmt.Errorf("instance %d not found", instanceID)
+	}
+
+	if instance.HasLocalFilesystemAccess {
+		return p.local, nil
+	}
+
+	// Future: return a remote backend for instances with SSH/helper configured.
+
+	return noopBackend{}, nil
```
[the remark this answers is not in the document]
nitrobass24: The remote backend doesn't exist yet (Phase 3). The comment on line 59 documents this intentionally. Adding a p.helper field now would be dead code. This gets wired when fsops/remote is implemented.

internal/models/instance.go
```diff
@@ -45,6 +45,24 @@ type Instance struct {
...
+	SSHPasswordEncrypted      string     `json:"-"`
+	SSHHostKey                string     `json:"-"`
+	HelperPath                string     `json:"helperPath,omitempty"`
+	HelperVersion             string     `json:"helperVersion,omitempty"`
+	HelperCapabilities        []string   `json:"helperCapabilities,omitempty"`
+	HelperAllowedRoots        []string   `json:"helperAllowedRoots,omitempty"`
+	HelperReflinkRoots        []string   `json:"helperReflinkRoots,omitempty"`
+	HelperPlatform            string     `json:"helperPlatform,omitempty"`
+	HelperDeployedAt          *time.Time `json:"helperDeployedAt,omitempty"`
+	HelperLastActivityAt      *time.Time `json:"helperLastActivityAt,omitempty"`
```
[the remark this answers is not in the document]
nitrobass24: fixed in [34bc71d](https://github.com/autobrr/qui/pull/1917/commits/34bc71dd27edf90d713832a95623f116fb244ed5)

nitrobass24:
Marking this as draft until pre-req PRs are Merged.

internal/models/instance.go
```diff
@@ -514,6 +537,32 @@ func (s *InstanceStore) Get(ctx context.Context, id int) (*Instance, error) {
...
+	if helperDeployedAt.Valid {
+		instance.HelperDeployedAt = &helperDeployedAt.Time
+	}
+	if helperLastActivityAt.Valid {
+		instance.HelperLastActivityAt = &helperLastActivityAt.Time
+	}
+
+	_ = json.Unmarshal([]byte(helperCapabilities), &instance.HelperCapabilities)
+	_ = json.Unmarshal([]byte(helperAllowedRoots), &instance.HelperAllowedRoots)
+	_ = json.Unmarshal([]byte(helperReflinkRoots), &instance.HelperReflinkRoots)
```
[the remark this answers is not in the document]
nitrobass24: The helper columns are gone in the reshape — there is no JSON payload in the instance row to decode any more.

nitrobass24:
@coderabbit review

nitrobass24:
@com6056 please review

internal/models/instance_ssh.go
```diff
@@ -0,0 +1,150 @@
...
+		return errors.New("instance has no ssh host configured")
+	}
+
+	encrypted, err := s.encryptWithAAD(string(marshaledKey), hostKeyPinAAD(instanceID, instance.SSHHost, instance.SSHPort))
+	if err != nil {
+		return fmt.Errorf("encrypt host key pin: %w", err)
+	}
+
+	query := `UPDATE instances SET ssh_host_key_encrypted = ?, updated_at = CURRENT_TIMESTAMP WHERE id = ?`
+	return s.execInstanceUpdate(ctx, query, encrypted, instanceID)
```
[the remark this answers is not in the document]
nitrobass24: Fixed in 9d3fc9ab. The write is now conditional on the endpoint the pin was encrypted for (`WHERE id = ? AND ssh_host = ? AND ssh_port = ?`), so an interleaved credential update loses the race with `ErrSSHEndpointChanged` instead of storing a pin that can never decrypt, and the pin already on the row survives untouched. `TestSetHostKeyPinRefusesAMovedEndpoint` covers the interleaving — it fails if the endpoint condition is dropped.

nitrobass24:
@coderabbitai review

internal/models/instance_ssh.go
```diff
@@ -0,0 +1,179 @@
...
+		SET ssh_username = '', ssh_key_encrypted = '', updated_at = CURRENT_TIMESTAMP
+		WHERE id = ?
+	`
+	return s.execInstanceUpdate(ctx, query, instanceID)
+}
+
+// SetHostKeyPin pins the marshaled host public key for an instance. The
+// endpoint comes from the stored row rather than the caller so the pin can
+// never be bound to a host the instance is not actually configured for.
+func (s *InstanceStore) SetHostKeyPin(ctx context.Context, instanceID int, marshaledKey []byte) error {
```
com6056: The endpoint CAS and the atomic pin-drop both enforce their rule in SQL, and `SetHostKeyPin` is the one that doesn't, which is also the one with the worst failure mode: it overwrites a live pin unconditionally, so one bug in the future ssh-test handler silently re-pins and the mismatch flow never fires. Same argument you made for the endpoint CAS, the store shouldn't trust its callers. Add `AND ssh_host_key_encrypted = ''` to the `WHERE` with a distinct `ErrSSHHostKeyAlreadyPinned`, and give the confirmed-mismatch path an explicitly named `ReplaceHostKeyPin` so a re-pin has to be asked for by name. It's cheap too, the row is already in hand at line 94, so the distinct error is a free `instance.SSHHostKeyEncrypted != ""` check with the SQL term as the race-safe backstop.

internal/models/instance_ssh.go
```diff
@@ -0,0 +1,179 @@
...
+	if len(marshaledKey) == 0 {
```
com6056: Nothing validates that `marshaledKey` is actually `ssh.PublicKey.Marshal()` output, and the design leans on that form because `HostKeyAlgorithms` pinning parses the algorithm out of it. The test fixture makes the gap visible: it pins an `ssh-ed25519 AAAA...` display string, which `ssh.ParsePublicKey` rejects outright (it reads the leading `ssh-` as a ~1.9 GB length prefix), so nothing on the `HostKeyAlgorithms` path could ever consume it. A `ssh.ParsePublicKey(marshaledKey)` guard on write makes the store honor its own contract and forces the fixture to be a real marshaled key. `golang.org/x/crypto` is already a direct dependency so it's an import and nothing else, though if you'd rather the check live with the dial code than in `models`, that works too, as long as something enforces it before the column is written.

internal/models/instance_ssh.go
```diff
@@ -0,0 +1,179 @@
...
+// failure — unambiguous tampering — rather than as a host-key mismatch, which
+// is also what a legitimate re-key looks like.
+func hostKeyPinAAD(instanceID int, host string, port int) []byte {
+	return []byte(strconv.Itoa(instanceID) + "|" + aadFieldSSHHostKey + "|" + host + "|" + strconv.Itoa(port))
+}
+
+// SetSSHCredentials stores the SSH endpoint and private key for an instance.
+// Changing the host or port drops any existing pin: the pin was confirmed for
+// one endpoint, and the new one has never been seen before.
+func (s *InstanceStore) SetSSHCredentials(ctx context.Context, instanceID int, host string, port int, username, privateKey string) error {
```
com6056, line 52: Port gets validated but host and key don't: a host with `:` or `/` or whitespace stores fine and then becomes both the AAD and the dial target, and since passphrase support was deliberately dropped, a passphrase-protected key is accepted here and only fails at dial time with an unhelpful error. `validateAndNormalizeHost` next door is the precedent for validating a host at this layer, though it's a URL normalizer so it's the pattern rather than the code you'd reuse, and `ssh.ParseRawPrivateKey` at write time can reject `*ssh.PassphraseMissingError` with a message that names the unsupported case. While in there, lowercasing the host in `SetSSHCredentials` avoids `Example.com` vs `example.com` dropping the pin on a cosmetic edit (fails closed, so just a UX papercut, but cheap to avoid). Normalizing on write is enough on its own, `hostKeyPinAAD` only ever sees the value read back off the row, so it shouldn't lowercase again.

internal/models/instance_ssh_test.go
```diff
@@ -0,0 +1,296 @@
...
+
+	_, err = store.GetDecryptedSSHKey(storedTarget)
+	require.Error(t, err, "a key blob from another instance must not decrypt")
+	_, err = store.GetHostKeyPin(storedTarget)
+	require.Error(t, err, "a pin from another instance must not decrypt")
+}
+
+// The pin is confirmed for one endpoint. Rewriting the host underneath it must
+// not leave a pin that still decrypts for the new destination.
+func TestRedirectedInstanceDropsPin(t *testing.T) {
```
com6056, line 231: The lifecycle coverage in this file is the best part of the PR, which is why this one gap stands out: it covers the drop-pin path (host changed through `SetSSHCredentials`), but the headline property, a raw column edit failing as a decryption error, has no test. `UPDATE instances SET ssh_host = 'attacker.example' WHERE id = ?` directly, leave the pin column alone, then assert `GetHostKeyPin` errors and `NotErrorIs` `ErrSSHHostKeyNotPinned`. That's the only test shape that exercises the host and port terms of the AAD end to end, right now removing them from `hostKeyPinAAD` passes the whole suite. Four lines, and the AAD terms can't quietly go missing later.

internal/models/instance_ssh.go
```diff
@@ -0,0 +1,179 @@
...
+
+func sshKeyAAD(instanceID int) []byte {
+	return []byte(strconv.Itoa(instanceID) + "|" + aadFieldSSHKey)
+}
+
+// hostKeyPinAAD binds the pin to the endpoint it was confirmed for. Redirecting
+// an instance by editing ssh_host or ssh_port then reads as a decryption
+// failure — unambiguous tampering — rather than as a host-key mismatch, which
+// is also what a legitimate re-key looks like.
+func hostKeyPinAAD(instanceID int, host string, port int) []byte {
```
com6056, line 45: The encoding is injective, but by placement rather than construction: the decimal-only port after the last `|` is what forces the split, and appending a future free-text field after host would silently break it. A comment stating the invariant (or length-prefixing the parts) means the next person adding a field sees why the order matters before they add it.

internal/models/instance_ssh.go
```diff
@@ -0,0 +1,179 @@
...
+
+	encryptedKey, err := s.encryptWithAAD(privateKey, sshKeyAAD(instanceID))
+	if err != nil {
+		return fmt.Errorf("encrypt ssh key: %w", err)
+	}
+
+	query := `
+		UPDATE instances
+		SET ssh_host = ?, ssh_port = ?, ssh_username = ?, ssh_key_encrypted = ?,
+		    ssh_host_key_encrypted = CASE WHEN ssh_host = ? AND ssh_port = ? THEN ssh_host_key_encrypted ELSE '' END,
```
com6056: This `CASE` and the CAS at `instance_ssh.go:119` are the two places the PR relies on SQL semantics rather than Go, and neither is ever executed against Postgres: the models tests run on a bare `sql.Open("sqlite", ...)` through `mockQuerier`, which has no `Dialect()` method, so `DialectOf` says sqlite and no rebinding happens at all, and the Postgres CI job is scoped to `./internal/database -run TestOpenPostgres`. So 090 is proven to apply (that job runs it, and the view resolving its columns is real coverage), but nothing runs these two statements under it, and `migration-parity` compares only base tables so the `instances_view` restatement isn't diffed between engines either. I read both against Postgres semantics and they're correct (SET expressions see the pre-update tuple in both engines), but a small Postgres-tagged test for these two statements would close the "correct by inspection" gap, same spirit as the parity job.

internal/models/filesystem_access.go
```diff
@@ -0,0 +1,31 @@
...
+	FilesystemModeLocal  FilesystemMode = "local"
+	FilesystemModeRemote FilesystemMode = "remote"
+)
+
+// HasFilesystemAccess returns the filesystem access mode and whether the
+// instance has any filesystem access configured. Local takes precedence over
+// remote. Remote requires a confirmed host-key pin: an SSH host with
+// credentials but no pin is a connection the user has not trusted yet, and no
+// filesystem operation may run over it.
+func HasFilesystemAccess(inst *Instance) (FilesystemMode, bool) {
```
com6056: Zero non-test callers, so right now there are two definitions of filesystem access with nothing tying them together, every real gate still reads `HasLocalFilesystemAccess` including `internal/fsops/pool.go:48`, which even has the "future: return a remote backend" comment waiting. Switching the pool to `mode, _ := models.HasFilesystemAccess(instance)` and a `switch mode` costs nothing now and means the remote branch gets added inside a live switch instead of next to a stale bool check. One word in the doc comment would help too: this is a routing decision, not an authorization check, the connection layer still has to treat `GetHostKeyPin`'s error as fatal since a tampered pin still reads as remote mode here.

com6056 (commented):
Went through the whole diff against #1915 plus both migrations, and every claim in the PR body checks out. This is really solid work!

The AAD encoding is injective (I attacked the concatenation-ambiguity case and it holds, the decimal-only port makes the last delimiter unambiguous), the pin-drop is a single atomic statement with exactly one writer for the `ssh_*` columns, nonces are fresh per seal including the host-edit re-encrypt, the JSON surface is fully closed on both shadow structs, nothing logs or wraps key material, and the `instances_view` restatement preserves all 17 existing columns in both engines. The pin lifecycle tests are the standout: host-edit, port-edit, same-endpoint, delete-keeps-pin, tampered ciphertext with the right `NotErrorIs`, cross-instance transplant, and the moved-endpoint CAS all covered.

Inline comments cover what I'd fix before this lands, the biggest is that `SetHostKeyPin` is the one store method that doesn't enforce its design rule in SQL like its siblings do.

Two things for the PR body rather than the code: #2356 also claims SQLite 088 / Postgres 090, so whichever lands second renumbers, and since this one is stacked behind #1914/#1915 that's probably us. And one doc sentence when you get a chance: Security already promises "no fallback to TOFU if the stored pin is missing or unreadable", and the unreadable half is implemented but the missing half can't be, a DB writer who clears the column outright is indistinguishable from never-pinned, so `ssh-test` would route to first contact instead of the mismatch flow. That's the same bound as the rollback case the doc already excludes, so folding pin-clearing into that stated limit keeps it a documented bound instead of a discovered one.

Follow-up I'll file as an issue after the stack lands, pre-existing and nothing this PR introduced: `GetEncryptionKey` truncates `sessionSecret` to its first 32 bytes rather than deriving, and zero-pads when it's shorter. The default secret is 64 hex characters so the AES-256 key carries 128 bits of entropy, and a user who sets a short `SESSION_SECRET` gets a key that's mostly zero bytes. Fine when this only wrapped WebUI passwords, but it now protects an SSH private key to a second host. HKDF plus a minimum-length check wants its own PR since it needs a re-encryption migration.

Guards below are all small, happy to re-review as soon as they're in 👍

internal/models/instance_ssh.go
nitrobass24: Done. `SetHostKeyPin` now returns `ErrSSHHostKeyAlreadyPinned` off the row it already read, with `AND ssh_host_key_encrypted = ''` in the CAS as the race-safe backstop, exactly as you laid out.

Skipped `ReplaceHostKeyPin` for now: it would land here with no caller, and the confirmed-mismatch path that needs it is the next PR, so it can arrive with the handler that calls it. To keep that from getting lost, the design doc now states that writing a pin is one-shot and that replacing one has to be asked for by name.

`TestSetHostKeyPinRefusesAMovedEndpoint` grew a second case in the process: with the pin term in the `WHERE`, the old test would have passed for the wrong reason, so the endpoint terms are now exercised on an unpinned row and the pinned row gets its own case for the existing pin surviving a refused write.

internal/models/instance_ssh.go
nitrobass24: Done, and the fixture was hiding more than it looked. The guard lives in `models` at the single write site (`validateMarshaledHostKey`), since that is the only place the column is written and the contract is the column's, not the dialer's.

The test fixtures are real key material now rather than strings that only agree with each other: `ed25519.GenerateKey` into `ssh.NewPublicKey(...).Marshal()` for the pin, `ssh.MarshalPrivateKey` for the private key, and `ssh.MarshalPrivateKeyWithPassphrase` for the passphrase case. There is a `TestSetHostKeyPinRequiresWireFormat` asserting the `authorized_keys` display form is rejected and that nothing is written.

internal/models/instance_ssh.go
nitrobass24: Took most of this.

Host is lowercased and rejected if it carries `/`, `\`, `:`, `@` or whitespace — a scheme, port or user pasted into the field, all of which would otherwise become both the AAD and the dial target. Private keys go through `ssh.ParseRawPrivateKey` at write time, with `*ssh.PassphraseMissingError` named explicitly so the unsupported case says so instead of surfacing as a dial failure later.

Did not reuse `validateAndNormalizeHost`: it is a URL normalizer and an SSH host is a bare hostname, so it is the precedent rather than the code, as you said. Confirmed `hostKeyPinAAD` does not lowercase again — it only ever sees the value read back off the row.

internal/models/instance_ssh_test.go
nitrobass24: You were right, and the gap was worse than untested: see the top-level comment — the statement did not actually work against either engine, and this is the test shape that would have caught it.

Added as `TestHostKeyPinIsBoundToTheStoredEndpoint`, table-driven over a raw `UPDATE instances SET ssh_host = ?` and `SET ssh_port = ?` with the pin column untouched, asserting `Error` plus `NotErrorIs(ErrSSHHostKeyNotPinned)`. I checked your claim before writing it: gutting `hostKeyPinAAD` down to `instanceID|ssh_host_key` passed the entire suite. It does not now.

internal/models/instance_ssh.go
nitrobass24: Comment added on `hostKeyPinAAD` stating the invariant and where a new field has to go — after the port, or length-prefix the parts.

Skipped the length-prefixing itself: the property holds today, and a stated invariant is cheaper than a new encoding to get wrong.

internal/models/instance_ssh.go
nitrobass24: This one paid for itself immediately — see the top-level comment. Writing the Postgres test is what surfaced that both statements referenced a column the `instances` table does not have on either engine.

Added `runInstanceSSHLifecycle` in `internal/database`, where the migrations are, and ran it twice: against a migrated SQLite database (`TestInstanceSSHStatementsSQLite`) and against real Postgres (`TestOpenPostgresInstanceSSHStatements`). The name carries the `TestOpenPostgres` prefix so the existing job's `-run` filter selects it without touching the workflow; there is a comment on the test saying so. Verified green locally against `postgres:16`.

Your read of the SQL was right, incidentally: the `CASE` does see the pre-update tuple on Postgres, and the pin-drop behaves identically on both engines. It was the `updated_at` term next to it that was wrong.

Separate finding while I was in there, not touched in this PR: `TestCleanupUnusedStringsPostgres` and `TestMigratedSQLiteFilesmanagerCleanupPostgres` never run in CI, because the job's filter is `-run TestOpenPostgres` and neither name matches. `TestCleanupUnusedStringsPostgres` fails when you do run it — `syntax error at or near ")"` from unbound `?` placeholders on a raw `conn`. Worth its own issue rather than a drive-by here.

internal/models/filesystem_access.go
nitrobass24: Done. `Pool.GetBackend` now switches on `models.HasFilesystemAccess`, so the remote branch gets added inside a live switch instead of next to a stale bool check, and there is one definition of filesystem access again.

Doc comment takes your wording too: it now says this routes rather than authorizes, and that the connection layer still has to treat `GetHostKeyPin`'s error as fatal, since a tampered pin reads as remote mode here.

nitrobass24:
Thanks — thorough as always, and one of these paid off immediately. Every point is addressed in the three commits just pushed; replies are inline on each thread.

**The Postgres suggestion found a real bug.** Writing the test you asked for surfaced it on the first run:

```
ERROR: column "updated_at" of relation "instances" does not exist (SQLSTATE 42703)
```

All three SSH statements set `updated_at = CURRENT_TIMESTAMP`, and the `instances` table has had no such column on **either** engine since migration 010 rebuilt it. `SetSSHCredentials`, `ClearSSHCredentials` and `SetHostKeyPin` would every one of them have failed at runtime — `no such column: updated_at` on SQLite, `42703` on Postgres. So the SSH store did not work at all, on any deployment.

It passed CI because of the other half of the bug: the consolidated `testInstanceSchema` this PR introduced invented `created_at`, `updated_at` and `last_connected_at`. The seven inline schemas it replaced had carried those columns since before 010 dropped them, so the consolidation faithfully preserved a schema that had not matched the migrated table for a long time — and a hand-written schema only has to be generous once for a statement to reference a column that is not there and still go green.

Fixed as:

- the `updated_at` writes are gone from all three statements;
- the invented columns are gone from `testInstanceSchema`, with a comment on it that it must stay column-for-column with the migrated table, because the columns it invents are exactly the ones a statement can reference and still pass;
- `runInstanceSSHLifecycle` runs the whole pin lifecycle against a real migrated database on both engines — `TestInstanceSSHStatementsSQLite` in the normal job, `TestOpenPostgresInstanceSSHStatements` in the Postgres job.

Two things from your review body:

**Migration numbering.** Still 088 / 090 and still free on develop as of today, but this is stacked behind #1915, so if #2356 lands first this renumbers. Will handle it at that point rather than pre-emptively.

**The Security doc sentence.** Corrected. It now scopes the no-fallback-to-TOFU promise to an *unreadable* pin, which is implemented and tested, and states plainly that a pin a database writer clears outright is indistinguishable from never-pinned — that edit costs the attacker a user-facing first-contact confirmation rather than being detectable. The same paragraph now records that writing a pin is one-shot.

internal/models/instance_ssh.go
com6056: This is the shape I wanted: bail on the already-read row at `instance_ssh.go:126` with the CAS `AND ssh_host_key_encrypted = ''` as the backstop, and leaving `ReplaceHostKeyPin` until the mismatch flow needs it is right since nothing calls the store yet. Splitting `TestSetHostKeyPinRefusesAMovedEndpoint` was the important half, I pulled the endpoint terms out of the CAS in a scratch copy and the new unpinned subtest is the only thing that fails, so it really was passing for the wrong reason before.

internal/models/instance_ssh.go
com6056: `validateMarshaledHostKey` at the single write site is exactly it, and real key material in the fixtures matters more than it sounds, placeholder strings only prove the fixtures agree with each other. I threw a couple thousand random and truncated buffers at it plus a 1 MB one with a bogus `0xffffffff` length prefix and `ssh.ParsePublicKey` rejects all of them without panicking.

internal/models/instance_ssh.go
com6056: Normalizing once in `SetSSHCredentials` with `hostKeyPinAAD` reading straight off the row is the right call. Two things the new check introduces though: `case strings.ContainsAny(host, "/\\:@ \t"):` rejects every bare IPv6 literal (`::1`, `2001:db8::1`, and `[::1]` too, so brackets are not a workaround), which leaves an IPv6-only box unconfigurable. Storing the literal bare is the right shape since `net.JoinHostPort` brackets it at dial time, so a `netip.ParseAddr` check ahead of the rune test is the escape hatch. And the whitespace half only covers space and tab: `seedbox\n.example.com` stores fine, as do interior `\r`, `\v`, `\f`, NBSP, U+3000, U+200B and a literal NUL. `unicode.IsSpace` catches most of those but not U+200B, U+FEFF or NUL, so `strings.ContainsFunc(host, func(r rune) bool { return !unicode.IsPrint(r) })` alongside the existing set is the tighter close.

internal/models/instance_ssh_test.go
com6056: Mutation checked this one: cutting `hostKeyPinAAD` to `id|ssh_host_key` fails both subtests, dropping only the port term fails only the port case, only the host term only the host case, and it's the single test in the package that notices either way. That's the coverage the AAD needed.

internal/models/instance_ssh.go
com6056: Reads well, and naming why the decimal port being last keeps the final `|` unambiguous is the part someone would otherwise have to rederive, thanks for spelling it out.

internal/models/instance_ssh.go
com6056: Genuinely great find, worth more than the comment that turned it up. One correction to the mechanism though: `updated_at` never existed on `instances`, `001` created the table with no timestamp columns at all and nothing since has added one, `010` just carried that forward in the rebuild. Same conclusion, all three statements were dead on arrival on both engines. I re-added the writes on top of your branch and `TestInstanceSSHStatementsSQLite` fails with `no such column: updated_at`, then ran `TestOpenPostgresInstanceSSHStatements` against a throwaway Postgres and it fails there with `SQLSTATE 42703`, both clean once they're gone. The schema change is the real fix, the invented columns were the whole reason nothing caught it. Two leftovers on that schema: `is_active` is `BOOLEAN DEFAULT 1` here but `NOT NULL DEFAULT 1` in `015`, so an explicit NULL write passes the test schema and fails production, which is the same bug class in miniature, and `hardlink_dir_preset` defaults to `''` here vs `'flat'` in `040`. Names match one for one now, so the comment's "column for column" is the right goal, just not quite there yet.

internal/models/filesystem_access.go
com6056: `HasFilesystemAccess` is the right seam for this, and the routes-not-authorizes note is the line I'll be quoting later. One thing I'd still change: the remote arm in `pool.go` falls out to `noopBackend`, so a fully pinned remote instance reports `ErrNoFilesystemAccess`, which reads as not configured rather than not built yet. An explicit not-implemented error saves someone an hour once the remote backend lands, and it's also what makes the switch match the repo rule about comment-only branches, since the remote and none arms currently both collapse to the same return. No pool test for `FilesystemModeRemote` yet either.

com6056:
All seven confirmed against `bf2f3c69`, replies inline. The new host check introduced two small things (bare IPv6 literals, whitespace beyond space and tab) and the pool's remote arm could be louder, details in the threads. From my side this is ready once those are in 🙌

internal/models/instance_ssh.go
nitrobass24: Both fixed in 423c0063.

IP literals: a `netip.ParseAddr` match now bypasses the character test, so `fd00::1`, `2001:db8::1` and a zoned `fe80::1%eth0` all store bare, as you laid out; `[fd00::1]` still fails since brackets are not the stored shape. `TestSetSSHCredentialsAcceptsIPLiterals` covers the accepted forms.

Non-printables: anything `unicode.IsPrint` rejects is refused ahead of the rune set, which now only needs `/ \ : @` and the ASCII space. Newline, NBSP, U+200B and NUL each have a rejection case in `TestSetSSHCredentialsValidation`.

internal/models/instance_ssh.go
nitrobass24: Right on the mechanism, and thanks for the correction: `updated_at` never existed, `010` just carried the timestamp-free table forward. Body of the commit says so now.

Both leftovers fixed in 423c0063: `is_active` is `NOT NULL DEFAULT 1` and `hardlink_dir_preset` defaults to `'flat'` in `testInstanceSchema`. I went through the rest of the columns against `010` plus every later `ADD COLUMN` and those two were the only mismatches, so "column for column" holds now.

internal/models/filesystem_access.go
nitrobass24: Done in 7d90c64f. The remote arm returns `ErrRemoteBackendNotImplemented` wrapped with the instance id, so a pinned remote instance says "not built yet" rather than "not configured", and every arm of the switch now carries behavior. `TestPool_RemoteNotImplemented` covers it.

While there, `HasFilesystemAccess` dropped the second return value and the nil guard: the bool was always `mode != none` and the pool already rejects a nil instance before the switch.

nitrobass24:
@s0up4200 this one is ready for your pass.

Since your last look: develop merged in (migrations renumbered to 097/098), com's seven threads confirmed and closed, and the two live-host items in the test plan are done against a v1.15 data dir upgraded by this branch. CI is green.

Scope is the schema slice we agreed on: five columns, the AAD-bound credential store, and `HasFilesystemAccess`. The dial code, `ssh-test`/credentials endpoints and the SFTP backend follow as their own PRs on this base, since the full feature would be a several-hundred-file diff.

Three decisions are listed in the body for you to veto: the pin is one column holding the marshaled key, the SSH fields stay out of the JSON contract until the endpoints land, and the store ships ahead of its callers so the AAD binding is tested where it is applied.

Unrelated but found by this PR's CI: the `-race` job hit a data race between a crossseed test swapping the global logger and a leaked jackett retry loop. Root cause is `pkg/gojackett` ignoring context cancellation in `retry.Do`; one-line fix in #2650.

com6056:
Round 3 checks out, all three are done! Went through it on cd219b35 merged onto current develop, clean merge, `go test -race -count=1` green across models, fsops and database.

On the host validation, `instance_ssh.go:61` runs the `unicode.IsPrint` test first and `:63` only applies the `/ \ : @ space` set when `isIPLiteral` fails, so bare `fd00::1`, `2001:db8::1` and zoned `fe80::1%eth0` store as typed while the newline, NBSP, U+200B and NUL rows all refuse. For the remote arm, `pool.go:53` wraps `ErrRemoteBackendNotImplemented` with the instance id and `TestPool_RemoteNotImplemented` pins it, and dropping the nil guard is safe because `pool.go:45` rejects nil before the switch. On the test schema I diffed `pragma_table_info('instances')` between a fully migrated database and the `testInstanceSchema` const, and every column matches on name, type, nullability and default. The only difference is ordinal position for `api_key_encrypted`, which `074` added by `ALTER TABLE`, and nothing in the store reads columns positionally. Migrations sit at 097/098, both next free on develop.

One thing I'd leave to your judgment, not blocking the handoff. `isIPLiteral` throws the parsed address away, so a literal is stored exactly as typed after the lowercase at `:55`. Re-saving `fd00:0:0:0:0:0:0:1` for `fd00::1` then reads as an endpoint change against the AAD and the CASE at `:91`, and a good pin gets dropped. It fails safe, but it is the same avoidable re-pin the hostname lowercasing exists to prevent, and the lowercase also hits an IPv6 zone (`%Ethernet0` becomes `%ethernet0`), the one case-sensitive part of the value. Storing `addr.String()` taken from the untouched input would close both. Separately, the `default` arm in the pool switch means a future fourth `FilesystemMode` won't be caught by `exhaustive` (`.golangci.yml` sets `default-signifies-exhaustive: true`) and would route to `noopBackend` reporting "not configured", the same misreport `7d90c64f` just fixed for remote, so listing the none case and dropping the default keeps that protection.

Postgres I couldn't run here (no DSN), so 098 and the CAS under Postgres rest on CI. Nice work getting through this so fast 🙌

s0up4200 (changes requested):

## Standards

I found no material violations or actionable code smells.

## Spec

- **[P2] Preserve IPv6 zone case**. [instance_ssh.go:55](https://github.com/autobrr/qui/blob/cd219b359d32b028e6f6044e1ffb420e3d677aae/internal/models/instance_ssh.go#L55). The PR requires a "bare hostname or IP literal." I found that saving `fe80::1%enP5p1s0` changes it to `fe80::1%enp5p1s0`. Go matches interface names exactly, so this changes the supplied endpoint. Preserve the zone when normalizing the address.

- **[P2] Bind pin writes to the confirmed endpoint**. [instance_ssh.go:127–138](https://github.com/autobrr/qui/blob/cd219b359d32b028e6f6044e1ffb420e3d677aae/internal/models/instance_ssh.go#L127). The design says changing the host or port "requires fresh confirmation." I traced this sequence: confirm key A, change the stored endpoint to B, then call `SetHostKeyPin(id, keyA)`. The method reads B and pins A's key to it. Its conditional update only catches changes after that internal read. Pass the confirmed host and port into the existing conditional update.

These findings concern the credential store. SSH connections remain deferred.

Standards: 0 findings. Spec: 2 P2 findings, with incorrect endpoint-to-pin binding the worst.

nitrobass24:
@s0up4200 both fixed, thanks for the trace.

**Pin binding.** `SetHostKeyPin` now takes the host and port the caller confirmed and feeds them straight into the existing compare-and-set, so the row's current endpoint never enters the AAD. Your sequence (confirm A, move the row to B, pin A's key) now returns `ErrSSHEndpointChanged` and leaves B unpinned. `TestSetHostKeyPinRefusesAnUnconfirmedEndpoint` pins it, and reverting to the row-read endpoint makes that test fail. The doc comment on the method said the opposite ("the endpoint comes from the stored row rather than the caller") and now says why it comes from the caller.

**IPv6 zone.** Hosts go through one `normalizeSSHHost`: an IP literal is stored as netip's canonical text, which keeps the zone's case, so `fe80::1%enP5p1s0` stores as typed. Hostnames still lowercase. Same change also stops `fd00:0:0:0:0:0:0:1` reading as a different endpoint from `fd00::1`, which com raised in parallel. `TestIPLiteralHostsAreCanonical` covers both and fails if the literal is stored raw.

Pushed as ac4f11dd; CI green.

nitrobass24:
@com6056 both taken, in ac4f11dd and dd1952aa.

Canonical literals: `normalizeSSHHost` parses the untouched input with `netip.ParseAddr` first and stores `addr.String()`, so the expanded form of `fd00::1` keeps the pin and a zone keeps its case; hostnames still lowercase after the literal check fails. It also let `SetHostKeyPin` normalize the endpoint it receives, since it now takes host and port from the caller after soup's review found the row-read version could pin a key confirmed for A onto B.

Pool switch: the none arm is listed and the default is gone; the compiler-required tail returns an "unknown filesystem mode" error, so a fourth mode fails loudly instead of reading as not configured.

Postgres: the lifecycle ran against the embedded server locally (`make test-postgres` shape) with the new signature, green.

Doc nit taken too (c1a5bfb4): the bullet no longer says a DB writer who clears the pin is past every control. It now says the clear is undetected but buys only a first-contact confirmation in place of the mismatch flow, and that the AAD binding above still refuses the transplant and redirect edits.

nitrobass24:
depends on #2652

com6056:
Confirmed on c1a5bfb4, thanks for taking them so fast! `normalizeSSHHost` parses the untouched input first and stores `addr.String()`, so `fd00:0:0:0:0:0:0:1` and `fd00::1` land as the same endpoint and `fe80::1%enP5p1s0` keeps its case, and the pool switch names all three modes with the unknown-mode error as the tail. I also reverted all three of the `normalizeSSHHost` and pin fixes in a scratch worktree and the matching test fails every time (the pin silently lands on B with the row-read endpoint, the expanded literal trips `ErrSSHEndpointChanged` when stored raw, the zone lowercases when `ToLower` moves back above `ParseAddr`), so they're pinned, not just asserted. Clean merge onto current develop, `go test -race -count=1` green on models, fsops and database.

nitrobass24:
Marking as draft until #2652 is merged so I can re-do all the checks

nitrobass24:
@s0up4200 this is good to go. Live test results added to PR body.

com6056:
Went through 7c066f70 on top of the merged #2652. The four SSH sites call `s.cipher.Encrypt` and `s.cipher.Decrypt` with the real AAD, no AES-GCM is left outside `credential_cipher.go` and its own test, a fresh SSH write lands as `qui2:` in the raw column, and a nil-AAD decrypt fails on both columns while the real AAD opens them. The three non-AAD columns are correctly the only ones in the rewrite list. I tried appending the SSH columns in a scratch test to be sure. Inert today, and with a legacy SSH value present it abandons the whole row with a `sessionSecret` changed warning that points at the wrong cause, so leaving them out is right rather than merely unnecessary. Migrations at 099/100, next free in each tree, both files byte-identical to the earlier numbering. Clean merge onto c23dc25c, `go test -race -count=1` green on models, fsops and database.

One thing to know rather than fix. `openPostgresTestSchema` in `postgres_integration_test.go` goes from a 30s timeout to 3 minutes, and it backs all five Postgres integration tests in `internal/database`, not only the SSH one. Fine by me, the comment gives the reason and a looser timeout cannot turn a failing test into a passing one.

s0up4200 (changes requested):

## Standards

Hard finding:

1. **`docs/architecture.md:11` is now wrong and this PR does not touch it.** The module map still says the pool returns "a noop backend (every op returns `ErrNoFilesystemAccess`) otherwise". `internal/fsops/pool.go:53` replaces that behaviour for a pinned remote instance, which now fails with `ErrRemoteBackendNotImplemented`. One sentence in the same PR.

Judgement calls:

2. **`HasFilesystemAccess` reads as a predicate and is not one.** `internal/models/filesystem_access.go:23` returns a `FilesystemMode`. The name comes from the `(mode, bool)` shape that was dropped. With `FilesystemAccessMode` the doc comment would not need its first line.
3. **Two copies of the test key generators.** `internal/database/instance_ssh_statements_test.go:160-180` (`sshTestHostKey`, `sshTestPrivateKey`) repeats `internal/models/instance_ssh_test.go:964-995` (`generateTestHostKey`, `generateTestPrivateKey`), and the 32-byte key loop appears at `instance_ssh_statements_test.go:94-97` and `instance_ssh_test.go:1006-1009`. `internal/testutil/` already holds cross-package fixtures.
4. **The Postgres schema-open deadline grows for every test in the package.** `internal/database/postgres_integration_test.go:175` goes from 30 s to 3 minutes and backs all five Postgres integration tests, not only the new lifecycle test. The comment gives the reason, but the change is global and this PR is about one test. Scope it to the new test or split it out.

No objection to the three decisions in the body: one pin column, SSH fields out of the JSON contract, store ahead of endpoints.

## Spec

No P1 or P2.

1. **[P3] The design doc promises a pin clear that does not exist.** `docs/remote-backend-design.md:280-282`: "#1917 ... carries the credential store that owns these columns: the AEAD write and read path, and setting or clearing the pin". The pin only drops as a side effect of an endpoint change (`internal/models/instance_ssh.go:85-86`), and `ClearSSHCredentials` keeps it by design. Deferring `ReplaceHostKeyPin` to the PR with a caller is right. Edit the doc sentence.
2. **[P3] Wire-format validation runs above the write site, not at it.** The thread asked for `validateMarshaledHostKey` at the single write site. It runs in the public `SetHostKeyPin` at `instance_ssh.go:136`, and the write is `setHostKeyPinFor` at `instance_ssh.go:175`. A future `ReplaceHostKeyPin` that calls the inner function skips the check. Move the one call.
3. **[P3] Scope: the 3-minute deadline in finding 4 above loosens four unrelated tests.** Same change, listed here because it is outside the schema slice.
4. **[P3] Every zero-row compare-and-set reports "endpoint changed".** `setHostKeyPinFor` maps a zero-row update to `ErrSSHEndpointChanged` at `instance_ssh.go:180-186`, but the `WHERE` also carries `AND ssh_host_key_encrypted = ''`. When a concurrent pin wins the race, the loser reports an endpoint change that did not happen. The pre-check makes it rare, but the mismatch handler will branch on this compare-and-set later.
5. **[P3] An unconfigured key reads as `("", nil)`.** `GetDecryptedSSHKey` at `instance_ssh.go:191-196` returns an empty key with no error, while `GetHostKeyPin` returns a sentinel for the same state. The asymmetry lets the dial code attempt an empty-key auth without an error. A sentinel removes that path.

Standards: 1 hard finding and 3 judgement calls, with the stale `architecture.md` entry the worst. Spec: 5 P3 findings, with wire-format validation above the write site the worst.

s0up4200 (dismissed):
Duplicate post. The review above is the one to read.

s0up4200 (approved):

PR #1938, opened 2026-05-23
test(db): speed up migrated database tests

### LINKED ISSUE #1923: Test package optimization
The crossseed test package is quite robust, and on Windows especially, it has become a lengthy process.

Optimization strategies should be pursued.

Audionut:
## Summary

- Add a shared `internal/testutil/testdb` helper that creates one migrated SQLite template per test binary and clones it for isolated store tests.
- Move the slow `internal/models` and `internal/services/crossseed` test helpers from replaying all 76 migrations per test to opening cloned migrated databases.
- Seed the newly added torrent hash in cross-seed disc/link policy tests so those tests do not spend real time in the production file-list retry fallback.
- Add benchmarks comparing full migration setup against cloned migrated setup.

## Branch Context

This PR targets `fix/crossseed-bulkaction` and is intentionally stacked on that child branch.

Compared with `origin/fix/crossseed-bulkaction`, this PR is 1 commit ahead and changes 11 files: 261 insertions and 178 deletions.

Compared with `origin/develop` (the requested `devlop` branch name does not exist locally or on the checked refs), this head currently includes the child branch work as well: 49 commits ahead, 54 files changed, 6388 insertions and 1047 deletions. This PR will need to be refactored/rebased after the child branch has merged so the final diff only contains the test DB helper work.

## Results

Before:

- `go test -race -count=1 ./internal/models`: 167.439s
- `go test -race -count=1 ./internal/services/crossseed`: 106.234s

After:

- `go test -race -count=1 ./internal/models`: 10.304s
- `go test -race -count=1 ./internal/services/crossseed`: 14.071s

Benchmark evidence:

```text
BenchmarkFullMigrationTestDB      ~117-125 ms/op
BenchmarkClonedMigratedTestDB     ~13.2-14.0 ms/op
```

## Validation

- `go test -race -count=1 ./internal/models ./internal/services/crossseed`
- `go test -race -count=1 ./internal/testutil/testdb`
- `go test -run '^$' -bench 'Benchmark.*TestDB' -benchmem -count=5 ./internal/testutil/testdb`
- `make lint` passed for changed Go files; frontend lint still reports existing warnings and exits successfully.

## Summary by CodeRabbit

* **Tests**
  * Added a shared migrated-SQLite test fixture plus benchmarks and tests to streamline and speed test DB setup.
  * Updated many test suites to use the new fixture and adjusted several test mocks for more consistent behavior.

* **Chores**
  * Standardized test database initialization and cleanup across the codebase for improved isolation and reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/autobrr/qui/pull/1938?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


nitrobass24 (commented):
Two things I'd suggest before merge:

### 1. `removeSQLiteSidecars` will fail on missing files

After `PRAGMA wal_checkpoint(TRUNCATE)` followed by `Close()`, the `-wal` and `-shm` files may already be gone (or never created if no writes followed the checkpoint). `os.Remove` returns `*os.PathError` with `ErrNotExist` in that case, which `errors.Join` propagates as a real failure — `buildMigratedTemplate` would then refuse to build the template on a clean run.

Suggest:

```go
if err := os.Remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
    errs = append(errs, fmt.Errorf("remove sqlite sidecar %s: %w", path, err))
}
```

### 2. Worth adding correctness assertions alongside the benchmarks

The current `Benchmark*` tests measure the speedup, which is great. They won't fail CI if isolation breaks though — e.g. a future refactor that accidentally reuses the same temp path, or a race in `sync.Once`. Three small tests would close that gap:

- `TestNewIsFast` — second call wall time under ~100ms (asserts the template is being reused, not rebuilt). Catches accidental `sync.Once` removal.
- `TestNewIsIsolated` — write to two DBs from two `NewMigratedSQLite` calls; assert no cross-visibility. Catches a regression where the template gets opened directly instead of cloned.
- `TestNewParallel` — 8 `t.Parallel()` subtests each calling `NewMigratedSQLite`; assert all succeed under `-race`. Pins thread-safety.

Happy to push these as a follow-up commit on your branch if it helps, or you can include them here — your call.

Audionut:
Local `race` run measured the second NewMigratedSQLite call around 129ms.
Changed to 250ms, which should still catch accidental template rebuilds.

A more deterministic improvement would seem to be, to add a separate cache assertion, for example calling `migratedTemplatePath()` twice and asserting it returns the same path. That would directly pin `sync.Once` reuse without relying on wall-clock timing.

I'm happy for you to make any further changes as needed, cheers.

nitrobass24:
Yeah you're right, path-equality is the cleaner pin. Pushed `TestTemplatePathIsCached` (62119cb0) — calls `migratedTemplatePath()` twice and asserts same path. Left `TestNewIsFast` in too as a broader perf check. Both pass under `-race`.

All good from my side.

nitrobass24:
@coderabbit review

PR #1948, opened 2026-05-26
fix(proxy): forward API key auth to qBittorrent on passthrough requests

nitrobass24:
## Summary

The reverse-proxy passthrough that backs the per-client `/proxy/{api-key}/...` endpoint (used by Sonarr, Radarr, and any other downstream qBit client) was never taught about API-key auth introduced in #1854. As a result, after switching an instance to API-key auth in v1.19.0:

- in-qui features kept working (qui's own qBit client correctly stamps `Authorization: Bearer <key>` via go-qbittorrent), but
- proxied *arr requests reached qBittorrent with no credentials and failed with 403.

`applyAuthHeaders` only attached the SID cookie from qui's cookie jar (empty under API-key auth, since there's no login flow) and basic auth (typically unused), then explicitly stripped `Authorization` when basic auth wasn't set.

## Fix

- Decrypt the instance API key in `prepareProxyContext` and stash it on `proxyContext`.
- In `applyAuthHeaders`, set `Authorization: Bearer <key>` after the basic-auth branch when an API key is configured. Order matches go-qbittorrent's own client: Bearer wins over basic, mirroring upstream behavior.

The `RetryTransport` 403-reauth path needs no change — `LoginCtx` is a no-op for API-key clients in go-qbittorrent, so a 403 retry harmlessly re-applies the same Bearer header.

## Behavior matrix (proxied *arr → qui → qBit)

| Instance auth mode    | Forwarded `Cookie` | Forwarded `Authorization` |
| --------------------- | ------------------ | ------------------------- |
| username/password     | `SID=...`          | stripped                  |
| API key               | none               | `Bearer <key>` ← was bug  |
| username/password + basic | `SID=...`      | `Basic ...`               |
| API key + basic       | none               | `Bearer <key>`            |

## Tests

Adds `TestProxyContextApplyAuthHeaders_APIKey` covering: Bearer is set when API key is present, Bearer overrides any incoming `Authorization`, Bearer overrides configured basic auth, and the baseline strip-when-nothing-set path.

## Summary by CodeRabbit

* **New Features**
  * Proxied requests now support API key-based authentication: when an instance API key is available it is applied as a Bearer token on outgoing requests; absent credentials are removed to avoid leaking auth.

* **Tests**
  * Added tests to validate API key authentication behavior, header override, and correct handling when no credentials are present.

[Review Change Stack](https://app.coderabbit.ai/change-stack/autobrr/qui/pull/1948?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


martylukyy (commented):
While this fixes API key passthrough for setups without basic auth it still won't work for setup with basic auth due to this:
https://github.com/qbittorrent/qBittorrent/issues/24184

@zze0s PR'ed a fix for us @ https://github.com/qbittorrent/qBittorrent/pull/24392
which would not clash with a Bearer token from an existing basic auth setup...

Leaving it up to @s0up4200 if he wanna merge it with the current limitations or wait for a basic auth fix

nitrobass24:
Thanks @martylukyy — yeah, the basic-auth + API-key combo is the one case this doesn't help, since qBit's Bearer check and the reverse proxy's `Authorization: Basic` collide on the same header. That's upstream #24184, not something I can paper over from qui's side.

Worth noting this PR doesn't regress that combo — it was already broken before, with or without qui in the path. What it does fix is the API-key-only setup, which is what the users have been reporting in Discord since v1.19.0. Happy to follow up with a second PR switching to `X-API-Key` once zze0s's #24392 ships in a qBit release. Whatever @s0up4200 prefers.

PR #1967, opened 2026-06-01
(fix) Fix case mismatch on orphan scan

drtaru:
Summary of the fix:

  Root cause: buildFileMapFromTorrents builds file map paths from filepath.Join(savePath, 
  f.Name) where f.Name comes from qBittorrent's files API using torrent metadata casing (e.g.
   Tracker-ONE/movie.mkv). But qBittorrent's content_path reflects the actual on-disk path
  which may have different casing (e.g. tracker-one). The filesystem walker finds the actual
  disk name → case-sensitive string comparison fails → false orphan.

  Fix: Added correctFileNamesCase(contentPath, files) in service.go that:
  - Extracts the metadata root folder from file names (via torrentRootFolder)
  - Gets the actual on-disk folder name from filepath.Base(contentPath)
  - If they're case-insensitively equal but case-sensitively different, replaces the prefix
  in all file names with the actual on-disk case
  
  This runs in buildFileMapFromTorrents before building file map entries AND before passing
  to actualSavePathFromContentPath (fixing scan root derivation too when TrimSuffix was
  failing due to the case mismatch).

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduce false orphan detections by aligning scan roots, saved paths, and file listings with the actual on-disk folder name casing (non-Windows platforms); path resolution is memoized to improve scan consistency.

* **Tests**
  * Added tests covering filename-casing reconciliation, scan-root normalization, and file-map construction using on-disk casing (includes platform-specific cases for Unix-like systems).

### DIFF
--- a/internal/services/orphanscan/filemap.go
+++ b/internal/services/orphanscan/filemap.go
@@ -130,11 +130,12 @@ func canonicalizeHash(hash string) string {
 	return strings.ToLower(strings.TrimSpace(hash))
 }
 
-// resolvePathCase returns path with its final component corrected to the actual
-// on-disk casing. Uses cache to avoid repeated ReadDir calls on the same parent.
-// If the path already exists as-is, or no case-insensitive match is found in the
-// parent, the original cleaned path is returned.
-// No-op on Windows: normalizePath already case-folds all paths there.
+// resolvePathCase returns path with every component corrected to the actual
+// on-disk casing. Recurses into the parent when the parent itself does not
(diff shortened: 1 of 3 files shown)


drtaru:
This seems to have bigger issues, works fine on my local but fails on my NAS in Docker, will come back to it this week.

s0up4200:
Superseded by https://github.com/autobrr/qui/pull/2316

Issue #1963, opened 2026-05-31
reactor: decompose TorrentTableOptimized.tsx into testable seams


nitrobass24:
# Plan: Decompose `TorrentTableOptimized.tsx` into testable seams

## Status & locked decisions (revised 2026-06-01)

This revision reconciles the plan against the current `develop` and the live file (re-validated by a multi-agent recon of the 3656-line component). **Prerequisites that were "risks" in the original draft are now all resolved/merged:**

- **#1927 / #1929 (test infra + Priority-1 util coverage): CLOSED & merged.** vitest + jsdom + `@testing-library/react` + `renderHook` + coverage-v8 are on `develop`. No infra branch to push — the framework is live. Reference templates exist: `web/src/hooks/useInstanceMetadata.test.tsx` (QueryClient wrapper), `web/src/test/mockTorrent.ts` (fixtures), `SyncStreamContext.test.tsx` (MockEventSource).
- **#1930 (lint/type tightening, incl. the `types/index.ts` split via #1968): CLOSED & merged.** `web/src/types/index.ts` is now a 28-line barrel over ~21 domain modules on `develop`. The table still imports from the `@/types` barrel, so the split is transparent. **The "don't run #1930 during decomp" sequencing risk no longer applies — decomp can start now.** Extracted hooks should keep importing from the `@/types` barrel to minimize diff churn.
- **#1925: CLOSED.** The fix lives in `selectAllFilters` (see §"#1925 regression pin", PR-05).

**Decisions locked with the maintainer (2026-06-01):**

1. **Start decomp now.** No #1930 sequencing wait — all prerequisites merged.
2. **Drop provably-dead code during extraction and call it out in the PR description.** Concretely, PR-03 drops `immediateSearch` (an always-`""` `useState` with no setter). NOTE: `navigate` (the `useNavigate()` result) is **not** dead — it is used in the render at line ~3335 and stays in the parent. This is the only sanctioned scope-expansion beyond pure relocation.
3. **Add one whole-table RTL smoke test in PR-01** as a regression tripwire that runs across all 10 PRs, on top of the per-seam `renderHook` tests.
4. **Keep PR-04 (`useTorrentSelection`) and PR-05 (`useTorrentSelectionDerivations`) split** so the #1925 regression test isn't buried; ship the 3-hook PR-06 and 3-hook PR-10 bundles as-is.
[...]

PR #1979, opened 2026-06-02
fix(web): emit allowSubcategories from the filtered-data callback

nitrobass24:
> ## ⚠️ Merge order — depends on #1978
>
> **#1978 (PR-08) must be merged first.** This PR's desktop half edits the `useTorrentTableNotifications` hook introduced in #1978, so it's based on that branch. It is **not** part of the #1963 decomposition chain — it's a standalone behavior fix. After #1978 merges, this PR auto-retargets to `develop`.

---

## Summary

Aligns the `useSubcategories` value emitted via `onFilteredDataUpdate` with the value the table/cards actually render from.

Both the desktop table (now via `useTorrentTableNotifications`) and the mobile cards (`TorrentCardsMobile`) forwarded the **raw** backend flags (`subcategoriesFromData` / `supportsSubcategories`) to the parent's `onFilteredDataUpdate` callback, while they render subcategories from the **fully-derived** `allowSubcategories` — which additionally folds in the `use_subcategories` preference and the `subcategoriesAlwaysEnabled` flag.

This PR emits `allowSubcategories` from **both** paths (so desktop and mobile stay consistent), and adds it to the effect dependencies so a pure preference toggle now re-fires the callback (previously a missing-dep gap — a preference change didn't re-emit).

### Why this is low-risk (alignment, not a user-visible bug)
The only concrete-instance consumer of the emitted value is `Torrents.tsx`'s `useSubcategories` state → `FilterSidebar`, which **re-derives** subcategory-enablement from its own capabilities + preferences and uses the incoming prop only as a terminal `??` fallback. So the parent already self-corrects the divergence; this change mainly makes the emitted contract honest and fixes the preference-toggle re-fire.

### Scope note
This is **not** part of the #1963 table decomposition — it's a small deliberate behavior fix surfaced by review on #1978. It's based on `feat/1963-pr08-notifications-columns` only because the desktop half edits the `useTorrentTableNotifications` hook introduced there; the mobile half is independent. **Merge after #1978.**

### Changes
- `useTorrentTableNotifications.ts` — param `subcategoriesFromData` + `supportsSubcategories` → `allowSubcategories`; effect emits it directly (drops the now-dead `?? previousUseSubcategories` fallback and the `supportsSubcategories` cache field); dep array updated.
- `TorrentTableOptimized.tsx` — pass `allowSubcategories` to the hook.
- `TorrentCardsMobile.tsx` — emit `allowSubcategories` + dep array.

## Testing
- [x] `pnpm test` — 490/490 (2 new: "emits allowSubcategories verbatim", "re-fires when only allowSubcategories changes")
- [x] `tsc -b` clean; `pnpm lint` — changed files clean; `pnpm build` succeeds

Resolves the CodeRabbit thread on #1978 ([discussion](https://github.com/autobrr/qui/pull/1978#discussion_r3337788635)).

## Summary by CodeRabbit

* **Refactor**
  * Simplified internal handling of the subcategory preference to make notifications and table updates more consistent.

* **Bug Fixes**
  * Notifications and filtered torrent views now reliably refresh when the subcategory preference is toggled so the UI reflects changes immediately.

* **Tests**
  * Added coverage verifying the preference is passed through and that updates re-trigger when it changes.


nitrobass24:
Rebased onto `develop` to clear the merge conflict.

To resolve it I rebased the branch onto current `develop` and dropped everything that's already merged, keeping only the actual fix:

- Dropped the 11 already-merged stack commits.
- Dropped the `no-selectAllFilters-leak` test commit — `git` reported its patch contents are already upstream (it came in with the squashed notifications extraction).
- Replayed only `fix(web): emit allowSubcategories from the filtered-data callback`.

The PR is now a single commit on top of `develop`, touching just the 4 intended files:

- `web/src/components/torrents/TorrentTableOptimized.tsx`
- `web/src/components/torrents/TorrentCardsMobile.tsx`
- `web/src/hooks/torrent-table/useTorrentTableNotifications.ts`
- `web/src/hooks/torrent-table/__tests__/useTorrentTableNotifications.test.ts`

Verified locally after the rebase: `tsc -b` passes and `useTorrentTableNotifications.test.ts` is green (11/11). Force-pushed, so a fresh pull/checkout of this branch is needed.

PR #1992, opened 2026-06-03
fix(sse): stop intermittent "offline" flips that fall back to polling

### LINKED ISSUE #1991: SSE stream intermittently drops to "offline" and falls back to REST polling
## Summary

The torrent table, dashboard, and title-bar speeds are kept live by a single Server-Sent Events (SSE) connection. When that connection looks unhealthy, the UI shows an "offline" status and the affected views automatically resume REST polling until the stream recovers.

The stream **intermittently** reports offline and reverts to polling for a few seconds even when qBittorrent and the network are healthy. This issue documents the root causes and a plan to fix them.

There is no data loss when it happens (polling takes over), but it's visible churn, it defeats the point of the streaming feature, and it adds avoidable backend load during the fallback window.

## Background: how the connection and fallback work today

- The frontend opens one `EventSource` to `/api/stream` and multiplexes all views over it (`web/src/contexts/SyncStreamContext.tsx`).
- The backend sends a **heartbeat every 5s**. If the frontend sees **no event of any kind for 15s** (`STREAM_STALE_TIMEOUT_MS`), a watchdog declares the stream dead and forces a reconnect.
- On any disconnect/error, the frontend tears the connection down and reconnects with exponential backoff (**minimum 4s**, `RETRY_BASE_DELAY_MS`).
- While the stream is not `connected`, consumers (`useTorrentsList`, `useTitleBarSpeeds`) re-enable REST polling.
- On the backend, all events are written to clients by a **single shared dispatcher** (the go-sse "Joe" provider), one client at a time.

## The problems

**Problem 1 — A momentary network blip turns into a guaranteed multi-second outage.**
Browsers handle brief SSE drops automatically: when the connection hiccups (a Wi-Fi roam, a proxy/CDN re-handshake — e.g. Cloudflare Access — an HTTP/2 GOAWAY, a load-balancer drain), the browser keeps the same `EventSource` and silently reconnects in ~1-3s with no visible impact. But our error handler doesn't recognize this "still reconnecting" state. It immediately *closes* the connection — which **cancels the browser's own fast reconnect** — and then starts our own reconnect on a **minimum 4-second** delay. So a blip the browser would have hidden in ~3s becomes a guaranteed ≥4s "offline + polling" window. Because these blips are inherently sporadic, this is the most likely source of the intermittent reports.

**Problem 2 — One slow client can stall heartbeats to everyone else.**
All clients are served by a single dispatcher that writes to them one at a time, and each write is allowed to block for up to **10 seconds** before timing out. If one client stops reading (laptop asleep, flaky link, a buffering proxy), it blocks the shared dispatcher, delaying heartbeats to *healthy* clients. The 10s write timeout sits only 5s under the 15s "stream is dead" threshold, so when a couple of clients stall around the same time — which correlated events like sleep/wake or a proxy buffering multiple connections make realistic — healthy clients can cross 15s of silence and falsely flip to offline. The heartbeat loops also fire without jitter, so multiple instances' heartbeats arrive at the dispatcher as a synchronized burst, deepening the serial backlog this problem depends on.

**Problem 3 — Returning to a backgrounded tab can falsely report offline.**
The 15s watchdog is a timer. Browsers heavily throttle timers (and event delivery) in background tabs. When you switch back to the qui tab, an overdue throttled timer can fire *before* the next heartbeat is processed, so a perfectly healthy connection is declared dead and reconnects. Symptom: "I switch back to the tab and it briefly says offline."

**Problem 4 — Changing a sort/filter/search rebuilds the whole connection, and a slow first refresh shows offline.**
The stream connection is keyed on the full view parameters, so any sort/filter/search change tears down the `EventSource` and opens a new one. The offline state during that swap is hidden only for a fixed **1.2 seconds**. If the new connection's first data snapshot takes longer than 1.2s — which happens with an uncached filter/sort combination (tracker-based filters), a cold cache just after a backend restart, or a network blip during the swap — the UI flips to offline + polling even though nothing is actually wrong.

## Implementation plan

Ships as a **single PR with one commit per fix**, so each change is independently reviewable and revertible. Problems 1 and 2 are the primary causes; 3 and 4 are narrower but survive the first two fixes, so they're included. Commit 3 is a small backend hygiene change that rides along with the Problem 2 work.

### Commit 1 — `fix(sse): don't cancel native EventSource reconnect on transient errors`

**File:** `web/src/contexts/SyncStreamContext.tsx` (`handleNetworkError` ~428-441; `source.onerror` ~589)

- In the `onerror` handler, read `EventSource.readyState` before doing anything.
- If `readyState === CONNECTING` (the browser is already auto-retrying a transient drop): **do nothing** — don't close the source, don't set `connected=false`, don't start the 4s backoff. Let native retry proceed. The existing 15s stale watchdog remains the escalation path.
- If `readyState === CLOSED` (terminal: bad status, content-type mismatch, explicit close): run the current path (close + `scheduleReconnect`).
- Keep the **stale-timer callback (~460) calling the forced-teardown path directly** — 15s of true silence must still escalate even if the source is stuck in `CONNECTING`. Without it, a connection the browser retries forever (a long outage) would never fall back to polling.

```ts
const handleSourceError = () => {
  const liveSource = connectionRef.current.source
  // Transient drop: the browser is auto-reconnecting (readyState CONNECTING).
  // Closing here aborts that and forces our >=4s backoff. Let it recover; the
  // 15s stale watchdog escalates if it doesn't.
  if (liveSource && liveSource.readyState === EventSource.CONNECTING) return
  handleNetworkError() // CLOSED/terminal -> existing teardown + scheduleReconnect
}
source.onerror = handleSourceError
```

**Tests:** `SyncStreamContext.test.tsx` — `MockEventSource.emitError` currently only models the CLOSED path. Add a CONNECTING variant and assert it does **not** close the source, **not** flip `connected=false`, and **not** schedule a reconnect; assert that a subsequent CLOSED error *or* the 15s watchdog *does* escalate.

### Commit 2 — `fix(sse): isolate slow clients with a per-session buffered writer`

**File:** `internal/api/sse/manager.go` (writer wrapping in `Serve` ~657; `deadlineResponseWriter` ~667-684)

Give each session a **bounded, non-blocking buffered writer** so a slow socket can never block the shared dispatch loop:

- Wrap the response writer so `Write` **enqueues** encoded bytes into a per-session bounded buffer (copying the slice, since go-sse may reuse it) and returns immediately. A dedicated per-session goroutine drains the buffer to the real socket, applying the rolling `streamWriteTimeout` deadline there.
- If the buffer is full (client too slow to keep up) **or** a drained write errors/times out: drop *that* client only — return an error from `Write` (go-sse's dispatcher removes a subscriber whose `Send`/`Flush` returns an error, `joe.go:249-258`) and/or cancel the request context so go-sse ends the session. Either way the shared loop is never blocked.
- Preserve `Flush()` (drain/flush the socket) and `Unwrap()` so go-sse's writer-capability detection keeps working.
- Spawn the drain goroutine in `Serve` and tear it down when the handler returns; thread a cancelable context (`ctx, cancel := context.WithCancel(r.Context())`) so a drain error can end the session cleanly.

Once writes can't head-of-line-block, the 10s-write-vs-15s-stale arithmetic stops mattering, so no constant changes are required — though `streamWriteTimeout` can optionally be lowered since it now only bounds a single client's own drain.

**Tests:** add a delivery test where one subscriber's writer blocks indefinitely and assert a second (healthy) subscriber still receives heartbeats/updates on time; add a test that an overflowing subscriber is dropped without affecting others. Reuse the harness in `manager_delivery_test.go`. Run `go test -race -count=1 ./internal/api/sse/...`.

### Commit 3 — `fix(sse): jitter heartbeat loops to avoid synchronized bursts`

**File:** `internal/api/sse/manager.go` (`startHeartbeatLoop` ~1565-1584; `activityHeartbeatLoop` ~1221-1233; existing `jitteredInterval` ~1531-1539)

- Both heartbeat loops use a fixed `time.NewTicker(heartbeatInterval)` with no jitter, unlike the sync loop which deliberately jitters so loops don't fire in lockstep. With several instances, their heartbeat ticks align and reach the shared dispatcher as a burst every 5s.
- Replace the fixed ticker with a `time.Timer` reset to `jitteredInterval(heartbeatInterval)` each iteration (mirroring `startSyncLoop`), reusing the existing `jitteredInterval` helper. This spreads heartbeat writes across the interval and shortens the serial backlog the dispatcher processes per tick — complementing Commit 2.
- *(Optional)* `publishHeartbeat` issues one `Publish` per subscriber for an instance; batching those into a single multi-topic publish (as `publishToActivityTopics` already does) cuts serial round-trips further.

**Tests:** light — assert the heartbeat loops still emit on roughly the expected cadence (jitter is timing, not behavior, so no strict assertion). Covered by the existing delivery tests in `manager_*_test.go`.

### Commit 4 — `fix(sse): pause stale watchdog while the tab is hidden`

**File:** `web/src/contexts/SyncStreamContext.tsx` (visibility handler ~829-861; stale timer ~446-462)

- On `visibilitychange → hidden`: clear the stale timer (its timing is meaningless under background throttling).
- On `visibilitychange → visible`: if the source is OPEN, **re-arm a fresh stale timer** (`resetStaleTimer()`) so an overdue throttled timer can't immediately fire; keep the existing "if CLOSED, force reconnect" behavior.
- `resetStaleTimer` is currently a local closure inside `openConnection`; expose it via a ref (mirroring the existing `scheduleReconnectRef` pattern) so the visibility effect can call it.

**Tests:** simulate hidden→visible with an OPEN source and an overdue timer; assert no teardown/reconnect occurs.

### Commit 5 — `fix(sse): keep stream connected during view-parameter swaps`

**File:** `web/src/contexts/SyncStreamContext.tsx` (handoff logic ~393-416; `HANDOFF_GRACE_PERIOD_MS` ~15; `onopen` ~558-588)

- When the new connection's `onopen` fires, **refresh/extend the handoff timer** instead of letting a fixed 1.2s window expire — a connection that has demonstrably opened but is waiting on a slow first snapshot should not be declared offline.
- Only flip `connected=false` in the handoff timer if the socket has **not** opened *and* no init has arrived.
- Optionally raise/adaptively size `HANDOFF_GRACE_PERIOD_MS` for safety.
- *(Optional backend follow-up, `manager.go` `onSession`)*: flush response headers / `200` **before** building the initial snapshot (`buildGroupPayload`), so `onopen` fires promptly and snapshot-build latency doesn't also delay the open.

**Tests:** rapid sort/filter/search changes never produce `connected=false` when the first init arrives within normal latency; inject an init delay longer than the base grace and assert the refreshed-on-open logic keeps it connected.

## Acceptance criteria

- A simulated transient drop (EventSource `onerror` with `readyState === CONNECTING`) does **not** flip the UI offline or re-enable polling; native reconnect recovers silently.
- A single stalled/slow client does **not** delay heartbeat/update delivery to other connected clients (verified by race-enabled backend test).
- Multiple instances' heartbeats are spread across the interval (jittered) rather than firing simultaneously, reducing per-tick dispatch backlog.
- Returning to a long-backgrounded tab does **not** momentarily show offline on a healthy connection.
- Changing sort/filter/search does **not** show offline when the new connection opens and its first snapshot arrives within normal latency.
- Genuine failures still fall back correctly: a truly dead stream still escalates via the 15s watchdog, and a terminal `CLOSED` error still reconnects with backoff.

nitrobass24:
## Summary

The SSE stream that keeps the torrent table, dashboard, and title-bar speeds live would intermittently report **"offline"** and fall back to REST polling for a few seconds even when qBittorrent and the network were healthy. This PR fixes the root causes (one backend, one frontend) and three narrower contributors. No data was ever lost (polling takes over), but the churn defeats the point of streaming and adds avoidable backend load during each fallback window.

Closes #1991.

## What was wrong & what changed

**1. A momentary network blip became a guaranteed multi-second outage.** The browser silently auto-reconnects transient SSE drops (`readyState === CONNECTING`, ~1–3s, invisible), but our `onerror` handler unconditionally closed the connection — aborting that native retry — and started its own ≥4s backoff. Now `onerror` defers to the browser's native retry while connecting and only tears down on a terminal `CLOSED` (the 15s stale watchdog still escalates a connection stuck connecting).

**2. One slow client could stall heartbeats to everyone else.** All clients share go-sse's single serial dispatch loop, and each write could block it for up to 10s — close to the frontend's 15s "stream dead" threshold. Each session now has a **bounded, non-blocking buffered writer** with its own drain goroutine; a client that can't keep up is dropped (its bounded queue overflows / its write deadline fires) without ever head-of-line-blocking healthy sessions.

**3. Synchronized heartbeat bursts.** The per-instance and activity heartbeat loops fired without jitter (unlike the sync loop), bunching through the single dispatcher. They now jitter like the sync loop.

**4. Returning to a backgrounded tab could falsely report offline.** Background-tab timer throttling could leave the 15s stale watchdog overdue, firing on refocus and killing a healthy connection. The watchdog is now cleared while hidden and re-armed fresh on refocus.

**5. Changing a sort/filter/search briefly showed offline.** A view-parameter change rebuilds the EventSource, and the swap was masked only by a fixed 1.2s grace; a slow first snapshot flipped it offline. The replacement connection now keeps the stream connected once its socket has opened, letting the snapshot (or the 15s watchdog) resolve it.

## Commits (one per fix)

- `fix(sse): isolate slow clients with a per-session buffered writer`
- `fix(sse): jitter heartbeat loops to avoid synchronized bursts`
- `fix(sse): don't cancel native EventSource reconnect on transient errors`
- `fix(sse): pause stale watchdog while the tab is hidden`
- `fix(sse): keep stream connected during view-parameter swaps`

## Automated testing

- [x] `go build ./...`
- [x] `go test -race -count=1 ./internal/api/sse/...` — green, incl. new isolation tests proving a stalled client doesn't block others, an overflowing client is dropped, and `Flush` after `Close` never panics
- [x] `tsc -b` clean
- [x] `vitest` SSE suite — 24/24 (new tests for the transient-error guard, the visibility re-arm, and the view-swap handoff)
- [x] `make lint` — 0 errors

## Manual / UA testing (to complete before merge)

These behaviors are timing/network/browser-dependent and can't be fully covered by unit tests. Watch the status-bar connection indicator and whether REST polling re-enables (visible in DevTools → Network).

**Regression — normal streaming still works (ideally through the real reverse proxy):**
- [x] Torrents table, dashboard, and title-bar speeds update live with the status bar showing **connected** (not reconnecting/fallback).
- [x] DevTools → Network shows one long-lived `/api/stream` connection open, and the REST list endpoint is **not** firing every ~1–3s while connected.
- [x] Events appear promptly — no visible batching/lag introduced by the per-session buffered writer.

**Slow-client isolation (the core C2 property):**
- [x] With two tabs/devices on the same view, pausing/throttling/sleeping one does **not** stall updates in the other; the healthy tab stays connected. Resuming the slow one recovers or drops cleanly without wedging the other.

**Per-fix scenarios:**
- [x] **Transient blip (C1):** a brief network drop / `docker restart` recovers **without** a visible multi-second offline→polling flash; a longer outage still falls back to polling and recovers when restored.
- [x] **Backgrounded tab (C4):** backgrounding the tab for >20–30s and returning does **not** flash "offline" on refocus.
- [x] **Param swap (C5):** rapidly changing sort / filters / search shows no "offline" flash during the swaps.
- [x] **Idle heartbeats (C3):** leaving the view idle (no torrent activity) for ~1 min keeps the stream connected.

## Notes

- No UI/visual changes (connection-resilience logic only) and no schema or API-contract changes.
- Known bounded trade-off: on a stalled-client disconnect, `Serve`'s deferred writer `Close()` can wait up to the 10s write deadline for the drain goroutine to unwind. Acceptable for a single-user self-hosted app, and only on that disconnect path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Isolated slow/stalled SSE clients so one connection no longer blocks others; stuck clients are dropped cleanly and do not cause resource leaks.
  * Connection error handling improved so transient/connectivity glitches don’t trigger unnecessary reconnects; tab visibility restores no longer kill healthy streams.

* **Performance Improvements**
  * Heartbeat scheduling now uses jitter to avoid synchronized bursts; per-connection write buffering prevents global stalls.

* **Tests**
  * New deterministic tests cover buffering, overflow/drop behavior, non-blocking flush, isolation between clients, and teardown race safety.


nitrobass24:
So the stream still goes offline, when the tab is in the background for 30s or so.

nitrobass24:
Looks good now. #1994 will need a rebase to catch the new commits to validate the tests after this merges.

PR #2017, opened 2026-06-09
fix(crossseed): drop resolution token from TV search queries

nitrobass24:
## Summary

TV cross-seed searches were silently failing on indexers whose free-text search matches only the **series name** (notably **BTN** and **IPT**): the search returned 0 results, so cross-seeds to those trackers were never created — while the same release cross-seeded fine elsewhere.

Reported on Discord by multiple users. The tell-tale: a manual Prowlarr search for `tvshow S01E01` returns results on BTN, but `tvshow S01E01 1080` returns **zero**.

## Root cause

For TV content, qui appended a bare resolution token to the free-text Torznab query, producing `q = "<title> S01E01 1080"`. Indexers that match `q` against the series name find nothing for the `1080` token and return zero results. (Sonarr works because it issues a structured `tvsearch` with IDs + `season`/`ep` and no resolution in `q`; RSS automation works because it matches the full feed title with no `q` search at all.)

The resolution was injected at **two** sites, both added in #1871 ("refine TV based searching", which also raised the result limit to 100 + `ReturnAllResults`):

1. The cross-seed query builder — `IncludeResolution` for TV (`internal/services/crossseed/search_query.go`).
2. The jackett Prowlarr **ID-driven** path — `idDrivenTVQuery` re-extracted the resolution from the release name even when the query itself was clean (`internal/services/jackett/service.go`).

## Fix

Drop the resolution token from the TV `q` at **both** sites:

- Remove the `SearchQueryOptions{IncludeResolution}` apparatus from the cross-seed query builder.
- ID-driven TV searches now use the bare `S01E01` token; removed the now-dead `idDrivenTVQuery` / `firstResolutionToken` / `searchResolutionToken` helpers.

**Kept unchanged:** the `S01E01` season/episode narrowing, the generic `appendSearchToken` season-token repositioning, and the 100-result / `ReturnAllResults` behavior.

Resulting query: `"<title> S01E01"` (non-ID) / `"S01E01"` (ID-driven) — the form confirmed to return results on BTN.

## Why this is safe

Resolution in `q` was only ever a result-*narrowing* heuristic, never required for correctness: `internal/services/crossseed/matching.go` still strictly rejects resolution mismatches **after** the search (`1080p` ≠ `2160p` are different files). Dropping the token changes which candidates an indexer returns, not which ones we accept.

## Testing

- `make test` — full `-race -count=1 ./...`: pass (0 failures / 0 races).
- Targeted `internal/services/crossseed` + `internal/services/jackett` pass; updated the unit tests that asserted the old `… 1080` / `S17 1080p` / `S22 720` queries.
- `make precommit` (lint): 0 errors. `make build`: frontend + Go binary compile.
- No schema/DB changes (Postgres / migration CI jobs unaffected).
- **Pending:** live confirmation against real BTN/IPT by the reporting users.

## Notes

Net −75 lines. One residual tradeoff (what #1871 was guarding against): without resolution narrowing, a very popular single episode could return more candidates and, worst case, risk truncation at the 100-result cap — small for episode-level searches. If it ever bites, a 0-result fallback retry (search with resolution, retry the indexer without it) would recover the narrowing without breaking series-name-only indexers; deliberately left out here to keep the change minimal.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed automatic resolution tokens from generated search queries (including ID-driven TV searches), producing cleaner, more consistent searches and avoiding duplicated resolution tags.
* **Tests**
  * Updated and extended tests to reflect the new query behavior and added a fallback to ensure queries are never empty.


Audionut:
Bah, I always have that as a source, so never noticed that resolution was a blocker.

We're trading better matching at poorly curated sites, and probably the occasional super long running shows, against working capacity for the listed indexers, a no brainier.

Ran the bot over this, and it's only concern, funnily enough, was the risk of missing content because of the underlying constraints of jackett/prowlarr.

LGTM

PR #2037, opened 2026-06-16
fix(sse): size full-sync timeout to prevent permanent disconnect on large instances

nitrobass24:
## Summary

Fixes a regression in v1.20.0 where qui permanently loses its connection to qBittorrent on large instances (~18k–20k+ torrents) and stays disconnected until the container is restarted. The UI shows a permanent loading skeleton and reverse-proxied *arr apps get no response.

## Root cause

PR #551 moved qBittorrent polling server-side. A per-instance loop (`startSyncLoop` → `forceSync`) drives `/sync/maindata`, and `forceSync` wrapped **every** sync in a fixed **10-second** timeout:

```go
ctx, cancel := context.WithTimeout(parent, 10*time.Second)
```

qBittorrent's `/sync/maindata` is incremental via a `rid`. The wedge:

1. A **full update** for 20k torrents takes longer than 10s → `forceSync`'s context cancels mid-response.
2. The library's `MainData.Update` returns the error **without advancing its local `rid`** (`dest.Rid = source.Rid` is never reached) — so qui keeps the stale rid.
3. But qBittorrent **has** advanced its internal rid, so the next request (stale rid) forces **another full update** → >10s again → cancelled → stale rid → repeats forever. Exponential backoff only slows retries to a 30s cap; it never escapes.

Small instances self-heal instantly because their full update finishes well under 10s — which is why this only bites large instances.

This is a **regression**: in v1.19 the `internal/api/sse/` path didn't exist; sync ran via `Sync(context.Background())`, bounded only by the **60s** qbit HTTP client timeout. #551 silently dropped the effective per-sync budget from ~60s to 10s.

## Fix

A **two-tier timeout** in `forceSync`:

- `syncTimeoutIncremental = 10s` — steady-state delta ticks (unchanged behavior for the common path; keeps the ~2s loop snappy).
- `syncTimeoutFull = 60s` — full/initial sync, aligned with the qbit HTTP client timeout so the budget is bounded by the transport.

A full update is expected when the instance has **never completed a sync** *or* is in a **failure streak** (`attempt > 0`):

```go
func (m *StreamManager) syncTimeout(instanceID int) time.Duration {
    m.mu.RLock()
    defer m.mu.RUnlock()
    state, ok := m.syncBackoff[instanceID]
    if !ok || !state.primed || state.attempt > 0 {
        return syncTimeoutFull
    }
    return syncTimeoutIncremental
}
```

A new `backoffState.primed` flag is set on first success and **never cleared by failure** (failure streaks are tracked by `attempt`). After a transient incremental timeout strands the rid, the next tick — which faces qBittorrent's forced full update — gets the full 60s, completes, advances the rid, and the instance returns to the 10s budget. That's the wedge escape.

`forceSync`'s context still derives from the loop `parent`, so shutdown/loop-restart still cancels an in-flight full sync promptly. The `syncMgr.Sync` error branch is unchanged (still relies on the library `OnError` → `HandleSyncError` path, no double-reporting).

## Testing

- `go build ./...` — clean
- `go test -race -count=1 ./internal/api/sse/... ./internal/qbittorrent/...` — pass
- `gofmt -l` clean; `golangci-lint --new-from-merge-base=develop` — 0 issues
- New tests: table-driven `TestStreamManager_syncTimeout` (all state combinations), `TestStreamManager_failureAfterPrimeKeepsFull_thenRecovers` (wedge-escape proof), `TestStreamManager_unprimedFirstSyncUsesFull` (exact wedge condition), and `TestStreamManager_syncTimeout_concurrent` (`-race` guard on the shared backoff state)

## Reports this addresses

Multiple users on Discord reported v1.20.0 disconnecting from qBittorrent "after a while" and staying disconnected until container restart, all correlated with large instances (18k–20k+ torrents). Reverting to v1.19.0 resolved it for some. No prior GitHub issue was filed.

## Summary by CodeRabbit

* **Improvements**
  * Updated SSE sync-loop timeout selection to better reflect instance health, using shorter budgets for incremental updates after initial success and longer budgets for full synchronization during failure streaks.
  * Updated forced sync to use the same adaptive timeout logic.

* **Tests**
  * Revised and expanded coverage for priming behavior, timeout transitions across failure streak depth, and race-safe concurrent state access.

### DIFF
--- a/internal/api/sse/manager.go
+++ b/internal/api/sse/manager.go
@@ -37,7 +37,18 @@ const (
 	streamEventActivity  = "activity"
 	defaultSyncInterval  = 2 * time.Second
 	maxSyncInterval      = 30 * time.Second
-	heartbeatInterval    = 5 * time.Second
+	// syncTimeoutIncremental bounds a steady-state incremental /sync/maindata tick.
+	// Healthy delta updates finish well under this even on large instances, so the
+	// short budget keeps the ~2s loop snappy and frees the goroutine when qbit is
+	// briefly slow.
+	syncTimeoutIncremental = 10 * time.Second
+	// syncTimeoutFull bounds a full /sync/maindata update (first sync for an instance,
+	// or any sync after a failure when qBittorrent has reset the rid and resends the
+	// whole torrent set). A 20k-torrent full fetch+parse can exceed 10s; this is
+	// aligned with the qbit HTTP client timeout (60s, see ClientPool.GetClient) so the
+	// per-sync budget is bounded by the transport, not a tighter inner deadline.
+	syncTimeoutFull   = 60 * time.Second
+	heartbeatInterval = 5 * time.Second
 	// maxStreamRequests caps the number of stream subscriptions a single SSE
 	// connection may request, bounding per-connection fan-out and resource use.
 	maxStreamRequests = 64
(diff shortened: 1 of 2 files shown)


s0up4200 (approved):
LGTM @nitrobass24 thanks 👍

PR #2056, opened 2026-06-23
fix(jackett): decouple cache persistence from SkipHistory

### LINKED ISSUE #1997: crossseed: alternate connector search pass disables Torznab cache reuse via SkipHistory
## Context

Follow-up from review of #1990 (cross-tracker relabel + alternate `and`/`&` connector search). Flagged by @Audionut's review bot as a P3.

## Problem

The alternate connector-spelling pass added in #1990 sets `altReq.SkipHistory = true` to avoid creating a parallel search-history job:

`internal/services/crossseed/service.go` (alt pass) → `altReq.SkipHistory = true`

However, in the Jackett service, `SkipHistory` gates **both** history recording and Torznab result-cache persistence:

```go
// internal/services/jackett/service.go:775
if cacheEnabled && cacheSig != nil && len(networkCoverage) > 0 && !req.SkipHistory {
    // ... persist cache
}
```

So every alternate-spelling pass bypasses the reusable Torznab search cache. Repeated cross-seed searches that trigger the alternate pass (any query containing an `and`/`&` connector with indexers that returned nothing for the primary spelling) re-hit indexers each time instead of reusing cached results.

## Suggested direction

Separate the two concerns currently conflated under `SkipHistory`:

- Keep "do not create a parallel history job" behavior for internal continuation passes.
- Allow result-cache persistence to proceed.

Likely a dedicated request flag (e.g. `SkipHistory` for the job/history entry vs. a distinct `SkipCache`) in `internal/services/jackett/models.go`, with the cache-persist gate at `service.go:775` keyed on the cache-specific flag. The alternate pass would then skip history but still populate cache.

This is intentionally scoped out of #1990 to keep that PR's diff focused; it touches the Jackett service request API.

## Acceptance

- Alternate connector-spelling passes still avoid creating duplicate history jobs.
- Torznab cache is populated/reused for alternate-spelling passes.
- No change to primary search behavior.

jussaw:
## Summary

Fixes #1997.

The cross-seed alternate connector-spelling pass (added in #1990) set `altReq.SkipHistory = true` to avoid creating a parallel search-history job. However, in the Jackett service `SkipHistory` gated **two** concerns: history recording *and* Torznab result-cache persistence. As a result every alternate-spelling pass bypassed the reusable Torznab cache, so repeated cross-seed searches containing an `and`/`&` connector re-hit indexers each run instead of reusing cached results.

This PR separates the two concerns:

- Adds a dedicated `SkipCachePersist` flag to `TorznabSearchRequest`.
- Keys the cache-persist gate in `performSearch` on `SkipCachePersist` instead of `SkipHistory` (cache *reads* remain governed by `CacheMode`, unchanged).
- The alternate pass keeps `SkipHistory = true` but leaves `SkipCachePersist` unset, so it skips history while still populating and reusing the cache.
- The indexer connectivity test sets `SkipCachePersist = true` explicitly to preserve its prior no-cache behavior (avoids persisting a bogus "test" entry).

## Changes

- `internal/services/jackett/models.go` — new `SkipCachePersist` field with docs clarifying its independence from `SkipHistory`.
- `internal/services/jackett/service.go` — persist gate now keyed on `SkipCachePersist`.
- `internal/services/crossseed/service.go` — comment clarifying alt pass intentionally leaves cache persistence enabled.
- `internal/api/handlers/jackett.go` — test-indexer request opts out of cache persistence explicitly.
- `internal/services/jackett/service_test.go` — table-driven regression test (`TestSearchCachePersistGate`) covering the persist gate and flag independence.

## Acceptance (from #1997)

- [x] Alternate connector-spelling passes still avoid creating duplicate history jobs.
- [x] Torznab cache is populated/reused for alternate-spelling passes.
- [x] No change to primary search behavior.

## Testing

- `go test -race -count=1 ./internal/services/jackett/ ./internal/services/crossseed/ ./internal/api/handlers/` — all pass.

## Summary by CodeRabbit

* **Improvements**
  * Refined caching behavior so test connectivity searches no longer get stored in the search result cache
  * Decoupled search-history recording from search-cache persistence for more granular control
  * Updated alternate search-pass behavior to reuse cached results when cache persistence is allowed
* **Tests**
  * Added coverage to ensure cache persistence is governed only by the new cache-persist skip setting, independent of history skipping


s0up4200 (approved):
Thanks for the PR!

PR #2086, opened 2026-06-30
feat(fs): respect process umask for content directory creation

s0up4200:
Content directories (hardlink/reflink/cross-seed/dir-scan link trees) were created with hardcoded `0755`/`0750` modes, so the group-write bit was never requested and the process umask could never preserve it. This requests `0o777` for those dirs (and `0o770` for the dir-scan base, which keeps its historical no-`other` posture) and lets the umask decide the final bits. It also adds a `UMASK` env var that qui applies at startup so the official root image can opt in too (default behavior unchanged), since that image runs as root with umask `022`.

So a content-sharing umask cannot expose secrets, the generated config file (which holds the session secret) is now written `0600` and the database directory is created `0o750` like the config and log dirs. Implements discussion #1704.

## Summary by CodeRabbit

* **New Features**
  * Added Unix `UMASK` support so newly created directories/files respect `UMASK` (no-op on Windows; inherited behavior documented).
  * Standardized directory permission modes across hardlink/reflink and link-tree base directory creation.
* **Bug Fixes**
  * Ensure `config.toml` is always owner-only (`0600`) even under permissive `umask`.
  * Tighten SQLite database and sidecar permissions to owner-only (`0600`), and restrict the parent directory mode.
* **Documentation**
  * Expanded troubleshooting with a “Directory permissions and umask” section and examples.
* **Tests**
  * Added Unix and mode-specific coverage for `UMASK`, config/file owner-only permissions, and database security.


connor-grady:
Unit tests pass cleanly on my NixOS system (go 1.26). A few non-blocking nits:

- **Docs**: `MkdirAll` only affects newly created dirs. Worth a troubleshooting line teling upgraders to `chmod` existing trees once.
- **Test**: every `TestApplyUmask` case sets `UMASK`, so the entirely missing case never gets tested.
- **Consistency**: `internal/backups` and `internal/services/trackericons` still hardcode `0o755`. Not load-bearing, just a mixed taxonomy now.

Nice call on the `0600` config/db hardening.

PR #2115, opened 2026-07-04
feat(rss): show feed titles in affected feeds selector

pengchongfu:
In the RSS rule form, the "affected feeds" selector now displays each feed's title as the primary label with the URL shown underneath in smaller muted text. If a feed has no title, the URL is shown instead.

Previously only the raw URL was displayed, which was hard to read for long URLs.

# screenshot

| before | after |
| --- | --- |
| 截屏2026-07-04 23 11 28 | 截屏2026-07-04 23 11 39
 |

# note

- When the parent has truncate (which means overflow-hidden, text-ellipsis, and whitespace-nowrap), flex items default to min-width: auto, preventing the long text from truncating. min-w-0 overrides this behavior, allowing truncate to actually work.

| with min-w-0 | without min-w-0 |
| --- | --- |
| 截屏2026-07-04 23 18 15  | 截屏2026-07-04 23 18 37 |

## Summary by CodeRabbit

* **New Features**
  * Improved RSS rule feed selection to show feed titles when available, while also displaying the feed URL as supporting detail.
  * Feed selection now supports richer feed information (title + URL) instead of plain URL lists.
* **Bug Fixes**
  * Updated affected-feed matching to consistently store and match by feed URL, while preserving title-based display in the form.

### DIFF
--- a/web/src/pages/RSSPage.tsx
+++ b/web/src/pages/RSSPage.tsx
@@ -1789,7 +1789,7 @@ const DEFAULT_RULE_FORM_STATE: RuleFormState = {
 interface RuleFormFieldsProps {
   state: RuleFormState
   onChange: <K extends keyof RuleFormState>(field: K, value: RuleFormState[K]) => void
-  feedUrls: string[]
+  feeds: FeedInfo[]
   categories: Record<string, Category>
   availableTags: string[]
   idPrefix: string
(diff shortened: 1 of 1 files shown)


s0up4200:
Thanks for the PR! Mind posting a screenshot of the change?

pengchongfu:
My bad for missing the screenshots in the initial PR. They've been uploaded now. Let me know if there's any UI/styling that needs tweaking

s0up4200 (approved):

PR #2158, opened 2026-07-18
fix: sanitize non-UTF-8 bytes from torrent-derived strings

Zmegolaz:
qui crashed (process exit status=2) when searching for and adding a cross-seed torrent which contained a file name with a raw non-UTF-8 byte, like á as Latin-1 `0xE1` instead of valid UTF-8 `0xC3 0xA1`. The byte reached regexp.MustCompile in the release parser's HDR enrichment, and Go's regexp panics on invalid-UTF-8 patterns, killing the process. When looking at this bug I found a few places with similar issues.

**Fix**

The BitTorrent spec requires torrent string fields to be UTF-8, so malformed bytes should be safe to drop. The new helper stringutils.SanitizeUTF8 (wraps strings.ToValidUTF8, no-op on valid input) sanitizes torrent-derived strings at their ingestion boundaries:

- Torrent files we parse: crossseed.ParseTorrentMetadataWithInfo and dirscan.ParseTorrentBytes (names + file paths) (the cause of my crash)
- qBt file lists: SyncManager.GetTorrentFilesBatch, in the fetched clone before caching
- qBt torrent names: no clean ingestion point (they live in the go-qbittorrent library's shared MainData map), so releases.Parser.Parse is the sanitization point for that source, as well as for filesystem names read by dirscan

**Testing**

Added regression tests at each boundary (TestGetTorrentFilesBatch_SanitizesInvalidUTF8, TestParseTorrentMetadataWithInfo_SanitizesInvalidUTF8, TestParseTorrentBytes_SanitizesInvalidUTF8, TestParser_HandlesInvalidUTF8WithoutPanicking), each fails before this change.

**Note**

Malformed bytes are replaced with U+FFFD, not repaired (Latin-1 "Amélie" -> "Am�lie"), the source encoding is unknowable from raw bytes. U+FFFD (rather than dropping) matches what encoding/json and the NFKD matching normalizer produce for invalid bytes, so sanitized strings still compare equal to those forms.

**Reproduction**

This is a bit hard to reproduce, I only hit it when searching for and adding a cross-seed torrent whose file names contained invalid UTF-8. The crash is fatal there because the panic happens in a background goroutine with no recovery, so it takes down the whole process.

The POST below triggers the same panic in the same parser, though it enters via the torrent name rather than a file name inside the torrent (that path additionally requires a matching candidate in qBittorrent), and since it runs on an HTTP request it's caught by the Recoverer middleware, so the app survives. Look for a recovered panic in the qui logs (ERR log system error with recover_info mentioning invalid UTF-8).

```
curl -X POST 'http://localhost:7476/api/cross-seed/apply' \
  -H 'X-API-Token: YOUR_API_KEY' \
  -H 'Content-Type: application/json' \
  -d '{"torrentData":"ZDg6YW5ub3VuY2UzOTpodHRwOi8vdHJhY2tlci5leGFtcGxlLmludmFsaWQvYW5ub3VuY2U0OmluZm9kNjpsZW5ndGhpMzBlNDpuYW1lMzI6TW92aWUuMjAyNC4xMDgwcC5CbHVSYXkueDI2NC1H4VAxMjpwaWVjZSBsZW5ndGhpMjYyMTQ0ZTY6cGllY2VzMjA6NW6npzyYc1PYTbO07PmYYaXIlbNlZQ=="}'
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed UTF-8 in torrent and release text, preventing bad encoding from propagating into parsed torrent names, file paths, and release titles/groups.
  * Reduced risk of failures or panics when processing invalid UTF-8 inputs from qBittorrent and metadata parsing.
  * Prevented incorrect directory root handling when sanitized names/paths contain replacement characters.
* **Improvements**
  * Added a shared UTF-8 sanitization helper and applied it across torrent metadata extraction, directory scan matching, and release parsing.
* **Tests**
  * Added targeted tests covering sanitization and non-panicking behavior.


s0up4200 (approved):
Thanks for the fix!

PR #2163, opened 2026-07-21
feat(automations): add file count condition field

B4reMetal:
## Summary
Adds a File Count condition field to the automations query builder — lets rules filter/sort torrents by how many files they contain

## Summary by CodeRabbit

* **New Features**
  * Added a **File Count** field to the automation query builder.
  * Use file count in conditions, sorting, previews, and automated rule execution.
  * Added translations for the new field across supported languages.

* **Documentation**
  * Documented the File Count field and its behavior in automation conditions.

* **Bug Fixes**
  * File Count rules now evaluate correctly using each torrent’s current file count.


B4reMetal:
Noticed that https://github.com/autobrr/qui/discussions/1190 is related.

s0up4200:
> **Note on provenance:** this review was produced by Claude Opus 5 running a multi-agent review workflow (six independent review lenses over the diff, each finding then passed to an adversarial verifier that tried to refute it). Roughly two thirds of the raw findings were discarded as wrong, unreachable, or pre-existing patterns. What survived is below, and I re-checked the load-bearing claims by hand. Please still verify anything before acting on it, especially the qBittorrent API behaviour in point 2. If something here is wrong, say so and I will retract it.

Overall: the wiring is competent and follows the existing patterns in this package. There is one CI-red blocker, one delete-safety bug, and one design cost that is worth talking through before this ships to large instances.

---

### 1. CI is red: `pt-BR` locale is missing the key

The PR adds `FILE_COUNT` to 8 locales but not `pt-BR`, which landed on develop in 3306d623.

```
$ cd web && node scripts/check-pt-br-coverage.mjs
[Missing Keys] 1 error
  - automations.queryBuilder.fields.FILE_COUNT: "File Count"
  Result:   FAIL
```

`pnpm check:i18n` chains that script and `.github/workflows/lint.yml` runs it. Verified failing on this branch and passing on develop; the other 7 locales pass.

Fix: add `"FILE_COUNT": "Contagem de Arquivos"` to `web/src/i18n/locales/pt-BR/automations.json` (Title Case matches the neighbouring `GROUP_SIZE` -> "Tamanho do Grupo").

### 2. Torrents without metadata report a confident `0`, so low-count delete rules remove them

`internal/services/automations/file_count.go:34` does an unconditional `result[hash] = len(files)`.

qBittorrent answers `GET /torrents/files` with HTTP 200 and an empty array for a magnet that has not resolved metadata yet. go-qbittorrent always returns a non-nil pointer (`methods.go:1404`, `var info TorrentFiles ... return &info`), so the `if files == nil` guard in `GetTorrentFilesBatch` does not reject it. The hash lands in the map with a count of `0`, and the `!ok` guard in `evaluateLeaf` cannot help because the hash *is* present.

Result: a rule such as "delete when `FILE_COUNT <= 1`" with mode delete_files will remove every torrent still sitting in metaDL on the next tick. No real torrent has 0 files, so `0` can only ever mean "unknown".

This package already encodes that assumption elsewhere. `missing_files.go:64` deliberately withholds the map entry when nothing could actually be checked, and `orphanscan/service.go:1247` uses `hasFiles := ok && len(files) > 0`. Two lines to match:

```go
for hash, files := range filesByHash {
    if len(files) == 0 {
        continue // no metadata yet, unknown rather than zero
    }
    result[hash] = len(files)
}
```

Worth a table case in `evaluator_test.go` asserting an empty-file torrent does not satisfy `FILE_COUNT <= 1`.

### 3. The cost model

`detectFileCounts` fetches full file lists for every torrent on the instance, one HTTP request each (there is no batch form of `/torrents/files`), capped at 16-way concurrency. The results are then never reusable across runs:

- `cacheIsFresh` is a flat 5 minute TTL (`internal/services/filesmanager/service.go:335`). The doc comment at `:60` claiming "5 min for active, 30 min for completed" is stale and does not match the code.
- `DefaultRuleInterval` is 15 minutes.

So on the default cadence every hash is already stale when the rule runs. A 5,000 torrent instance means 5,000 HTTP calls every 15 minutes, indefinitely, to compute 5,000 integers that cannot change for a given infohash. Unlike the hardlink index and missing-files detection, there is no `HasLocalFilesystemAccess` gate, so remote seedboxes (the high latency case) now pay this and previously never did. There is also no memo or singleflight, whereas `GetHardlinkIndex` has both, and preview "load more" re-runs the whole detection for every 25 row page.

The cheaper path already exists and already runs on this exact code path. `torrent_files_sync.file_count` is written on every cache fill (`filesmanager/service.go:181`) and `GetSyncInfoBatch` already SELECTs it (`repository.go:430`) immediately before the file rows are hydrated. Today that count is discarded in favour of pulling every file row through two `string_pool` joins just to call `len()`. A count-only read of that column, ignoring the TTL since a file count is immutable per infohash (with the empty-list caveat from point 2), would make steady state one indexed query and only HTTP-fetch hashes that have no sync row at all.

This is the part I would most like your read on, since it is a design call rather than a defect.

### 4. Sort support is half-wired

`FieldFileCount` was added to `IsNumeric()`, which is the sortable-field contract rather than a generic numeric predicate: it gates `SortingConfig.Validate` (`internal/models/automation.go:126` and `:144`) and `SortTorrents`. The backend therefore fully supports simple sorting and score multipliers on FILE_COUNT, and the PR adds the matching `getNumericFieldValue` case and the `sortingConfigUsesField` gate to serve it.

The frontend was not updated. FILE_COUNT is absent from both `SIMPLE_SORT_FIELD_SET` and `SCORE_MULTIPLIER_FIELD_SET` in `WorkflowDialog.tsx`, so it renders greyed out as "Not supported for simple sorting" while the API accepts it, and a workflow imported with `sortingConfig.field = "FILE_COUNT"` cannot be re-saved from the dialog. The `SYSTEM_*` fields already have this same gap so it is not novel, but here the backend plumbing exists only to serve sorting, which makes the inconsistency internal to the PR.

Either add FILE_COUNT to both TS sets, or drop it from `IsNumeric()` plus the `processor.go` case and leave it condition-only.

Related, and only actionable if the sort path stays: `getNumericFieldValue` (`processor.go:1146`) uses a bare map index, so a hash whose fetch failed is ranked as `0` and sorts ahead of genuine single-file torrents in an ASC delete order. That is the established convention of that function (`FREE_SPACE` behaves the same way), so it is a consistency question rather than a clear bug.

### 5. Tests

Four table rows in `evaluator_test.go` and nothing else. `file_count.go` has no test file, the `setupFileCountContext` and `applyRulesForInstance` gating is untested, and the sort path is untested. `TestSortTorrents_Score` in `processor_test.go` is a ready-made template for the last one.

`go test -race -count=1 ./internal/services/automations/...` passes, and `gofmt`, `go vet` and `tsc -b` are clean.

### 6. Docs and scope

The docs row is well placed and alphabetical, but it sits under "Filesystem Fields" next to three entries that all say "requires local filesystem access". File Count does not need local access; it costs one API call per torrent instead, which the docs do not mention. Worth a line given point 3.

Separately, the linked FR (#1190) asks for `count(Files, {# matches "\.(rar|r\d\d)$"}) > 3`, that is a filtered file count for detecting packed releases. A plain total does not answer that, so this should not close it.

---

### What is correct

The gate uses `rulesUseCondition`, which already covers `sortingConfig` through `ruleUsesCondition`, so sorting-only rules do trigger detection in the apply path. Unknown resolving to `false` in `evaluateLeaf` matches the `HAS_MISSING_FILES` and `HARDLINK_SCOPE` convention. Both preview paths are wired and those two are the only `Preview*` functions on the service. No swagger change is needed since `ConditionField` is not enumerated there.

Minimum to merge from my side is points 1 and 2. Point 3 is the one worth discussing before this reaches instances with 10k+ torrents.

B4reMetal:
I have realized that is not as useful as I thought it was. I'm going to close this PR for now and think on the design for a bit. Raw file counts isn't very useful. It needs to be more like the FR, with the extension info included.

PR #2205, opened 2026-08-01
fix(crossseed): treat one-sided HDR and collection tags as compatible

s0up4200:
Many trackers remove HDR tags and collection/service tags from release names. The matcher rejected every pair where only one name carried the tag, so identical content failed with "hdr mismatch" or "collection mismatch". These two checks now reject only when both names carry a tag and the tags differ. This is the same rule that codec, edition, and language already use. The per-file size comparison in the apply stage still rejects candidates whose content differs.

The change makes the special cases for this problem unnecessary. The unknown-season and bracket-anime collection exemptions are removed. The per-indexer webhook collection fallback is removed, because the general rule now covers its one use case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cross-seeding matches when incoming releases have missing collection, HDR, or streaming-service metadata.
  * Reduced false mismatches for compatible TV and movie releases, including scene and site-tagged releases.
  * Improved fallback matching accuracy when release sizes align.
  * Preserved rejection of genuine metadata conflicts, such as incompatible collection tags.
  * Aligned webhook matching with standard release matching for more consistent results.


s0up4200:
@coderabbitai The exact-size fallback scope finding doesn't hold: the live strict mismatch reason for this ATV/ATVP pair is `hdr mismatch`, not `collection mismatch` — verified empirically by asserting your claimed value and watching the test fail with `actual: "hdr mismatch"`. The test comment was correct as written. ce5c0f88 pins the live decision values (`StrictMismatchReason == "hdr mismatch"`, `RelaxedDifferences == {collection, hdr}`) so the test now proves its comment and this can't drift silently.

PR #2213, opened 2026-08-02
feat(crossseed): curb season pack churn in automation flows

s0up4200:
The first ensemble field run downloaded 47 season packs and applied 2, and each later RSS run downloaded the same failed packs again. This PR adds a cooldown for failed diversions. The key is the canonical release name, because the same pack has a different GUID on each indexer. The outcomes `drifted`, `layout_mismatch`, and `already_exists` write a `packfail:` row to `cross_seed_search_history` (no migration), and both automation flows check this row before they download a `.torrent`. An `invalid_torrent` outcome does not start a cooldown, because a different indexer can serve a valid payload. The cooldown expires after 720 minutes, and then the next run tries the pack again.

Four more fixes come from the same field run. Ensemble results with the same name now use one attempt slot, so duplicates from many indexers cannot starve later variants. Ensemble admission now rejects multi-season range packs such as `S01-S05`. Pack suppression now reads the full torrent list of the instance, so a seeded pack outside the run scope also suppresses its search. Drifted season pack runs now record their best real coverage instead of a flat 0/N.

## Summary by CodeRabbit

* **New Features**
  * Added cooldown handling for failed season-pack diversions, with automatic retries after expiration.
  * Improved multi-season detection and suppression of existing seeded or downloading packs.

* **Bug Fixes**
  * Prevented duplicate or unsuitable season packs from appearing in search results.
  * RSS automation now skips recently failed downloads for later reevaluation.
  * Drifted runs preserve partial match details, including matched episodes and coverage.
  * Filtered searches now more reliably detect existing season packs.

### DIFF
--- a/internal/models/crossseed.go
+++ b/internal/models/crossseed.go
@@ -1585,6 +1585,30 @@ func (s *CrossSeedStore) GetSearchHistory(ctx context.Context, instanceID int, t
 	return last, true, nil
 }
 
+// GetLatestSearchHistory returns the most recent search timestamp for a key
+// across all instances. Used for pseudo-keys whose verdict is global, like
+// season-pack diversion failures.
+func (s *CrossSeedStore) GetLatestSearchHistory(ctx context.Context, torrentHash string) (time.Time, bool, error) {
+	const query = `
+		SELECT last_searched_at
+		FROM cross_seed_search_history
+		WHERE torrent_hash = ?
+		ORDER BY last_searched_at DESC
+		LIMIT 1
+	`
+
+	var last time.Time
+	err := s.db.QueryRowContext(ctx, query, torrentHash).Scan(&last)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return time.Time{}, false, nil
+		}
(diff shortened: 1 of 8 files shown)


s0up4200:
Re the itoqa finding on `canonicalReleaseNameKey`: the many-to-one mapping is a deliberate tradeoff, now documented in the code comment and pinned by a test assertion (a23e6c96). Distinct real releases differ by tokens (group, codec, resolution, REPACK), not by re-punctuating an identical letter-digit sequence, so the contrived `ab-c`/`a-bc` collision has no practical analogue in release names; if it ever occurred, the cost is one skipped attempt for one cooldown window (12h default), self-healing. The suggested alternative (keep boundaries plus per-format normalization) was field-tested implicitly today and loses: BeyondHD respaces arbitrary tokens (`DDP 5.1`, `H.265`, `Dual-Audio`), so a known-formats dictionary is unbounded, and each miss costs a real re-download every feed poll rather than a hypothetical rare skip. Field evidence on both sides is in the PR run: rows 2029/2030 (the miss the strip fixes) and run 4162 (both title stylings correctly blocked after the fix).

PR #2214, opened 2026-08-02
fix(torznab): keep structured season/ep params when indexer caps support them

s0up4200:
Cross-seed TV searches against Prowlarr fold season/episode into the free-text query as an `SxxEyy` token for every indexer. API-based indexers whose search field matches release names return zero results for that token (BTN gets `{"Search":"S05","Tvdb":...}` → 0 results, while `{season:5}{tvdbId:...}` returns everything), which made BTN cross-seeding effectively dead for TV. This also blinds the new ensemble season searches on exactly the trackers where season packs live.

Now, when an indexer's advertised Torznab caps cover every structured TV param in the request (`tv-search-season`, and `tv-search-ep` when an episode is set), the params stay structured and Prowlarr translates them to the indexer's native syntax. Indexers without those caps (or with caps never fetched) keep the existing token workaround unchanged. Fixes #2036.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Prowlarr TV searches by preserving season and episode filters when supported by the indexer.
  - Added fallback handling for indexers with missing or unsupported capabilities.
  - Restored the original title query when structured search parameters leave the query empty.
- **Tests**
  - Expanded coverage for supported, unsupported, and missing TV search capabilities.

### DIFF
--- a/internal/services/jackett/service.go
+++ b/internal/services/jackett/service.go
@@ -2620,6 +2620,29 @@ func (s *Service) applyProwlarrTVTokenWorkaround(idx *models.TorznabIndexer, par
 	}
 
 	currentQuery := strings.TrimSpace(params["q"])
+	if prowlarrStructuredTVSupported(idx.Capabilities, params) {
+		// The indexer's caps advertise native season/ep params, so let Prowlarr
+		// translate them per-indexer. Folding an SxxEyy token into q returns zero
+		// results on API-based indexers (e.g. BTN) whose free-text search matches
+		// release names (discussion #2036).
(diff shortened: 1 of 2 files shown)


s0up4200:
Field-validated on a live instance: BTN-scoped TV library scan now sends structured queries (Prowlarr history shows empty q with `{tvdbId:...}{season:N}` per show) and BTN returns results for season searches that previously always came back empty. Note for anyone re-testing an existing install: cached search results can mask the fix for up to 24h — hit "Sync caps" on the indexer to invalidate its cache entries first.

PR #2234, opened 2026-08-04
fix(AutodiscoveryDialog): use composite key for existing indexer map; use synthetic row keys; filter empty-id rows

### LINKED ISSUE #2233: Discover dialog matches existing indexers by name only, so duplicate names update the wrong row
The import step of the Discover Indexers dialog builds a map of existing indexers keyed by name only (`web/src/components/indexers/AutodiscoveryDialog.tsx`, the `existingIndexersMap` built in `handleDiscover`). The database has no unique constraint on the indexer name, so two rows can share a name. When that happens, `Map.set` keeps only the last row, and an import that matches the name updates that row even when the user meant the other one. Discovered rows also use the upstream id for the React key, the checkbox id, and the selection set, so two rows with an empty id share selection state and an empty-id Prowlarr row still reaches the server, which rejects it after the fact.

This behavior predates PR #2231 (surfaced by automated review there, confirmed present on develop). Suggested fix: match existing indexers on backend + base URL + name instead of name alone, use a synthetic per-row key for list rendering and selection, and filter or disable discovered rows that have no usable id before import.

waterWang:
## Summary

Fixes three issues in the Discover Indexers dialog (AutodiscoveryDialog.tsx):

### 1. Duplicate-name collision in existing indexer map

The existing-indexer map was keyed by name only (Map keyed by string), but the database has no unique constraint on the name column. When two rows share a name, Map.set keeps only the last one, causing an import to update the wrong row.

**Fix**: Use a composite key (backend, base_url, name) instead of name alone. Two new helper functions (existingMapKey and discoveredMatchKey) build the composite key consistently.

### 2. Synthetic per-row key for list rendering and selection

indexer.id (the upstream Prowlarr/Jackett indexer id, typed as string) was used as the React key and checkbox id. Discovered rows with an empty id shared the same key and checkbox id, breaking list rendering and selection state.

**Fix**: A rowKey() helper returns the non-empty id when available, otherwise falls back to a row-{index} synthetic key.

### 3. Empty-id rows reaching the server before import

Rows without a usable id were still sent to the import endpoint, which rejected them after the fact.

**Fix**: An isRowUnusable() helper marks rows with no usable id. These rows are rendered dimmed (opacity-50, line-through label) with a disabled checkbox, and the import loop skips them with a toast error. The Select All button also excludes them.

Closes #2233

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved autodiscovery matching to correctly identify existing indexers using backend, URL, and name.
  * Added clearer handling for discovered entries missing usable IDs, including disabled selection and import error reporting.
  * Improved selection, update indicators, and import behavior for discovered indexers.

### DIFF
--- a/web/src/components/indexers/AutodiscoveryDialog.tsx
+++ b/web/src/components/indexers/AutodiscoveryDialog.tsx
@@ -31,6 +31,29 @@ interface AutodiscoveryDialogProps {
   indexers: TorznabIndexer[]
 }
 
+/**
+ * Build a composite key that uniquely identifies an existing indexer.
+ * The database has no unique constraint on the name column alone, so two
+ * rows can share a name.  Keying on (backend, base_url, name) prevents
+ * the wrong row from being updated during import.
+ */
+function existingMapKey(idx: TorznabIndexer): string {
+  return `${idx.backend}|${idx.base_url}|${idx.name}`
+}
+
+/**
+ * Build a composite key for matching a discovered indexer against the
+ * existing-indexer map.  Uses the connection's backend and base URL
+ * together with the discovered indexer's name.
+ */
+function discoveredMatchKey(
+  backend: string,
+  baseUrl: string,
+  name: string,
+): string {
+  return `${backend}|${baseUrl}|${name}`
+}
+
 export function AutodiscoveryDialog({ open, onClose, indexers }: AutodiscoveryDialogProps) {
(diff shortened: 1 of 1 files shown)


s0up4200:
thanks for the PR and the work on this. However I went with a different approach in #2236 and will continue with that version.

PR #2241, opened 2026-08-05
perf(automations): update the hardlink index instead of rebuilding it

s0up4200:
The hardlink index records which torrent files share data on disk. Building it reads every torrent's file list from the database and stats every file, which measures at about 700 ms for 5000 torrents. It was discarded whenever any torrent was added or removed, and every two minutes regardless. Profiling an idle instance and one running a cross-seed library scan put it at the top of the CPU profile in both cases: 28% of the process during the scan, where every added torrent forced another full rebuild, against a scan that is otherwise waiting on indexers. Adding or removing a torrent changes the link counts of the torrents that share its files and of nothing else, so only those are re-read now, and the derived maps are recomputed in memory from the retained scans. The full rebuild remains as a backstop for link changes that no torrent reports, at a longer interval, and the update falls back to it when there is no previous scan, when a file list cannot be fetched, or when more than half the set changed at once.

Deletes must not inherit that lag, so they no longer trust the index. Candidates chosen with hardlink data are re-read off disk immediately before the delete runs, and any whose scope changed since the rule decided, or whose files stopped being readable, are held back and logged. This is stricter than what is on develop today, where a delete can already act on an index up to two minutes old with no verification at all. Rules that delete on ratio or an unregistered tracker never consult the index and are untouched.

## Field results

Measured on an instance with 5163 torrents, 4255 outside links and 1075 duplicate groups, against the same host running develop.

During a cross-seed library scan, `buildHardlinkIndex` was 27.7% of a 180 second profile on develop and does not appear at all here. `updateHardlinkIndex` is 1.76%. A torrent arrival costs about 48 ms instead of about 1110 ms, because the update re-reads the one changed torrent plus the six or seven that share its files, rather than all 5147. Over a 91 minute window, seven of ten arrivals took the incremental path at 19 to 32 ms each, and the three full rebuilds were all the interval backstop. Those rebuilds measured 2432 to 4256 ms on this host, so the stall they remove is longer than the 700 ms figure above suggests.

Removals were verified separately by deleting two torrents whose files stayed on disk through other torrents' hardlinks. The update ran in 15 ms, re-read the three surviving sharers, and left the outside-link count unchanged at 4255. That count is the discriminator: had the survivors kept their stale link counts instead of being re-read, each would have carried more links than the torrent set could explain and flipped to `outside_qbittorrent`, raising it.

Total process CPU is not a useful measure of this change. It sits near 1.4% of one core with or without it, and the periodic spikes on this host come from cross-seed candidate matching, which does over a thousand title lookups in a five minute burst. The case for this change is the work it stops doing, not the CPU it returns.

The delete verification did not run in the field. It only engages for rules that consult the index, and the only delete rule on the test host selects on unregistered trackers.

## Summary by CodeRabbit

* **Performance**
  * Improved hardlink index updates with faster incremental rescans and a 10-minute refresh period.
  * Preserved scan results to avoid rereading unchanged torrent files, with full rebuilds when needed.

* **Reliability**
  * Added safer handling for moved, removed, inaccessible, or newly linked files.
  * Delete actions and previews now revalidate hardlink-dependent decisions before execution.

* **Consistency**
  * Applied consistent hardlink behavior across live runs, previews, and dry runs.
  * Improved cross-instance hardlink scope tracking and link-count accuracy.

### DIFF
--- a/internal/services/automations/hardlink_index.go
+++ b/internal/services/automations/hardlink_index.go
@@ -337,6 +337,12 @@ func (s *Service) updateHardlinkIndex(ctx context.Context, instanceID int, cache
 
 	rescan, staleFileIDs := planTorrentRescan(previous.torrentInfoByHash, torrentByHash)
 
+	// Bail before reading anything when the changes alone already exceed the budget,
+	// so a mass addition does not read half the library off disk twice.
+	if len(rescan)*hardlinkIncrementalChangeRatio > len(torrents) {
+		return nil
+	}
+
 	// Torrents that arrived raise the link count on every file they hardlink to, so
 	// the holders of those files need a second look. Their identities are only known
 	// once the new torrents have been read, hence the two passes.
@@ -405,10 +411,18 @@ func (s *Service) updateHardlinkIndex(ctx context.Context, instanceID int, cache
 // link on it reads as a link outside the client. That is the answer that keeps the
 // torrent out of the delete-safe groups.
 func (idx *HardlinkIndex) scopeAfterRescan(info *torrentFileInfo) string {
-	if idx == nil || idx.buildState == nil || info == nil {
+	if idx == nil || info == nil {
 		return ""
 	}
 	if !info.allAccessible || len(info.fileIDs) == 0 {
+		return ""
+	}
+
+	// Cross-instance augmentation raises uniquePathCount in place, and a preview
+	// request can run it while a scheduled apply verifies its delete candidates.
+	idx.crossScopeMu.Lock()
+	defer idx.crossScopeMu.Unlock()
+	if idx.buildState == nil {
 		return ""
 	}
 
--- a/internal/services/automations/hardlink_index.go
+++ b/internal/services/automations/hardlink_index.go
@@ -12,6 +12,7 @@ import (
 	"maps"
 	"os"
 	"path/filepath"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


s0up4200:
## Field results

Deployed to a live instance with 5147 torrents, 4235 outside links and 1068 duplicate groups, and measured against the same host on `develop`.

The index rebuild was the top CPU consumer during a cross-seed library scan. It no longer appears in the profile.

| | develop | this PR |
|---|---|---|
| Hardlink index share of a 180 s profile | 27.7% `buildHardlinkIndex` | 1.76% `updateHardlinkIndex` |
| Cost per torrent added | ~1110 ms average | 48 ms |
| Total CPU during the scan | 2.3% of a core | 1.26% (pprof), 1.37% (sampler) |

Two torrents arrived during the profiled window. Each one re-read 1 changed torrent plus the 6 or 7 that share files with it, instead of all 5147:

```
08:08:58  hardlink index updated  47ms  rescanned=1 reshared=6  total=5147
08:11:18  hardlink index updated  49ms  rescanned=1 reshared=7  total=5148
```

On `develop` those same two additions cost two full rebuilds. The recorded baseline was 5 rebuilds in 9 minutes at 667, 687, 2577, 897 and 722 ms.

The 10-minute TTL rebuild fires on schedule as the backstop for disk changes that no torrent event reports.

### What these numbers do not say

Total CPU halved, but only part of that belongs to this PR. The hardlink saving accounts for about a quarter of the old profile. The rest of the gap is cross-seed reaching different indexers with different results between the two runs. The attributable claim is the 27.7% to 1.76% shift.

Resident memory reads 338 MB against a 234 MB baseline, with a 385 MB peak. A heap profile puts the retained scan state at about 6 MB, and all of `internal/services/automations` at 10.7 MB, with `HeapAlloc` 93 MB against `Sys` 218 MB. So the index is not holding that memory. The gap stays unexplained, because there is no baseline heap profile to compare against.

The delete verification path did not run in the field. It only engages for rules that consult the index, and the one delete rule on the test host selects on unregistered trackers.

PR #2252, opened 2026-08-06
fix(crossseed): match titles whose word boundaries differ

s0up4200:
Trackers do not put the word boundaries of a show name where the scene name puts them, and the title check compared the two spellings literally. `Kiss.X.Sis` parses to the title `Kiss X Sis`, while Aither lists the same release as `KissXSis`. `Re.ZERO.Starting.Life.in.Another.World` parses to `Re ZERO Starting Life in Another World`, while Aither lists `Re:ZERO -Starting Life in Another World-`, and `NormalizeForMatching` deletes the colon instead of turning it into a space, so the candidate normalizes to `rezero starting life in another world`. Both candidates were byte-identical in size and from the same release group, and both were rejected as `title mismatch`. Reported on Discord by Asinine

The title check compares sets of normalized titles and accepts any overlap, so the fix stores the space-free spelling of each title next to the normal one. The set only grows, so this cannot turn a match into a rejection. Distinct shows still stay apart once the spaces are gone, which the tests pin: `ef a tale of melodies` does not equal `ef a tale of memories and melodies`, and `fbi` does not equal `fbimostwanted`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved title matching when spacing, punctuation, or word boundaries differ between releases.
  * Prevented genuinely different titles from being incorrectly treated as matches.

* **Tests**
  * Added coverage for title variations involving removed spaces and punctuation.

### DIFF
--- a/internal/services/crossseed/matching.go
+++ b/internal/services/crossseed/matching.go
@@ -332,9 +332,16 @@ func rawAKATitleParts(rawName string) []string {
 
 func addNormalizedTitle(titles map[string]struct{}, title string) {
 	normalized := stringutils.NormalizeForMatching(title)
-	if normalized != "" {
-		titles[normalized] = struct{}{}
+	if normalized == "" {
+		return
 	}
+	titles[normalized] = struct{}{}
+
+	// Trackers disagree with scene names about where the word boundaries are:
+	// "KissXSis" for "Kiss.X.Sis", and "Re:ZERO" normalizes to "rezero" because
+	// NormalizeForMatching drops the colon instead of turning it into a space.
+	// Store the space-free spelling too so the two readings can overlap.
+	titles[strings.ReplaceAll(normalized, " ", "")] = struct{}{}
 }
 
 func normalizedTitleSetsOverlap(sourceTitles, candidateTitles map[string]struct{}) bool {
--- a/internal/services/crossseed/matching_title_test.go
+++ b/internal/services/crossseed/matching_title_test.go
@@ -451,6 +451,54 @@ func TestReleasesMatch_AKAVariants(t *testing.T) {
 	}
 }
 
+// Trackers place word boundaries differently from scene names, and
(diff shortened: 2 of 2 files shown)


Audionut:
This is probably more correctly scoped as a, `filename` vs `release name` problem.

Where the filename often has problematic characters removed, and the naming object via torznab is the actual release name of the content, regardless of characters.

Nice to see improvements, but there is a reason that the docs suggest leveraging *arrs for better matching.

s0up4200:
AI GENERATED COMMENT

Closing this. qui already solves these cases without a new matching rule.

The exact-size title rescue accepts all three reported names on develop, as class `title-rescue`, with no code change:

```
rescue=false  Kiss.X.Sis    => rejected  "title mismatch"
rescue=true   Kiss.X.Sis    => accepted  class="title-rescue"
rescue=false  Re.ZERO       => rejected  "title mismatch"
rescue=true   Re.ZERO       => accepted  class="title-rescue"
rescue=false  Fate.strange  => rejected  "title mismatch"
rescue=true   Fate.strange  => accepted  class="title-rescue"
```

Rescue demands a byte-for-byte equal size, which is the safety this PR lacked. Storing the space-free spelling of every title widens the gate for all content, and that makes real shows collide: `Home Town` (HGTV 2016) matches `Hometown` (tvN 2021), and `Blindspot` matches `Blind Spot`, with group, season, episode and resolution all agreeing. The worst path is a season pack episode search, where `ignoreSizeCheck` removes the size gate as well.

To get these matches today, turn on **Rescue title mismatches** in the cross-seed settings. Note that it does nothing while **Skip Recheck** is on, because the two are combined at `service.go:3293`.

@Audionut you called it right as a filename versus release name problem. Worth adding that the arr path cannot repair the candidate side: alias titles only ever enter the source title set, and `CandidateTitles` is never populated anywhere in the codebase. That is why the slash truncation in #2249 needs a fix of its own.

PR #2258, opened 2026-08-07
fix(backups): restore category-managed save paths

ultronknowhere:
## Summary

- request qBittorrent category placement with `autoTMM=true` when a backup manifest has a category but no explicit per-torrent save path
- preserve explicit divergent save paths with `autoTMM=false`
- update the restore regression test to assert category-managed placement

## Problem

Backup capture intentionally omits a torrent-level `savePath` when it matches the category path. During restore, passing the category alone does not make qBittorrent 5.2.2 select that category path; the torrent is added at the default save path and can enter `missingFiles` even when its payload exists under the category path.

## Testing

- [x] Red regression test observed against v1.25.0 behavior
- [x] `go test -race -count=1 -v ./internal/backups`
- [x] `make precommit`
- [x] `make build`
- [x] `golangci-lint run --new-from-rev=origin/develop --timeout=5m`
- [x] Disposable qBittorrent 5.2.2 contract probe: category-only failed; category + Auto TMM and explicit save path passed
- [x] Disposable snapshot → exact preview → incremental restore with the patched build
- [x] Direct qB read-back: correct category path/content path, `stoppedUP`, full payload visibility, zero trackers

No production restore was performed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved torrent restoration when no saved download path is available.
  * Category-based placement now correctly enables automatic torrent management.
  * Restorations with saved paths continue using those paths as expected.
  * Eliminated unnecessary warnings during category-based restores.


ultronknowhere:
Exact-commit integration validation completed for `f62335bb3f0791b972a4e369ebcde2497dfa29d6` against qBittorrent v5.2.2.

- Built the committed revision for `linux/amd64`; runtime `/api/application/info` reported the same full commit.
- Snapshotted a trackerless category-managed torrent whose manifest intentionally omitted `savePath` because it matched the category path.
- Deleted only the torrent metadata, retained the payload, then ran exact preview + incremental restore in an internal Docker network with no external egress.
- Direct qB read-back confirmed `auto_tmm=true`, `save_path=/downloads/restore-extension`, `content_path=/downloads/restore-extension/extension-canary.txt`, correct category/tag, and no `missingFiles` state.
- With active hash checking, qB reached progress `1.0`; after an explicit disposable-only stop it settled at `stoppedUP`.

The validation also exposed a separate pre-existing post-add metadata race for active restores (`startPaused=false`), tracked independently in #2259. It does not change the category-placement result or this PR's scope.

Source gates run on the exact branch:

- `make precommit`
- `make build`
- `go test -race -count=1 -v ./internal/backups`
- independent fail-closed review: no findings

PR #2260, opened 2026-08-07
fix(torrents): resolve peer country flags from country names and codes

MaKTaiL:
## Summary

Fixes an issue where peer country flags (e.g. Brazil) were not displaying in the peer table and peer details panel when qBittorrent returned `country` names instead of 2-letter ISO `country_code` values or when `country_code` was unpopulated or unmapped.

- Added `getCountryCode` helper (`web/src/lib/country.ts`) with strict ISO 3166-1 alpha-2 validation, alias handling (`uk` → `gb`, `el` → `gr`, `tp` → `tl`, `su` → `ru`), 3-letter ISO mappings (e.g. `ala` → `ax`, `cpt` → `cp`), and expanded country name coverage (e.g. `Brazil` / `Brasil` → `br`, `Cabo Verde` → `cv`).
- Updated `web/src/components/torrents/details/PeersTable.tsx` and `web/src/components/torrents/TorrentDetailsPanel.tsx` to resolve country codes from both `country_code` and `country` properties, falling back gracefully when 2-character codes are invalid (e.g. `--` or `XX`).
- Added `resolveCountryCode` in the Go backend handler (`internal/api/handlers/helpers.go` / `internal/api/handlers/torrents.go`) to normalize and populate `CountryCode` in API responses.

## Screenshots

| Before | After |
| :---: | :---: |
| image | image |

## Testing Checklist

- [x] Tested frontend helper unit tests (`npx vitest run src/lib/country.test.ts`)
- [x] Tested backend handler unit tests (`go test -count=1 -v ./internal/api/handlers -run TestResolveCountryCode`)
- [x] Verified alias precedence (`uk` → `gb`, `el` → `gr`), invalid candidate fallback (`--` + `Brazil` → `br`), and boundary alpha-3 entries
- [x] Verified peer country flags render correctly in peer table and mobile/card details view

## Improvements

- Country information is now normalized consistently across torrent peer lists and details.
- Country indicators recognize two-letter and three-letter codes, country names, aliases, and localized names.
- Matching is case-insensitive and ignores surrounding spaces.
- Country values are resolved from available peer data, improving sorting and display consistency.
- Missing or unrecognized country information no longer produces empty flag elements.
- Country flags now use normalized codes and clearer fallback titles when applicable.


s0up4200:
Hi! Thanks for the PR.

Both `country` and `country_code` are always set together in qBit from the same GeoIP lookup, so I'm not sure how you'd get a country name without a code. What qBittorrent version/build are you running?

MaKTaiL:
> Hi! Thanks for the PR.
> 
> Both `country` and `country_code` are always set together in qBit from the same GeoIP lookup, so I'm not sure how you'd get a country name without a code. What qBittorrent version/build are you running?

It seems the AI overengineered this. Upon further investigation the country codes were indeed being sent correctly from qBitorrent. The fix in this PR does fix the issue but it was not the root cause. The actual fix is much simpler than that. I'll cancel this PR and reopen it with the simple fix instead.

PR #2264, opened 2026-08-08
fix(crossseed): replace anacrolix/torrent with autobrr/go-torrent

arjunrn:
I've replaced the depdency `github.com/anacrolix/torrent` with `github.com/autobrr/go-torrent` which is a lightweight fork with the functionality stripped down to only what is needed to load torrent files. It also changes how torrent files are parsed which will address the issue described [here](https://github.com/autobrr/qui/discussions/2262).

Unfortunately there's a small catch. There were tests which relied on [metainfo.BuildFromFilePath](https://pkg.go.dev/github.com/anacrolix/torrent/metainfo#Info.BuildFromFilePath) to write dummy test files to disk and then generate metadata based on that. This helper is not present in the fork. So rather than replicate the function as-is I checked and saw that most of what the helper does is not needed. So the info object is now constructed directly thereby removing the need for the helper.

There was also a test that has been dropped which required the functionality from the missing function. It was used to verify that the pieces on the boundaries have different hashes when the "extra" files are different. This is not a test of any behavior or logic in the codebase itself but simply a demonstration or documentation of the problem. So I deemed it safe to remove. I can put it back but IMO it simply tests that the upstream library actually does what it's supposed to do. So this project is probably not the right place for that.

I'm open to making any changes like putting back the function.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tolerance for malformed tracker-list metadata during torrent parsing.
  * Preserved non-standard torrent metadata and oversized values when updating trackers.
  * Maintained existing cross-seeding, directory matching, and backup behavior.

* **Refactor**
  * Updated torrent metadata handling for improved compatibility and reliability.

* **Tests**
  * Expanded coverage for malformed metadata and metadata preservation.
  * Simplified torrent test generation with deterministic in-memory data.


s0up4200 (approved):
Thanks for the PR! This is a clean swap. I pushed three commits.

Also opened autobrr/go-torrent#5 once that's tagged I'll bump `go.mod` here so this lands with no known parse regressions.

PR #2295, opened 2026-08-11
fix(backups): apply restore category and tags on add only

### LINKED ISSUE #2259: bug(backups): active restore can race post-add category update
## Summary

An incremental restore with `startPaused=false` can report a `set_category` error immediately after qBittorrent accepts the torrent.

Observed while validating #2258 against qBittorrent v5.2.2 on current `develop` (`2a61f00` base plus the PR commit). This appears separate from the save-path change: restore already passes `category` in the initial `AddTorrent` options, then immediately calls `SetCategory` for the same hash.

## Reproduction

1. Create a qB category with save path `/downloads/restore-extension`.
2. Add a trackerless torrent in that category and snapshot it.
3. Delete only the torrent object while retaining its payload.
4. Execute incremental restore with:

```json
{
  "mode": "incremental",
  "startPaused": false,
  "skipHashCheck": false,
  "autoResumeVerified": false
}
```

## Observed

The restore response reports:

```json
{
  "errors": [
    {
      "operation": "set_category",
      "message": "no valid torrents found to set category"
    }
  ]
}
```

Direct qB read-back after settlement showed that the torrent was successfully added and became complete (`stalledUP`, progress `1.0`) with the expected category, tag, save path, and content path. The test torrent was trackerless and the disposable Docker network had no external egress.

## Expected

A successful add should not produce a restore error merely because qB has not exposed the new hash to the immediate follow-up category mutation yet.

## Likely boundary

`applyTorrentPlan` passes category and tags in the `AddTorrent` options, then immediately calls `SetCategory` and `SetTags` for the same new hash. Possible approaches:

- avoid redundant post-add mutations when the add options already supplied those values; or
- use a small bounded retry after add until qB exposes the hash before applying post-add metadata.

This should remain separate from #2258 so the category-derived save-path fix stays minimal.

s0up4200:
## Description

Restore sends `category` and `tags` in the `AddTorrent` options. It then called `SetCategory` and `SetTags` for the same hash. `SetCategory` first looks for that hash in the sync cache. A restore that qBittorrent accepted can therefore report `set_category: no valid torrents found to set category`. This change removes both calls after the add, because the add already applies the values.

A retry with a limit is not sufficient. go-qbt collapses concurrent syncs into one request. The forced sync inside `SetCategory` can therefore return maindata that qBittorrent sent before the add. In one test the torrent was present for 150 ms and the check still failed.

Fixes #2259

## How has this been tested?

I reproduced the reported error against qBittorrent 5.2.2 in Docker. With background syncs, 5 of 10 adds failed. A proxy that delays only the maindata response makes the failure deterministic. qBittorrent applies the category, the tags, the save path, and Auto TMM from the add options alone.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored torrents now receive their categories and tags during creation, improving consistency and avoiding follow-up update failures.
  * Restore operations no longer report errors from separate category or tag updates after a torrent is added.

* **Tests**
  * Enhanced restore coverage to verify that categories and tags are applied correctly during torrent creation.

### DIFF
--- a/internal/backups/restore_execute.go
+++ b/internal/backups/restore_execute.go
@@ -287,6 +287,8 @@ func (s *Service) applyTorrentPlan(ctx context.Context, plan *RestorePlan, appli
 		if len(spec.Manifest.Tags) > 0 {
 			options["tags"] = strings.Join(spec.Manifest.Tags, ",")
 		}
+		// Category and tags ride on the add. A post-add SetCategory raced the
+		// sync cache and failed for a torrent qB had accepted (#2259).
 
 		// Pin the captured per-torrent save path so torrents whose on-disk
 		// location diverges from their category (cross-seed hardlinks, manual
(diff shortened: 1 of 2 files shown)


s0up4200:
@coderabbitai full review

PR #2339, opened 2026-08-15
feat(crossseed): collapse video search matching to size and group

s0up4200:
## Description

Video candidates from search are now admitted on two checks: total size within tolerance, and group/site agreement enforced only when both sides carry one. The title, season, episode, year, date, resolution and checksum gates are gone from video result matching only: search queries still use all of these fields to find candidates, and apply keeps per-file size validation plus recheck as the real identity check. Music keeps the previous cascade unchanged.

Byte-identical candidates kept dying on label noise: AnimeBytes stamps S01 on every entry so season packs failed with `season mismatch`, and absolute-numbered episodes failed against renumbered candidates. The group gate survives because WEB-DLs and remuxes from different groups are the same stream, so sizes collide across groups; a group missing on either side counts as absence of evidence and passes to recheck.

Field test (deep scan on a live library, history cleared, PR image): 408 torrents processed, 63 cross-seeds added across 12 indexers, 11 of them from AnimeBytes, the class that matched zero before this change. Apply-side file validation rejected 31 candidates for size differences and 36 for missing files before any add. Of the adds that needed a recheck, 5 resumed confirmed and 1 stopped at 99.9% and stayed paused for manual review, which is the byte-budget failsafe working. No apply-side group drops, no error-level log lines during the run.

The run also caught one real bug in the first build: the name inside a downloaded .torrent can parse to an invented group from a stray trailing tag (`[Batch]`) while the real group sits in the site field, which silently killed correct pairings at apply. Fixed by comparing group and site as identity sets with overlap counting as agreement, plus a log on that drop path.

## How has this been tested?

Table tests for the new classifier path, a routing test that pins music to the old cascade, and an end-to-end test for the apply-side prefilter. Each gate was mutation-checked: break the gate, watch only its covering test fail, revert. Full `go test -race -count=1` on the package holds the same pre-existing failure set as develop. `make precommit` and `make build` pass.

## Checklist

### DIFF
--- a/internal/services/crossseed/alt_pass_usable_test.go
+++ b/internal/services/crossseed/alt_pass_usable_test.go
@@ -24,7 +24,7 @@ func TestSearchResultUsable(t *testing.T) {
 	const (
 		webripDotted  = "Law.and.Order.Special.Victims.Unit.S05.1080p.AMZN.WEBRip.DD2.0.x264-NTb"
 		webdlDotted   = "Law.and.Order.Special.Victims.Unit.S05.1080p.AMZN.WEB-DL.DD+2.0.x264-NTb"
-		otherShow     = "Law.and.Order.Organized.Crime.S05.1080p.AMZN.WEB-DL.DD+2.0.x264-NTb"
+		otherGroup    = "Law.and.Order.Organized.Crime.S05.1080p.AMZN.WEB-DL.DD+2.0.x264-XYZ"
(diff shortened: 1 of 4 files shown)


s0up4200:
closing in favour of #2344.

replayed every byte-identical rejection from the 2026-08-15 deep scan through current develop: 53 of them, all true matches, and they die on a one-sided `checksum mismatch` (35), `season mismatch` (10) and `episode mismatch` (1). three narrow fixes get all 53 without touching the rest of the cascade.

this PR gets the same matches, but the floor becomes 5% of the total size plus a group tag, and a perfect-layout add sets `skip_checking` so the recheck it leans on never actually runs.

PR #2345, opened 2026-08-16
fix(crossseed): recover exact-size matches with compatible names

### LINKED ISSUE #2344: crossseed: exact-size matches rejected by harmless name differences
## Problem

Cross-seed matching can reject a real match before qui downloads the candidate torrent. This happens when two release names describe the same files in different ways.

qui has three different size values:

- qBittorrent reports the local torrent's `TotalSize`. qui falls back to its wanted `Size` only when `TotalSize` is unavailable.
- Torznab, RSS, or autobrr can advertise a size before download.
- After download, qui reads the actual total from the `.torrent` metadata.

Two positive, equal reported sizes are useful evidence. They do not prove that the bytes are equal.

A deep scan on 2026-08-15 found 53 rejected candidates with the same reported size as their source. Manual review found that all 53 were real matches. The rejected names had one of these differences:

- a CRC checksum appeared in only one name;
- an indexer changed a season or episode number;
- `rls` split one fansub name across its `Group` and `Site` fields.

## Required matching behavior

When both reported sizes are positive and exactly equal, qui can reconsider these narrow differences:

- A checksum on only one side is missing information, not a conflict.
- Different season numbers can match when both releases are season packs.
- Different episode numbers can match when both releases are individual episodes.
- A split fansub name can match when one side's `Site` equals the other side's explicit release group.

The fansub rule checks the public name and selected-file tags. Every explicit group must agree with the proposed identity. A shared site label is not enough.

## Where the rule applies

- Interactive search, Library Scan, scheduled search, and completion search use the same Torznab classifier.
- RSS uses the title and byte count already present in the feed. It does not fetch extra torrent files to measure candidates.
- autobrr `/check` uses the passive `.Size` value. This value can be exact, rounded, or `0`. Reading it does not download the torrent file.
- autobrr `/apply` receives the intended torrent file and the original announcement name. qui reads the actual total and repeats the match against current local sources.

If autobrr reports `0`, `/check` uses a narrow name-only preflight. That preflight can approve one intended download. It cannot approve an add. `/apply` must still create and replay a new source-bound decision from the actual torrent.

No autobrr code change, check-to-apply token, or extra scrape is required.

## Safety rules

Exact reported size must not override:

- different titles, except through the existing optional title rescue;
- different resolutions, years, artists, or two conflicting checksums;
- TV matched with non-TV;
- a season pack matched with an individual episode;
- two different explicit release groups;
- zero, missing, or merely similar sizes.

The decision is private. It stays bound to one qBittorrent instance, source hash, and advertised candidate name.

After download, qui parses the torrent name and selected files again. It checks each current difference separately. A recorded difference cannot hide a new codec, group, or other conflict.

If the downloaded name changes a populated field, that field must belong to a difference that the advertised match already accepted. Cached ARR aliases remain valid during this check.

## Recheck behavior

A one-sided checksum or an ordinary descriptive difference still goes through the normal file and layout checks.

Season, episode, title-rescue, and split-group matches are more ambiguous. qui adds them paused and requests a full qBittorrent piece check. It resumes only after the worker observes a live piece check and then sees 100% completion.

qBittorrent's `checkingResumeData` state does not count as a piece check. A qBittorrent restart clears any earlier observation. If qui misses a very fast check, it leaves the torrent paused for manual review.

When **Skip recheck** is enabled, qui skips the source that needs verification before RSS or autobrr performs its intended download. A separate strict local match can still be used.

## Result

The original 53-pair replay accepts all 53 matches:

- 11 require a full recheck;
- 42 keep the existing fast path.

Tests cover both checksum directions, advertised-name changes, WEB-DL/WEBRip relabeling, TV shape boundaries, group provenance, ARR aliases, source binding, strict alternatives, RSS, autobrr, completion search, and regular, hardlink, and reflink verification.

Implemented in #2345.

s0up4200:
## Description

Cross-seed matching rejected real matches when two names described the same files differently.

This PR uses equal, positive reported sizes as controlled evidence. It can recover a one-sided CRC, compatible season or episode numbering, and a fansub group that `rls` split across `Group` and `Site`.

Equal size is not byte proof. The downloaded torrent, file list, paths, layout, piece boundaries, and current source identity must still pass validation.

## What changed

The matching decision now records the exact source and the advertised candidate name. Apply checks every current difference and rejects new metadata that the advertisement did not authorize.

The same decision flow now covers:

- interactive, scheduled, Library Scan, and completion Torznab searches;
- RSS automation when the feed has a positive exact byte count;
- autobrr's external `/check` and `/apply` flow.

RSS checks the feed title and size before its one intended torrent download. Relaxed decisions stay private and bind to one qBittorrent instance and source hash.

autobrr `/check` uses only data that autobrr already has. The `.Size` template does not fetch the torrent file. An exact value can enable the controlled fallback. A rounded value can only pass normal strict matching within tolerance. A zero value uses a narrow advisory name check.

autobrr `/apply` now accepts the original `torrentName` with `torrentData`. qui reads the actual total from the torrent metadata, repeats the match against live sources, and then uses the existing source-bound replay. Clients that omit `torrentName` keep the old strict apply path.

No autobrr source change, check-to-apply token, or extra scrape is required. The season-pack webhook keeps its separate behavior.

## Candidate rejection order

`classifySearchCandidate` now checks required identity before it collects detailed difference evidence, and it compares the plain `Group` and `Site` fields before it reads tag provenance. The matching rules are unchanged. Only the order of the work changed, so a candidate that cannot match stops at the first cheap check.

This matters most for the zero-size announcement path, which scans every eligible local torrent. Many trackers announce without a size, so that path is common rather than rare. On a library of about 6600 torrents, an announcement that matched nothing took 2 to 3 seconds before this change and now completes in well under one second. The shared search path benefits from the same ordering.

## Safety and rechecks

Hard boundaries stay hard. The fallback does not permit different titles, resolutions, years, media shapes, conflicting checksums, conflicting explicit groups, or unequal reported sizes.

Season, episode, title-rescue, and split-group matches are added paused. qui requests a full qBittorrent piece check and resumes only after it observes a live check followed by 100% completion.

`checkingResumeData` does not count as a piece check. A restart clears earlier check evidence.

When **Skip recheck** is enabled, RSS and autobrr reject verification-required sources before their planned download. A separate strict or soft-safe source can still be used.

Codec, source, HDR, edition, one-sided checksum, and similar descriptive differences retain the normal fast path after file and layout validation.

Fixes #2344

## How has this been tested?

The original field replay contains 53 same-size candidates that manual review identified as real matches. All 53 are accepted:

- 11 require full verification;
- 42 keep the existing fast path.

Focused race tests cover:

- both one-sided checksum directions and downloaded metadata drift;
- season, episode, title, and TV-shape boundaries;
- raw-name and selected-file release-group provenance;
- ARR aliases, source binding, and strict alternatives;
- RSS source selection, filters, duplicate checks, one-download behavior, and multi-instance results;
- autobrr passive-size checks, unknown-size preflight, actual metainfo size, legacy apply, and partial success;
- completion-search decision transport;
- regular, hardlink, and reflink recheck behavior;
- qBittorrent restart and resume-data states.

Checks completed:

- `make precommit`
- targeted `go test -race -count=1` suites for crossseed and API handlers
- `make test-openapi`
- `make build`

The crossseed package passes in full. An earlier revision of this description reported 15 failing tests as pre-existing on `develop`. That was wrong. Those failures come from running the tests with `TMPDIR` on a different filesystem than the repository, which makes the hardlink tests compare mismatched device IDs. Setting `TMPDIR` to a directory on the same filesystem makes the package green, and CI agrees.

The repository-wide `make test` was not run, following the project guidance for crossseed changes.

## Checklist

- [x] My PR title follows the [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#summary) format.
- [x] I have read [CONTRIBUTING.md](https://github.com/autobrr/qui/blob/develop/.github/CONTRIBUTING.md).
- [x] I ran `make precommit` and the relevant tests locally.
- [x] I updated the OpenAPI contract for the new optional request field.
- [x] This change does not modify the database schema.

## Summary by CodeRabbit

- **New Features**
  - Improved exact-size cross-seed matching for compatible metadata differences.
  - Added reported-size matching for RSS and Autobrr workflows with safer validation.
  - Improved file-derived release details and strict-match selection.
  - Added optional announcement-name validation for Autobrr applications.
- **Bug Fixes**
  - Verification-required matches now pause for complete rechecks, including hardlink and reflink workflows.
  - Invalid structural, checksum, or ambiguous group matches remain rejected.
- **Documentation**
  - Clarified fallback matching, recheck requirements, and skip-recheck behavior.


luckylittle:
This had undesirable effect on accepting the REPACKS. See this example:
`Title.2026.1080p.WEB.H264-SOMEGROUP`
`Title.2026.REPACK.1080p.WEB.H264-SOMEGROUP`

PR #2356, opened 2026-08-17
feat(crossseed): coordinate pooled partial completion

### LINKED ISSUE #1590: Add pooled partial cross-seed completion automation for hardlink/reflink mode
## Summary

When cross-seeding in hardlink/reflink mode, qui already knows when a torrent is a valid partial match: enough files can be reused, the remaining differences are acceptable under the existing size tolerance and safety checks, and only a subset of files still need to be acquired.

At that point, qui should be able to automate completion of the missing files across the entire related cross-seed set, instead of leaving each partial torrent isolated and requiring manual start/recheck handling.

This issue proposes a new user-facing automation mode that:

- allows eligible partial hardlink/reflink cross-seeds to enter a started/downloading state,
- groups related cross-seeds into a shared pool,
- chooses the best torrent in that pool to download each missing file,
- hardlinks/reflinks completed missing files into the other related cross-seeds that need them,
- triggers rechecks on those other cross-seeds,
- and starts them automatically once they reach 100%.

This should minimize duplicate physical downloads and turn partial reuse into a coordinated completion workflow.

Related: #1589

## Problem

Today, when multiple cross-seeds are all related to the same original source content, partial reuse can leave several torrents in a state where:

- each torrent already reuses most of the data,
- each torrent is missing one or more additional files,
- several torrents may need the same missing file,
- and the user must manually decide which torrent to start, wait for it to complete, then rely on manual or ad hoc follow-up behavior for the others.

That misses a large automation opportunity.

If four related cross-seeds all need the same missing file, the ideal outcome is:

- one torrent downloads the real bytes,
- the other three get those bytes via hardlink/reflink,
- all related torrents are rechecked,
- all become complete without downloading duplicate copies.

More generally, if one cross-seed adds some missing files and another cross-seed adds even more, qui should continue coordinating the pool until all possible shared files have been propagated and all eligible torrents have been completed.

## Proposed feature

Add a user-facing option for hardlink/reflink mode that enables automated pooled completion for partial matches.

Suggested behavior:

1. When a cross-seed candidate passes the existing partial-reuse safety checks, qui may allow it to be added in a started/downloading state instead of always leaving the user to handle it manually.
2. qui groups all related cross-seeds into a pool based on the original source content that triggered the cross-seed search.
3. qui tracks which files are:
   - already present,
   - missing,
   - being downloaded by one torrent in the pool,
   - or available to be linked/cloned into other torrents in the pool.
4. Before auto-starting a partial torrent, qui evaluates the pool and chooses the best torrent to acquire the missing files.
5. Once a missing file completes in one torrent, qui hardlinks/reflinks that file into every other eligible torrent in the same pool that needs it.
6. qui triggers rechecks on those dependent torrents.
7. Torrents that become complete are automatically started or resumed when they reach 100%.
8. If later torrents in the same pool contain additional missing files not yet acquired, the process repeats until the pool is fully resolved or no further safe automation is possible.

## Key idea

qui should treat related partial cross-seeds as a coordinated data-sharing pool, not as independent torrents that happen to overlap.

The system should aim to ensure that each distinct missing file is physically downloaded only once whenever that is possible under the existing matching and safety rules.

## User-facing option

Add a setting for cross-seed automation, conceptually similar to:

Automatically complete eligible partial hardlink/reflink cross-seeds

Expected semantics:

- only applies when hardlink or reflink mode is enabled,
- only applies when the torrent already passed the existing reuse acceptance rules,
- only applies when qui can safely coordinate the remaining files,
- allows qui to start one or more selected pool members automatically,
- and allows qui to finish the rest of the pool through link/clone propagation plus rechecks.

This should be opt-in.

## Pool model

A pool should represent cross-seeds that are all ultimately related to the same original source content.

At minimum, the pool likely needs to track:

- the source torrent that triggered the search or completion workflow,
- all cross-seeds derived from that source,
- their hashes and current states,
- the files each torrent already has,
- the files each torrent is missing,
- which missing files overlap across torrents,
- which torrent is currently designated to acquire each missing file,
- and whether a completed file has already been propagated to dependent torrents.

The grouping needs to survive beyond the initial add so that later completions can continue benefiting the earlier pool members.

## Selection strategy

Before starting a partial torrent, qui should ideally determine which pool member is the best downloader for the remaining missing files.

That decision could consider:

- which torrent covers the largest number of currently missing shared files,
- which torrent would unlock the most other pool members,
- which torrent minimizes total duplicate download volume,
- and whether one torrent is effectively a superset of another.

The earlier file comparison step already has most of the information needed to estimate this. It would be better to make the decision there rather than after several partial torrents have already been added without coordination.

## Completion workflow

A high-level workflow could be:

1. Build or update the pool when cross-seeds are added.
2. Compute the file overlap graph for related torrents.
3. Select the best torrent(s) to auto-start for missing-file acquisition.
4. Wait for file completion events.
5. For each newly completed file, propagate it via hardlink/reflink into other eligible pool members.
6. Recheck affected torrents.
7. Auto-start/resume any torrents that now reach 100%.
8. Re-evaluate the pool for remaining missing files and repeat if needed.

## Safety constraints

This automation should still respect the existing acceptance rules.

It should not bypass:

- size mismatch tolerance,
- piece-boundary safety checks,
- any rules that currently reject unsafe partial reuse,
- or any constraints on when hardlinking/reflinking is valid.

This feature is about automating the completion of already-accepted partial matches, not relaxing the safety model.

## Why this belongs with #1589

Issue #1589 fixes a lower-level prerequisite: preserving the intended managed destination root across hardlink/reflink fallback and reuse chains.

This issue builds on that. A pooled completion workflow only works reliably if each incoming torrent keeps its own intended managed destination root instead of collapsing into another related torrent's save path when modes change.

In short:

- #1589 is about correct destination-root ownership
- this issue is about coordinated lifecycle automation across related partial cross-seeds

## Expected behavior

With this feature enabled:

- qui may automatically start the best eligible partial torrent in a related pool,
- missing files downloaded by that torrent are reused across the other eligible pool members,
- duplicate physical downloads are minimized,
- dependent torrents are rechecked automatically,
- and related cross-seeds become complete and startable with minimal manual intervention.

## Acceptance criteria

- A new opt-in user-facing setting exists for pooled partial completion automation in hardlink/reflink mode.
- Related cross-seeds are grouped into a persistent pool keyed to shared source content.
- qui can identify overlapping missing files across pool members.
- qui can choose a preferred torrent to acquire missing files automatically.
- When a missing file completes in one torrent, qui can propagate it to other eligible pool members via hardlink/reflink.
- qui rechecks dependent torrents after propagation.
- qui auto-starts or resumes dependent torrents once they become complete.
- The process can repeat when later pool members contain additional files not yet acquired.
- The automation respects all existing size and safety checks.
- The workflow preserves each torrent's intended managed destination root as described in #1589.

Audionut:
## Description

Adds opt-in pooled partial cross-seed completion for hardlink and reflink modes.

- Persists pools, members, exact file matches, claims, retries, and recovery state for SQLite and PostgreSQL.
- Coordinates one downloader per pool, propagates verified files safely, rechecks targets, and resumes eligible members.
- Exposes the setting through the API, OpenAPI schema, Web UI, translations, and user documentation.

Closes #1590

## How has this been tested?

- `make precommit`
- `go test -race -count=1 ./internal/models ./internal/api/handlers ./internal/services/crossseed ./pkg/hardlinktree ./pkg/reflinktree ./pkg/sharedextents`
- `make test-openapi`
- `pnpm check:i18n`
- `npx vitest run src/pages/CrossSeedPage.pooledCompletion.test.tsx src/lib/__tests__/api.crossseed.test.ts`
- `make build`

- Live-tested successfully with qBittorrent on Windows/ReFS in reflink mode; hardlink mode remains automated-test-only.

Not included; the UI change is one opt-in Cross-Seed settings checkbox.

## Checklist

## AI disclosure

Yes. Codex wrote most of the implementation under user direction using GPT-5.

## Summary by CodeRabbit

- **New Features**
  - Added optional pooled partial completion for cross-seed torrents, disabled by default.
  - Coordinates related downloads, verification, file propagation, retries, and recovery.
  - Added global controls for pooled completion and automatic download limits.
  - Supports hardlink and reflink workflows, including cross-volume handling.
  - Added search decision details for clearer filtering and matching outcomes.

- **Bug Fixes**
  - Improved torrent detection, recheck readiness, and incomplete-file handling.
  - Improved post-add file visibility and retry behavior.
  - Prevented unsafe path traversal and symlink escapes.
  - Added safeguards against stale updates and incomplete verification.

- **Documentation**
  - Added guidance for pooled partial completion, limits, and recovery behavior.


Audionut:
@coderabbitai review

Audionut:
@coderabbitai review

Audionut:
@coderabbitai full review

cmd/qui/main.go
```diff
@@ -881,6 +887,12 @@ func (app *Application) runServer() {
 	// Graceful shutdown with timeout
 	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 	defer cancel()
+	partialPoolCancel()
+	select {
+	case <-partialPoolDone:
+	case <-ctx.Done():
+		log.Error().Msg("timed out waiting for partial completion coordinator shutdown")
+	}
```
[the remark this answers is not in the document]
Audionut: No change: the implementation plan explicitly requires the coordinator wait to share the existing 30-second shutdown context, so HTTP shutdown receives the remaining budget and the whole sequence stays bounded. A separate context would change that specified lifecycle contract.

internal/services/crossseed/partial_pool_files.go
```diff
@@ -0,0 +1,322 @@
...
+	resolvedExisting, err := filepath.EvalSymlinks(existing)
+	if err != nil {
+		return fmt.Errorf("resolve partial pool path: %w", err)
+	}
+	for i := len(suffix) - 1; i >= 0; i-- {
+		resolvedExisting = filepath.Join(resolvedExisting, suffix[i])
+	}
+	resolvedTarget := filepath.Join(resolvedExisting, filepath.Base(targetAbs))
+	if err := pathInside(resolvedRoot, resolvedTarget); err != nil {
+		return fmt.Errorf("partial pool path escapes through symlink or reparse point: %w", err)
```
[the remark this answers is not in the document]
Audionut: Fixed in e3bf8bf5: existing final targets are resolved with filepath.EvalSymlinks before containment validation, with a passing Windows file-symlink escape regression.

internal/services/crossseed/service.go
```diff
@@ -15089,46 +15192,60 @@ func (s *Service) processReflinkMode(
...
+				s.recordPartialPoolRecheckRequested(ctx, pooledMember)
+				s.signalPartialPoolWake(partialPoolWake{poolID: pooledMember.PoolID})
+				statusMsg += " - pooled completion pending"
+			case addPolicy.ShouldSkipAutoResume():
+				statusMsg += s.titleRescueMonitorSuffix(candidate.titleRescue, candidate.InstanceID, torrentHash)
+				log.Debug().
+					Int("instanceID", candidate.InstanceID).
+					Str("torrentHash", torrentHash).
+					Msg("[CROSSSEED] Reflink mode: skipping auto-resume per add policy")
+				statusMsg += addPolicy.StatusSuffix()
```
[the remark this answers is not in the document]
Audionut: No change to the verdict: the plan explicitly permits incomplete reflink clones, including disc layouts after full recheck, when actual missing bytes fit the configured budget. e3bf8bf5 adds a regression proving within-budget disc reflinks wait while over-budget ones block.

internal/services/crossseed/partial_pool.go
```diff
@@ -0,0 +1,1802 @@
...
+	if err != nil {
+		s.markPartialPoolPropagationManual(ctx, targetMember, targetFile, "file propagation failed: "+err.Error())
+		return false
+	}
+	if targetMember.Mode == models.CrossSeedPartialPoolModeHardlink {
+		if s.partialPoolCreated == nil {
+			s.partialPoolCreated = make(map[int64]*hardlinktree.Created)
+		}
+		s.partialPoolCreated[targetFile.ID] = created
+	}
```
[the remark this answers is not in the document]
Audionut: Fixed in e3bf8bf5: manual propagation now deletes only the target file rollback handle. TestPartialPoolManualPropagationDropsCreatedHandle passes.

internal/web/swagger/openapi.yaml
```diff
@@ -6850,7 +6850,11 @@ components:
         autoResumeMaxDownloadMb:
           type: integer
           minimum: 0
-          description: Max missing data (MiB) a new cross-seed may auto-resume with after its recheck. 0 means only fully complete torrents auto-resume. Default 50.
+          description: Max missing data (MiB) a new cross-seed may auto-resume with after its recheck; also gates pooled partial acquisition. 0 means only fully complete torrents auto-resume. Default 50.
+        pooledPartialCompletionEnabled:
+          type: boolean
+          default: false
```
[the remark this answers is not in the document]
Audionut: Fixed in e3bf8bf5: the PATCH schema omits the default while the full settings schema retains default false. make test-openapi passes.

internal/services/crossseed/partial_pool.go
```diff
@@ -0,0 +1,1802 @@
...
+			continue
+		}
+		torrent, found := partialPoolInventoryTorrent(inventory, member)
+		if !found {
+			for _, file := range member.Files {
+				delete(s.partialPoolCreated, file.ID)
+			}
+			_ = s.automationStore.MarkPartialPoolMemberRemoved(ctx, member.ID, "torrent no longer exists in qBittorrent")
+			continue
+		}
```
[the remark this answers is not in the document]
Audionut: No change to removal semantics: the implementation plan explicitly makes a successful loaded inventory authoritative and requires missing torrents to be marked removed immediately. e3bf8bf5 adds TestObservePartialPoolMembersRemovesMissingTorrent covering durable member and pool removal.

internal/models/crossseed_partial_pool.go
```diff
@@ -0,0 +1,887 @@
...
+		CrossSeedPartialPoolMemberStatusWaiting,
+		now,
+		CrossSeedPartialPoolMemberStatusAcquiring,
+	)
+	if err != nil {
+		return false, fmt.Errorf("claim partial pool downloader: %w", err)
+	}
+	changed, err := result.RowsAffected()
+	return changed == 1, err
+}
```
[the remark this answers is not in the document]
Audionut: No change: SQLite accepts this UPDATE alias in the supported runtime. go test -race -count=1 ./internal/models -run TestCrossSeedPartialPoolClaimsAndTransitionsPersist exercises ClaimPartialPoolDownloader and passes.

internal/database/migrations/091_add_cross_seed_partial_pools.sql
Audionut, line 1: @coderabbitai No change for "duplicate instance/torrent index": the implementation plan explicitly requires both UNIQUE (instance_id, torrent_key) and an index on those columns in both dialects, so the migrations retain that contract.

internal/models/crossseed_partial_pool_test.go
Audionut, line 1: @coderabbitai Addressed "replace the hardcoded Windows root path" in e3bf8bf5: the fixture now uses filepath.Join(t.TempDir(), "pool").

internal/services/crossseed/service.go
Audionut, line 1: @coderabbitai No change for "extract duplicated hardlink/reflink recheck flow": the branches retain mode-specific safety, logging, and queue behavior; the alleged disc divergence is intentionally enforced at the shared partialPoolPostRecheckVerdict seam. A wide parameterized helper would add abstraction without changing behavior.

internal/models/crossseed_partial_pool.go
Audionut, line 1: @coderabbitai Addressed "validation error text identifies only the file index" in e3bf8bf5: each invalid field now has a specific error, covered by table tests.

internal/services/crossseed/partial_pool_coordinator_test.go
Audionut, line 1: @coderabbitai Addressed "add coverage for the member-removal path" in e3bf8bf5: a loaded inventory miss now has a regression asserting durable member and empty-pool removal.

internal/services/crossseed/partial_pool_filesystem_test.go
```diff
@@ -0,0 +1,287 @@
...
+		},
+	}
+	service.markPartialPoolPropagationManual(t.Context(), member, member.Files[0], "synthetic verification failure")
+	require.NotContains(t, service.partialPoolCreated, member.Files[0].ID)
+
+	pool, err = store.GetPartialPool(t.Context(), pool.ID)
+	require.NoError(t, err)
+	require.Equal(t, models.CrossSeedPartialPoolMemberStatusManual, pool.Members[0].Status)
+	require.Equal(t, models.CrossSeedPartialPoolFileStatusManual, pool.Members[0].Files[0].Status)
+}
```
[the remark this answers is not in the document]
Audionut: Fixed in 758e2ebc: the regression now asserts the exact persisted LastError string. go test -race -count=1 ./internal/services/crossseed -run TestPartialPoolManualPropagationDropsCreatedHandle passes.

Audionut:
@coderabbitai full review

internal/services/crossseed/service.go
```diff
@@ -15399,6 +15476,13 @@ func (s *Service) processReflinkMode(
 			Msg("[CROSSSEED] Reflink mode: failed to build plan, aborting")
 		return handleMaterializationError(fmt.Sprintf("Failed to build reflink plan: %v", err))
 	}
+	addPolicy := PolicyForSourceFiles(sourceFiles)
+	pooledCompletion := s.partialPoolAdmissionEnabled(ctx, instance, hasExtras, req, verifyBeforeSeed || addPolicy.DiscLayout)
```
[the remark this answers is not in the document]
Audionut: Fixed in 56679677: pooled admission now reserves the disc-layout full-completion gate for hardlinks; reflinks retain the configured-budget verdict. Targeted race tests pass.

internal/web/swagger/openapi.yaml
Audionut, line 1: @coderabbitai Not changed "Document the 200 MiB auxiliary-file exception": a zero budget intentionally requires AmountLeft <= 0 in both pendingResumeSatisfied and postRecheckBudgetSatisfied, so documenting a sidecar exception at 0 would be inaccurate.

Audionut, line 1: @coderabbitai Addressed "Include missing files in the recheck causes" in 56679677: both OpenAPI descriptions now include extra or missing files. make test-openapi passes.

web/src/i18n/locales/ko/crossseed.json
Audionut, line 1: @coderabbitai Addressed "Add title mismatches to all four locale descriptions" in 56679677: Korean, Portuguese, Ukrainian, and Simplified Chinese now include title. pnpm check:i18n passes.

s0up4200:
did a deep review pass, a few things need sorting before merge.

re-admission: the alias-found branch in `RegisterPartialPoolMember` (`internal/models/crossseed_partial_pool.go:248`) returns the old member row untouched, so a torrent removed and re-added before reconciliation notices keeps stale status/root/files and never gets the initial recheck. since admission only runs right after an add, an existing row there is stale by definition. needs a deliberate reset of status, root path and file rows, not just swapping return-old for insert.

cross-device: any materialization error at `partial_pool.go:2594` marks the file manual permanently, including EXDEV and the windows reflink volume rejection. multi-fs pools are legal since nothing checks devices across members. heads up on the fix: sending the file back to missing on its own loops forever re-picking the same donor, so it needs to remember the failed pair and let downloader selection fall through to self-download or another donor.

save-root guard: `normalizePathForComparison` lowercases both sides (`partial_pool.go:961`), so a case-only move passes on linux and fs ops run under the stale root. dont change the shared normalizer, the folding elsewhere is deliberate. add a host-semantics compare local to that guard, case-sensitive on unix.

two changes touch non-pool paths even with the setting off and need a second look: `waitForPostAddRecheckReady` now skips issuing a recheck when torrents are already checking, and `extractSuccessMessage` rewrites success messages for all search results, not just pooled ones.

is the reflink disc-layout path skipping `ShouldSkipAutoResume` intentional? its commented as deliberate, but #1590 says pooling shouldnt bypass the existing partial-reuse safety rules. want an explicit yes on that one.

and while 090/092 are still drafts I want the schema sorted before merge: the retry/review sub-state serialized into `last_error` as prefixed strings and parsed back (`partial_pool.go:1584`) should be real columns, `ReportedSeeders` can go (it only feeds the 5th sort tiebreaker), and the new BOOLEAN columns should be INTEGER like every other table. the settings insert already writes an int into the pg BOOLEAN.

Audionut:
@s0up4200 Thanks for the deep pass. Addressed all points in the latest commits:

- **Re-admission** (`5b727aa1`): alias-equivalent registration now keeps the member identity/pool but atomically refreshes its hashes, mode, root, admission state, and retry/review state; invalidates dependent in-flight source claims; and replaces the stale file rows. The member returns through initial verification/recheck. A regression test covers stale member/file state and dependent claims.
- **Cross-device propagation** (`a08956bf`): `EXDEV`, native Windows `ERROR_NOT_SAME_DEVICE`, and the Windows reflink volume rejection are now pair-specific failures. The target returns to `missing`, and the source/target file pair (including current roots) is remembered process-locally. Source selection, pending-source gating, and downloader deferral all skip rejected pairs, allowing another donor or normal self-download. Other materialization failures still go manual. Tests cover alternate-donor selection, downloader fallthrough, and Windows error normalization.
- **Save-root guard** (`65543027`): added a guard-local `partialPoolRootsEqual` comparison. It normalizes paths but remains case-sensitive on Unix and uses case folding only on Windows. The shared normalizer is unchanged.
- **Non-pool behavior** (`65543027`, with lint cleanup in `5f49b435`): a running piece check is treated as ready to receive the explicitly requested recheck; only resume-data checking is waited out. `extractSuccessMessage` now returns pool detail only for the private `partialPoolPending` result, so ordinary successes retain `added via <indexer>`.
- **Disc-layout safety** (`65543027`): no, the pooled reflink bypass was not intended. Hardlink and reflink now derive the same recheck policy; disc layouts require full completion, are not admitted to partial pooling, and use a zero-byte resume budget while the existing add-policy auto-resume guard remains honored.
- **Draft schema** (`31cf5c23`): the migrations are now `091`/`093` after the numbering moved on `develop`. Review pause, resume attempts, and recovery attempts are real columns; `last_error` no longer serializes/parses workflow state. `ReportedSeeders` was removed end-to-end. Downloader ranking remains verified state, reusable bytes, unlocked members, amount left, then stable instance/key identity. All new boolean columns use `INTEGER`, including PostgreSQL.

Validation completed: `make precommit`, `make build`, and `go test -race -count=1` for `./internal/models`, `./internal/database`, `./internal/qbittorrent`, `./internal/services/crossseed`, and `./pkg/reflinktree`. A fresh built server also returned HTTP 200 from `/health`.

s0up4200:
Thanks @Audionut!

Is the PR body stale, or has this not been tested on a live instance yet?

internal/models/crossseed_partial_pool.go
```diff
@@ -0,0 +1,1109 @@
...
+
+// MarkPartialPoolMemberRemoved records qBittorrent removal only while the
+// member's pool and admission timestamp still match, then deletes an empty pool.
+func (s *CrossSeedStore) MarkPartialPoolMemberRemoved(ctx context.Context, poolID, memberID int64, admittedAt time.Time, reason string) error {
+	tx, err := s.db.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	defer func() { _ = tx.Rollback() }()
+	result, err := tx.ExecContext(ctx, `UPDATE cross_seed_partial_pools SET status = status WHERE id = ?`, poolID)
```
Audionut, line 1192: @coderabbitai Addressed "Serialize removals with concurrent re-admissions" in 22a3a42c. Removal now locks the pool first and matches the observed pool, member, and admission generation; registration validates that pool lock and retries alias resolution if deletion wins. Covered by TestCrossSeedPartialPoolReAdmissionReplacesStaleState. Please verify this fix path.

✅ Confirmed as addressed by @Audionut

internal/services/crossseed/partial_pool.go
```diff
@@ -0,0 +1,3361 @@
...
+	resetRejectedPairClaim := func() {
+		empty := ""
+		if !s.transitionPartialPoolFile(ctx, targetFile, models.CrossSeedPartialPoolFileStatusMissing, models.PartialPoolFileMutation{
+			SourceFileID: models.NullableInt64Update{Set: true},
+			LastError:    &empty,
+		}) && targetSnapshot != nil {
+			targetSnapshot.stateRetryPending = true
+		}
+	}
+	if targetFile.SourceFileID == nil {
```
Audionut, line 2685: @coderabbitai Addressed "Recover propagation claims after source deletion" in 22a3a42c. An orphaned propagating claim with source_file_id NULL now CAS-resets to missing and clears source/error state; a failed CAS schedules snapshot retry. Covered by TestPartialPoolOrphanedPropagationClaimResetsOrRetries. Please verify this fix path.

✅ Confirmed as addressed by @Audionut

internal/services/crossseed/partial_pool.go
```diff
@@ -0,0 +1,3361 @@
...
+			if file.SourceFileID != nil {
+				sourceMember, sourceFile = partialPoolFileByID(pool, *file.SourceFileID)
+			}
+			rolledBack, retry := s.rollbackLivePartialPoolHardlink(ctx, file, pool)
+			if retry {
+				return false
+			}
+			if rolledBack {
+				if sourceMember != nil && sourceFile != nil {
+					s.rejectPartialPoolPropagationPair(sourceMember, sourceFile, member, file)
```
Audionut, line 1909: @coderabbitai Addressed "Reject donors whose hardlinks fail verification" in 22a3a42c. The donor pair is captured before rollback, rejected after a successful hardlink rollback, and only then rechecked so selection can advance. Covered by TestPartialPoolHardlinkMissingFilesSettlesRollbackBeforeExceptionalState. Please verify this fix path.

✅ Confirmed as addressed by @Audionut

internal/services/crossseed/partial_pool.go
```diff
@@ -0,0 +1,3361 @@
...
+	pool *models.CrossSeedPartialPool,
+	targetMember *models.CrossSeedPartialPoolMember,
+	targetFile *models.CrossSeedPartialPoolMemberFile,
+	snapshots map[int64]*partialPoolMemberSnapshot,
+) *models.CrossSeedPartialPoolMemberFile {
+	for _, sourceMember := range pool.Members {
+		if sourceMember.ID == targetMember.ID || sourceMember.Status == models.CrossSeedPartialPoolMemberStatusManual || sourceMember.Status == models.CrossSeedPartialPoolMemberStatusRemoved {
+			continue
+		}
+		if !s.partialPoolMemberModeEnabled(ctx, sourceMember) {
```
Audionut: @coderabbitai Addressed "Skip disabled members during donor selection" in 22a3a42c. Both source selection and the availability predicate now exclude donors whose recorded link mode or local-filesystem access is disabled, allowing fallback donors or download. Covered by selector and downloader fallback tests. Please verify this fix path.

✅ Confirmed as addressed by @Audionut

internal/services/crossseed/partial_pool.go
```diff
@@ -0,0 +1,3361 @@
...
+	fileByIndex, valid := partialPoolCurrentFiles(member, files)
+	if !valid {
+		return false
+	}
+	snapshot.files = files
+	snapshot.fileByIndex = fileByIndex
+	return true
+}
+
+func (s *Service) refreshPartialPoolPropagationMembers(
```
Audionut, line 1161: @coderabbitai Addressed "Batch live refreshes across propagated files" in 22a3a42c. Reconciliation now caches each participant refresh result across both propagation passes, so file lists and torrent state refresh once per member rather than once per file. Covered by TestPartialPoolCompletedFilesPropagateAndSettleEveryDeferredMember refresh-count assertions. Please verify this fix path.

✅ Confirmed as addressed by @Audionut

Audionut:
@s0up4200 The PR body was stale—this has been live-tested with a real qBittorrent instance on Windows/ReFS in reflink mode, and I have amended the PR body.

Two representative examples from the live logs (release names, hashes, tracker names, and paths intentionally anonymized):

- **Example A — multi-file release:** the completion search added 11 matching cross-seeds with 0 failures. qui selected one pool member to fetch missing data, then reflinked the newly available files into two other members (four successful file propagations), asked qBittorrent to recheck them, and resumed the eligible torrents. It then selected another member to fetch its remaining ~5.4 MiB. The captured run had no warnings or errors.
- **Example B — single-file release:** the completion search added 6 matching cross-seeds with 0 failures. qui selected one downloader, then reflinked the newly available ~100 MiB file into two other members, rechecked both through qBittorrent, and resumed the eligible torrents. Both file propagations completed, with no warnings or errors.

In plain terms, each missing file was downloaded by one pool member, reused locally for the compatible torrents without another full copy, verified by qBittorrent, and then the pool continued automatically. Both selected runs exercised the expected workflow through propagation, verification, and resume.

Audionut:
@coderabbitai review

Audionut:
@coderabbitai Addressed the Windows-form traversal coverage from review 5030086779 in 31dc0954. The test now covers leading slash, leading backslash, drive-letter, and UNC escape inputs; the focused test, full crossseed race suite, and pre-commit gate pass.

Audionut:
@coderabbitai review

PR #2396, opened 2026-08-22
test(services): regression coverage for the fsops callsite refactors

com6056:
## Description

Test-only coverage for the fsops call-site refactors in this branch. Three things that the refactors made correct but that nothing asserted:

- **Rollback under cancellation.** Both link modes and the season-pack apply path wrap their rollback in `context.WithoutCancel`, because the fsops methods early-return on `ctx.Err()` and a run cancelled inside `AddTorrent` would otherwise leave a partial link tree on disk forever. New tests cancel the run from inside `AddTorrent` and assert the base dir is empty afterwards; each fake records how many entries existed at add time so the assertion cannot pass vacuously. Removing the `WithoutCancel` at either link-mode site or in `rollbackSeasonPackTree` fails them.
- **Fail-closed backend resolution.** The local-match path resolves a filesystem backend at three points (`getSourceFileIDs`, the candidate lookup, and the source re-resolve before the shared-extent probe). A backend outage discards all hardlink evidence, so each site records `verificationErr` rather than returning "not linked", otherwise a strict-mode delete dialog reads a transient outage as "no cross-seeds found" and lets you delete the source. One test per site, plus one through `FindLocalMatches(..., strict=true)` to show the recorded error actually reaches the caller. The re-resolve test also pins the call ordering (exactly three lookups) via a counting store.
- **`resolveLocalTorrentFile`'s backslash and drive-prefix rejection.** Table test over POSIX and Windows escaping shapes, including a `C:/evil.mkv` row since that is the one Windows shape the backslash clause does not already catch. The interesting cases are the ones that carry no traversal at all, `AC\DC - Back In Black.mkv` and `dir/AC\DC.mkv`, because backslash is a legal filename byte on Linux, and rewriting it into a nested path makes a file that exists read as missing. Deleting that clause fails five cases: the two legal Linux names plus the Windows traversal, rooted and UNC shapes, which on Linux are single path elements the traversal guards never see.

It also fixes an existing test: `TestInjector_Inject_RollsBackLinkTreeOnAddFailure` never reached `AddTorrent`. The request was a partial match without `DownloadMissingFiles`, so `Inject` bailed at the partial-match check (`inject.go:250-252`) and `failingTorrentAdder.AddTorrent` was never called, meaning the test was exercising the "download missing files disabled" path, not the add failure it is named for. It now sets `DownloadMissingFiles: true` and asserts the adder was called, and a cancellation variant sits beside it.

Scope caveat: `rollbackSeasonPackTree`'s own `WithoutCancel` was already pinned by `TestRollbackSeasonPackTree_RunsUnderCancelledContext`, so the season-pack test's increment is the apply call site reaching rollback with the threaded `planBuild.backend` under a cancelled ctx, since `ApplySeasonPackWebhook` gets that backend from `assembleSeasonPack`. The `context.WithoutCancel` fallback resolve at `seasonpack.go:435` is only reachable when `planBuild.backend` is nil, which this test does not reach; a comment in the test says so.

No production code changed. Targets #1915's branch directly so it lands with the refactor.

## How has this been tested?

- `go test -race -count=1 ./internal/services/crossseed/ ./internal/services/dirscan/`, green.
- Every new test was mutation-verified: the guard it protects was removed, the test was confirmed to fail, and the guard restored. That covers the three `context.WithoutCancel` rollback sites, the three `verificationErr` assignments, the backslash clause, and (for the existing-test fix) the `DownloadMissingFiles` flag that makes the add actually happen.
- `go vet` and `GOOS=windows go vet` on both packages; the new traversal table carries POSIX and Windows cases.
- `make fmt gofix-changed` plus golangci-lint v2.13.0 via `go run` with `--new-from-merge-base=develop` (the Go half of `make precommit`, run directly because the binary on my PATH predates the repo's Go target) reports nothing in these files. No frontend files touched.

## Checklist

- [x] My PR title follows the Conventional Commits format
- [x] I have read CONTRIBUTING.md
- [x] I ran `make precommit`'s Go half (fmt, gofix, golangci-lint) and the tests pass locally

## AI disclosure

Written with Claude Code. The tests were drafted and iterated by the model, then each one was mutation-verified against the guard it claims to protect before being committed; what to assert, the scope caveat on the season-pack fallback branch, and the existing-test fix were reviewed by hand.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup after interrupted or failed torrent additions, ensuring partially created link trees are removed.
  * Strengthened local match handling so backend resolution failures fail safely and report verification errors.
  * Improved validation of local file paths, rejecting traversal, rooted, UNC, drive-based, and invalid path names.

* **Tests**
  * Added regression coverage for cancellation, rollback, hardlink/reflink behavior, failed additions, and invalid local paths.


nitrobass24:
@com6056  Ran your mutations myself, they all check out — including the `DownloadMissingFiles` one, that test really was on the wrong branch.

One that didn't: the season pack test. Nil out `planBuild.backend` and it still passes, since the fallback resolve is already wrapped in `WithoutCancel`. So it was only re-pinning `rollbackSeasonPackTree`, which we already had. Made the pool store go offline on add failure so the fallback can't cover for it.

Same on the path table — delete `cleanName == "."` and every row stays green. Added a row.

Pushed both to your branch, plus folded the hardlink/reflink cancel tests into a table, reused your fixture in `DownloadMissingDisabled`, and merged the base branch in.

Separate thing I noticed while reading the table,  a name like `AC\DC - Back In Black.mkv` gets skipped in `forEachLocalTorrentFile` with a bare `continue` and no `verificationErr`, so a torrent where every file has a backslash gives us no hardlink evidence and strict mode says "no cross-seeds found" instead of "failed to check". My call in c41f74d9, just noting these tests lock it in. Opened a separate issue since this exist on develop #2402

nitrobass24 (approved):

PR #2418, opened 2026-08-24
fix(torznab): wait for the test search before reporting ok

### LINKED ISSUE #2388: fix(torznab): indexer test reports ok when the test search actually failed
## Summary

`POST /api/torznab/indexers/{id}/test` reports success for an indexer whose test search actually failed. The endpoint returns `{"status":"ok"}`, writes `last_test_status=ok`, and the UI paints a green badge and a success toast — for an indexer that is unreachable, returning HTTP 500, or rejecting the query.

Nothing corrects it afterwards, so the indexer keeps showing as Working until someone tests it again and gets the same false pass.

## Why it happens

`SearchGeneric` is asynchronous: it schedules the search and returns `nil` as soon as the task is queued. The handler treats that `nil` as the test result.

`internal/api/handlers/jackett.go` (TestIndexer):

```go
testReq := &jackett.TorznabSearchRequest{
    ...
    OnAllComplete: func(*jackett.SearchResponse, error) {
        // the search's actual outcome arrives here, and is dropped
        cancelTest()
    },
}

if err = h.service.SearchGeneric(testCtx, testReq); err != nil {
    cancelTest()
}

if err != nil { /* error path */ }

// reached for every asynchronous failure
if updateErr := h.updateTestStatusWithTimeout(id, "ok", nil); updateErr != nil { ... }
RespondJSON(w, http.StatusOK, map[string]string{"status": "ok"})
```

The service does forward the failure — the handler just has nowhere to put it. `internal/services/jackett/service.go`, in `resultCallback`:

```go
if req.OnAllComplete != nil {
    req.OnAllComplete(nil, err)
}
return
```

So the only failures the endpoint catches today are the ones `SearchGeneric` rejects synchronously before scheduling: a nil service, an empty query, or a failure to resolve the indexer selection. Every network-level failure — connection refused, DNS failure, HTTP 500, an indexer that returns an error document — arrives at the callback and is discarded.

## Impact

- The indexer list shows a green "ok" badge (`web/src/components/indexers/IndexerTable.tsx`), and filtering by test status counts the broken indexer as working.
- The UI trusts the HTTP 200 directly rather than re-reading the record: `IndexersPage.tsx` `handleTest` calls `updateIndexerTestState(id, "ok")` and raises a success toast; `handleTestAll` reports "N successful" the same way.
- `TorznabIndexerStore.UpdateTestStatus` has exactly one caller, this handler, so no background path ever repairs the stored status.

For contrast, the ARR equivalent is synchronous and reports the real outcome: `ArrHandler.TestInstance` returns `{success: testErr == nil}`, and the service writes the status after the test actually ran (`internal/services/arr/service.go`).

## Reproduction

1. Configure a torznab indexer whose URL returns HTTP 500 for every request (or point it at a closed port).
2. `POST /api/torznab/indexers/{id}/test`.
3. Response is `200 {"status":"ok"}`.
4. `GET /api/torznab/indexers/{id}` — `last_test_status` is `ok`, `last_test_error` is null.
5. In the UI, the indexer shows the green badge and the test raises a success toast.

## Fixing it

The mechanical part is small: read the error the callback already receives and record it.

```go
OnAllComplete: func(_ *jackett.SearchResponse, err error) {
    defer cancelTest()
    if err == nil {
        return
    }
    msg := err.Error()
    if updateErr := h.updateTestStatusWithTimeout(id, "error", &msg); updateErr != nil {
        h.logTestStatusUpdateError(updateErr, id, "error")
    }
},
```

The part that needs a decision is the HTTP contract, because the current one is what makes the UI lie:

- **Wait for the outcome** — block on the callback (or a channel it closes) bounded by the detached context, and report the real result like the ARR handler does. Matches what a "Test" button means to a user and needs no frontend change. The detached context is already 30s, and `timeouts.WithSearchTimeout` respects an existing deadline, so that 30s governs the search rather than being cut short by the adaptive budget.
- **Keep returning on schedule** — leave the 200 meaning "queued", and change the frontend to stop asserting `ok` from the response and re-read `last_test_status` instead. Cheaper on the server, but the success toast keeps lying until the refetch lands.

Either way `handleTest` and `handleTestAll` in `IndexersPage.tsx` need to stop deriving "ok" from the HTTP status alone.

## Provenance

Not a regression from anything in flight. `OnAllComplete` has been discarding the error since #629 (2025-11-24), where its body was `// Ignore results for connectivity test`. Surfaced by an automated reviewer on #2386, which touches that callback only to release the detached context; the false-pass behaviour predates it, and fixing it is a behaviour change to the test endpoint rather than part of a lint pass.

AshSgDe29071999:
`POST /api/torznab/indexers/{id}/test` reported success as soon as `SearchGeneric` queued the job. That call returns nil before the search runs, so an indexer that then failed (connection refused, HTTP 500, error document) still got `last_test_status=ok`, a green badge, and a success toast.

The handler now waits for `OnAllComplete` (or the existing 30s timeout) and records that result. Synchronous schedule errors still fail immediately.

Fixes #2388

## Testing

`go test ./internal/api/handlers -run TestWaitForIndexerTestResult -count=1`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved indexer connectivity tests to accurately report asynchronous search failures and timeouts.
  * Prevented tests from being marked successful before search completion is confirmed.
* **Tests**
  * Added coverage for callback errors, scheduling failures, successful completion, and timeout handling.

### DIFF
--- a/internal/api/handlers/jackett.go
+++ b/internal/api/handlers/jackett.go
@@ -778,7 +778,9 @@ func (h *JackettHandler) TestIndexer(w http.ResponseWriter, r *http.Request) {
 	// itself. The timeout bounds the detached context, and the completion
 	// callback releases it as soon as the search is actually done.
 	testCtx, cancelTest := context.WithTimeout(context.Background(), 30*time.Second) //nolint:gosec // G118: deliberately outlives the request, see above
+	defer cancelTest()
 
+	resultCh := make(chan error, 1)
 	testReq := &jackett.TorznabSearchRequest{
 		Query:            "test",
 		Limit:            1,
@@ -786,16 +788,13 @@ func (h *JackettHandler) TestIndexer(w http.ResponseWriter, r *http.Request) {
 		CacheMode:        jackett.CacheModeBypass,
 		SkipHistory:      true,
 		SkipCachePersist: true,
-		OnAllComplete: func(*jackett.SearchResponse, error) {
-			// Results are irrelevant for a connectivity test; this is only here
-			// to stop holding the detached context once the search finishes.
-			cancelTest()
+		OnAllComplete: func(_ *jackett.SearchResponse, searchErr error) {
+			resultCh <- searchErr
 		},
 	}
 
-	if err = h.service.SearchGeneric(testCtx, testReq); err != nil {
-		cancelTest()
-	}
+	scheduleErr := h.service.SearchGeneric(testCtx, testReq)
+	err = waitForIndexerTestResult(testCtx, scheduleErr, resultCh)
 
 	// Update test status in database
 	if err != nil {
@@ -815,6 +814,22 @@ func (h *JackettHandler) TestIndexer(w http.ResponseWriter, r *http.Request) {
 	}
 
 	RespondJSON(w, http.StatusOK, map[string]string{"status": "ok"})
+}
+
+// waitForIndexerTestResult returns the error from scheduling the search, or
+// the error delivered to OnAllComplete. SearchGeneric is asynchronous: a nil
+// return only means the job was queued, so treating it as success made
+// POST .../test report ok for indexers that then failed (#2388).
+func waitForIndexerTestResult(ctx context.Context, scheduleErr error, result <-chan error) error {
+	if scheduleErr != nil {
+		return scheduleErr
+	}
+	select {
+	case err := <-result:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
 }
 
 func (h *JackettHandler) updateTestStatusWithTimeout(id int, status string, errorMsg *string) error {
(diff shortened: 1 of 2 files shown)


s0up4200 (changes requested):
You did not follow the PR template. Simply running go tests does not mean you verified this is a working fix.

nitrobass24:
Heads-up: #2422 landed a few hours after this one for the same issue (#2388) — I didnt catch someone had already started on it, sorry for the overlap. 

One thing worth carrying over whichever PR moves forward: waiting on `OnAllComplete` alone still passes one failure shape. `resultCallback` in the service converts a deadline-exceeded search into a nil-error partial response (`OnAllComplete(response, nil)` with the indexer absent from `CoveredIndexerIDs`), so an indexer that hangs past the adaptive search timeout — which can be shorter than the 30s bound here — still records `ok`. CodeRabbit flagged it on #2422 and it's reachable; the fix there gates ok on `resp.FullyCovered()`. Happy to close mine if you'd rather fold that in here, or vice versa.

s0up4200:
#2422 supersedes this one

PR #2423, opened 2026-08-24
fix(crossseed): fail closed on unresolvable torrent file names

### LINKED ISSUE #2402: Unresolvable torrent file names fail open in strict-mode local match checks
`forEachLocalTorrentFile` skips any file whose name `resolveLocalTorrentFile` rejects with a bare `continue` — nothing is recorded. So if every file in a torrent gets rejected, the source produces zero hardlink evidence and the local match check reports "not linked" rather than "couldn't check". In strict mode that surfaces in the delete dialog as "no cross-seeds found", and you can delete a source that does have cross-seeds.

Everywhere else in that path we deliberately fail closed — the three backend-resolution sites record `verificationErr` for exactly this reason. This is the one spot that doesn't.

Names that hit it: anything carrying a backslash (legal filename byte on Linux, e.g. `AC\DC - Back In Black.mkv`), plus anything shaped like traversal or a rooted/UNC/drive path.

Pre-existing on develop, not from #1915 — develop rewrites the backslash into `AC/DC - ...`, which doesn't exist, so it dies at the `os.Lstat` miss instead of the guard. Same silent skip either way.

The fix needs to split two cases that currently share the `continue`:

- **rejected by the guard** — the name can't be mapped to a path at all, so we have no evidence and should say so (record a `verificationErr`)
- **stat miss** — normal for a torrent that isn't fully downloaded, must stay a best-effort skip or every incomplete torrent trips strict mode

Low frequency, but the failure mode is deleting data.

nitrobass24:
## Description

`forEachLocalTorrentFile` skipped any file name the resolver refused with a bare
`continue` — nothing was recorded. If every name in a torrent is refused, the source
produces zero hardlink evidence and the local match check concludes "not linked"
rather than "couldn't check". In strict mode (delete dialogs) that surfaces as
"no cross-seeds found", so a user can delete a source that does have cross-seeds.
Every other failure in that path deliberately fails closed by recording
`localMatchContext.verificationErr`; this was the one spot that didn't.

The name that hits it in practice carries a backslash — a legal filename byte on
Unix, e.g. `AC\DC - Back In Black.mkv`. The resolver rewrote `\` into `/`, aiming
the lookup at a nested path the torrent never named, so it died at the `os.Lstat`
miss instead of the guard. Same silent skip either way.

Changes:

- `resolveLocalTorrentFile` already refuses backslash-carrying names on develop
  (landed with #1915); this PR does not change the guard. qBittorrent emits
  torrent file names `/`-separated on every platform, so a backslash can only be a
  literal filename byte from a Linux or macOS client, and the table row for
  `Season 01\Episode 01.mkv` records that a nested-looking name is refused, not
  rewritten.
- `forEachLocalTorrentFile` returns the first refused name as an error. Callers with
  a `localMatchContext` — `getSourceFileIDs` and the candidate FileID pass in
  `localLinkedMatchType` — record it into `verificationErr` (first error wins, same
  pattern as the existing sites), so strict mode returns "failed to verify local file
  relationship" instead of an empty match list.
- Blast radius, stated so nobody is surprised: `verificationErr` lives on the
  `localMatchContext` for the whole `FindLocalMatches` call, so one refused name on
  the source or on any candidate that reaches the linked-copy check fails the entire
  strict response for that source torrent until the file is renamed. A delete dialog
  can therefore show a verification error while listing no cross-seeds. Best-effort
  mode is unaffected.
- A resolved path that is missing or not a regular file stays a silent best-effort
  skip. Partially downloaded torrents are normal and must not trip strict mode.
- The positive-match short-circuit is preserved: when a shared inode is found, the
  candidate pass reports the match and no error.

Reflink pairing (`collectLocalTorrentFiles`) and the mediainfo lookup ignore the new
error explicitly: both sides of a pairing already went through the FileID pass that
records refusals, and the mediainfo path has no match context.

Fixes #2402

## How has this been tested?

Package tests only. A live repro needs a qBittorrent instance holding backslash-named,
hardlinked cross-seeded torrents, which I could not provision.

Added to `internal/services/crossseed`:

- a strict-mode `FindLocalMatches` test whose source file list contains only a
  backslash-carrying name: expects the verification error, not an empty match list
  (best-effort mode still returns the match)
- a strict-mode test where only the candidate torrent carries the unresolvable name,
  with the source file hardlinked so the candidate FileID pass actually runs — this
  pins the candidate-side recording, which the source-side test alone masks
- a test that a resolvable-but-missing file does not set `verificationErr` and still
  returns best-effort results
- a table test for the resolver covering POSIX and Windows escaping shapes
  (`../`, `..\`, rooted, drive-letter, UNC, `.`, `..`, backslash-in-name)

Mutation-checked per site: restoring the bare `continue` fails the source-side test;
dropping the recording in `getSourceFileIDs` fails it too; dropping the candidate-side
recording in `localLinkedMatchType` fails the candidate-side test.

Ran: `gofmt -l`, `go vet` and `GOOS=windows go vet` on the package,
`go test -race -count=1 ./internal/services/crossseed/`,
`golangci-lint run --new-from-merge-base=develop`, `make build`. All clean.

## Checklist

## AI disclosure

Yes. Effectively all of the diff — code and tests — was written with Claude Code
(Claude Opus) from a description of the bug and the intended fail-closed behavior,
then adversarially reviewed by the maintainer: each recording site was mutation-tested
individually, which surfaced and closed an unpinned candidate-side case.


s0up4200:
> *This was generated by AI during code review.*

One question about the guard change, on Windows.

`resolveLocalTorrentFile` now refuses every name that carries a backslash, where develop rewrote it into `/`:

```go
strings.ContainsRune(name, '\\')
```

On Linux this is right. `AC\DC - Back In Black.mkv` is one file name, and the rewrite aimed the lookup at a directory the torrent never named.

Windows is the open case. If qBittorrent on a Windows host can return `Season 01\Episode 01.mkv` from `torrents/files`, that name resolved before and fails the guard now. Under this PR a failed guard is no longer a silent skip, it sets `verificationErr`, so every torrent with a nested path would give a hard error in the delete dialog instead of its cross-seed list. That trades a rare silent fail-open for a frequent loud fail-closed on one platform.

Two small signals that separators are why the rewrite exists: `resolveLocalTorrentFile` was added in #2183, the ReFS block-clone feature, and the rewrite was in it from the first commit. `TestResolveLocalTorrentFile` covers `..\` and a drive letter, but has no nested `Season 01\Episode 01.mkv` row, so nothing pins the behaviour either way.

If qBittorrent always sends `/` on every platform, the change is safe, and a table row asserting that a nested backslash name is refused would record that decision. If it does not, the guard cannot tell the two shapes apart from the string alone and needs another signal, such as trying the rewritten form when the literal name misses.

nitrobass24:
No, and the row in 26bd4dea records that. qBittorrent's `torrents/files` emits `Path::data()`, the internal `/`-normalized string every Path is built from via `QDir::cleanPath`; `toNativeSeparators` only runs in `Path::toString()`, which the API uses for `save_path`. That is why qui runs `filepath.FromSlash` on the save path and never on file names. qui already depends on it elsewhere: the file tree splits `file.name` on `/` only (`TorrentDetailsPanel.tsx:836`) and crossseed uses `path.Base`/`path.Ext` on torrent names (`seasonpack.go`, `parsing.go`). If a Windows client ever sent `Season 01\Episode 01.mkv`, the file tree would already be broken for every Windows user. So a literal backslash can only arrive from a Linux or macOS client where it is a filename byte, exactly the `AC\DC` case.

One correction to the body while I am here: the refusal itself is not this PR's change. #1915 landed it on develop; this PR only turns that silent skip into a strict-mode `verificationErr`. Body updated to say so.

com6056:
Two small things from a pass over the merged result on current develop, neither blocking.

`verificationErr` lives on the `localMatchContext` for the whole `FindLocalMatches` call, so one refused name on the source or on any candidate that reaches the linked-copy check fails the entire strict response for that source torrent until the file is renamed. The body already says first error wins and this is the right fail-closed shape, but the blast radius is worth one more clause so nobody is surprised by a delete dialog that errors while listing nothing.

`normalizeTorrentRelativePath` still carries the `\` to `/` rewrite. Its only caller is `collectLocalTorrentFiles`, which only ever sees names that already cleared `resolveLocalTorrentFile`, so that branch is dead and it reads against the guard develop already has. Not yours to fix here now that #1915 landed the guard, just easier to delete while the file is open.

Clean merge onto develop locally, `go vet` and `go test -race -count=1 ./internal/services/crossseed/...` both green. Nice job turning the table into a real fix 🙌

nitrobass24:
@com6056 both taken.

Dead rewrite: gone in 454c8c32. `normalizeTorrentRelativePath` is now `strings.ToLower(path.Clean(name))` with a one-line note that every name reaching it already cleared `resolveLocalTorrentFile`, so the guard is the only place backslashes are decided.

Blast radius: the body now says it outright. One refused name on the source, or on any candidate that reaches the linked-copy check, fails the whole strict response for that source torrent until the file is renamed, so a delete dialog can error while listing nothing. That is the fail-closed shape #2402 asked for, and it is better stated than discovered.

com6056 (approved):
Approving. Both points from my last pass are in `454c8c32`, and I checked it again today. It merges clean onto `develop` at `7eb4fa1c`, `go test -race` on `internal/services/crossseed` is green, and a refused name still lands in `verificationErr`. Thanks for carrying this one through the #1915 squash! 🙌

PR #2432, opened 2026-08-24
fix(crossseed): disable download path on recheck adds so complete cross-seeds don't stall at 0%

statelycurmudgeon:
## Description

Cross-seed adds that pin an explicit `savepath` (the common case — the cross category's path usually diverges from the matched torrent's, so autoTMM is turned off and the save path is pinned) rely on an **in-place recheck** to reach 100%.

If the target qBittorrent instance has a temp/incomplete **download path** enabled (*Preferences → Downloads → "Keep incomplete torrents in"*), qBittorrent routes the still-incomplete (0%, pre-recheck) torrent to that temp dir. The recheck then hashes an **empty directory**, stays at 0%, never completes — so it never moves to `savepath` — and the recheck-resume worker eventually abandons it **paused at 0%**:

```
info  Successfully added cross-seed torrent   (matchType=exact)
warn  Recheck resume absolute timeout reached, removing from queue   (elapsed ≈ 3,600,000 ms)
```

A manual force-recheck can't recover it either, because the active storage is still the empty temp dir. The `.fastresume` shows the split: `qBt-savePath` = the (populated) link dir, active `save_path` = the empty temp dir, progress 0.

**Fix:** set `useDownloadPath=false` whenever the cross-seed add pins an explicit `savepath`, so the recheck reads the real files at `savepath` instead of the empty temp dir. This only applies alongside an explicit savepath (autoTMM off), mirroring how the manual add API already pairs `savepath` / `autoTMM` / `useDownloadPath` (`internal/api/handlers/torrents.go`). The `skip_checking` complete-match paths (hardlink/reflink/season-pack) are unaffected — they complete instantly, so the temp path never applies.

Scope note: the link-mode "filesystem fallback requires recheck" branch falls through to this same regular add path, so it's covered too.

Related discussion: https://github.com/autobrr/qui/discussions/2433

## How has this been tested?

- Extended `TestProcessCrossSeedCandidate_DivergentCrossCategoryPinsSavePath` to assert `useDownloadPath=false` is set alongside the pinned `savepath` (autoTMM off).
- `go test ./internal/services/crossseed/` — full package passes; `gofmt` and `go vet` clean.
- Real-world: on an instance with a temp/incomplete download path on a different filesystem from the link tree, exact-match cross-seeds were stalling at 0% and hitting the 1-hour "Recheck resume absolute timeout"; repointing the incomplete path to the link dir made the recheck reach 100% immediately, which is what this change does at add time. (Tracker/torrent names and paths omitted per the template.)

N/A — no UI changes.

## Checklist

- [x] My PR title follows the Conventional Commits format
- [x] N/A — no database schema change (no migrations needed)

## AI disclosure

Yes — this was developed with AI assistance (Claude, Opus 4.8). The root-cause investigation, the code change, and the added test assertion were AI-assisted; roughly all of the diff was AI-drafted and then reviewed. I'm not an experienced Go developer, so please review it closely — especially whether `useDownloadPath=false` is the right lever here versus, say, `SetLocation` before recheck, and whether the autoTMM-on path (which I deliberately left alone, matching the manual-add convention) also deserves handling.

## Summary by CodeRabbit

* **Tests**
  * Improved coverage for link-mode fallback behavior by verifying that the download path setting is correctly disabled when adding a torrent.

### DIFF
--- a/internal/services/crossseed/category_divergence_test.go
+++ b/internal/services/crossseed/category_divergence_test.go
@@ -128,4 +128,8 @@ func TestProcessCrossSeedCandidate_DivergentCrossCategoryPinsSavePath(t *testing
 	// location so qBittorrent reuses the existing files instead of re-downloading.
 	require.Equal(t, "false", base.addedOptions["autoTMM"], "autoTMM must be disabled when the cross category save path diverges")
 	require.Equal(t, "/downloads/tv/Show.S01E01", base.addedOptions["savepath"])
+	// ...and the temp/incomplete download path is disabled for the add, so the in-place
+	// recheck reads the pinned savepath instead of the instance's (empty) temp dir. Without
+	// this, an instance with "Keep incomplete torrents in" set would stall the cross-seed at 0%.
+	require.Equal(t, "false", base.addedOptions["useDownloadPath"], "explicit savepath must disable the temp/incomplete download path so the recheck reads the pinned dir")
 }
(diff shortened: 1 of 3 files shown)


statelycurmudgeon:
Note that this refers to this discussion: https://github.com/autobrr/qui/discussions/2433

s0up4200 (approved):
Thanks for the fix!

PR #2452, opened 2026-08-26
fix(search): require an explicit search type

s0up4200:
## Description

Remove the unreliable Auto option from indexer search and default to Movies. Every Search-page request now carries explicit Torznab categories, and recent searches without a recognizable category fall back to Movies.

Update the localized page copy and search-type tests to match the explicit choices.

## How has this been tested?

- Built and started `qui` with isolated configuration and data.
- Opened `/search` and confirmed Movies is selected by default, Auto is absent, and the page reports no browser console errors or framework error overlay.

Not included. The live check used an empty isolated instance with synthetic credentials.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Search now uses Movies as the default search type.
  * Search type selection is consistent across supported languages.
  * Removed the automatic search category option.

* **Bug Fixes**
  * Recent searches and suggestions now correctly fall back to Movies when no type is specified.
  * Search category handling is more consistent for all available search types.

* **Documentation**
  * Updated search-page descriptions and localized text to reflect manual search-type selection.


bitingsock:
May I ask why it has to default to a category at all? Torznab does not require it and many trackers, while supporting torznab, do not correctly categorize their torrents, use custom categories, or sometimes do not categorize at all. A category-less search is often the best way to get results for some things.

PR #2456, opened 2026-08-26
fix(database): truncate cross_seed_feed_items.last_seen_at to day precision

ColinHebert:
## Description

`MarkFeedItem` rewrote `last_seen_at` with a fresh, millisecond-precision timestamp on every single feed poll via `ON CONFLICT ... DO UPDATE`. Because `last_seen_at` carries an index (`idx_cross_seed_feed_items_last_seen`), and the column changes on every write, Postgres cannot use HOT (Heap-Only Tuple) updates, forcing full index maintenance for every poll.

`PruneFeedItems` only reads `last_seen_at` for a coarse 30-day cutoff — it doesn't need millisecond precision. Truncating the value to day precision in Go (dialect-neutral, no SQL-side logic) means repeated same-day polls write an identical value, so only the first poll of a new day actually touches the index. The index itself is kept (unlike the alternative of just dropping it), since it's genuinely used by `PruneFeedItems`.

While adding a regression test for this, I found a SQLite-only trigger, `cross_seed_feed_items_touch`, that fires `AFTER UPDATE` and unconditionally overwrites `last_seen_at` with `CURRENT_TIMESTAMP` — duplicating what the upsert's own `SET last_seen_at = excluded.last_seen_at` already does, and silently discarding the truncated value on SQLite. Postgres never had this trigger. It's dropped here since it's pure redundant overhead (an extra full `UPDATE` per write) and directly undermines the fix on SQLite.

Addresses discussion #2375.

## How has this been tested?

- `go build ./...`, `go vet ./...`, `gofmt -l .` all clean
- `go test -race -count=1 ./...` passes (verified in a Go 1.27 container since no local toolchain was available)
- Added `TestCrossSeedStore_MarkFeedItemTruncatesLastSeenToDayPrecision`, which caught the SQLite trigger issue during development: it asserts a same-day poll leaves the stored value unchanged and a next-day poll advances it

## Checklist

## AI disclosure

This PR was written by Claude Code (Sonnet 5), based on investigation of the linked discussion and the codebase, with human review and verification (build/vet/test run in a container) before submission. The SQLite trigger issue was caught by a test written as part of this change, not anticipated by the original discussion.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed feed item timestamps being overwritten during updates.
  * Feed item timestamps now remain consistent for polls on the same day and advance correctly on subsequent days.
  * Improved completion handling for partial link-mode cross-seed additions.
  * Webhook notifications now more accurately report successful and failed cross-seed applications.
  * Improved completion search error handling and logging.
  * Improved database migration compatibility across existing installations.


ColinHebert:
An alternative here would be the same approach as #2455: just drop `idx_cross_seed_feed_items_last_seen` and accept a seq scan.

I went with truncation instead of dropping the index because, unlike `torrent_files_cache`/`torrent_files_sync` in #2455, this index isn't dead — `PruneFeedItems` actually filters on `last_seen_at < cutoff`. Dropping it would trade write amplification for a seq scan on every automation run's prune step, with no ceiling on the table size that scan runs against. Truncating to day precision removes the write amplification while keeping the indexed lookup for pruning.

Note these aren't a "why not both" situation — truncation only pays off because the index exists to protect; if we drop the index there's nothing left for HOT-eligibility to save, so it'd just be dead code. Happy to switch to dropping the index instead if that's the preferred tradeoff — it's a smaller diff and matches #2455's pattern.

s0up4200 (approved):

PR #2457, opened 2026-08-26
fix(jackett): schedule opportunistic cleanup of torznab indexer latency

ColinHebert:
## Description

`CleanupOldLatency` existed on `TorznabIndexerStore` and was exercised by two unit tests, but had no production caller — `RecordLatency` inserts one row per Torznab search with no corresponding cleanup, so `torznab_indexer_latency` grew without bound.

This wires `CleanupOldLatency` up using the exact interval-gated, opportunistic-cleanup pattern this service already uses for its search cache (`maybeScheduleSearchCacheCleanup`) and torrent cache (`maybeScheduleTorrentCacheCleanup`): a mutex-guarded "next cleanup" timestamp, checked and advanced synchronously, with the actual delete running in a background goroutine. It's triggered after each recorded search latency, gated to once per 6 hours (matching the existing caches' interval), retaining the last 14 days — double the 7-day window `torznab_indexer_latency_stats` needs for its rolling stats.

Addresses discussion #2376.

## How has this been tested?

- `go build ./...`, `go vet ./...`, `gofmt -l .` all clean
- `go test -race -count=1 ./...` passes (verified in a Go 1.27 container since no local toolchain was available)
- Added `TestMaybeScheduleLatencyCleanupTriggersAndDebounces`, asserting the cleanup fires once and is debounced on a second call within the interval
- Added `CleanupOldLatency` to the `IndexerStore` interface, which required updating three test doubles (`mockTorznabIndexerStore`, `failingDirScanIndexerStore`, `failingEnabledIndexerStore`) to keep them compiling against the interface

## Checklist

- [x] If this changes the database schema, I have added migrations for both SQLite and PostgreSQL (n/a — no schema change in this PR)

## AI disclosure

This PR was written by Claude Code (Sonnet 5), based on investigation of the linked discussion and the codebase, with human review and verification (build/vet/test run in a container) before submission.

## Summary by CodeRabbit

* **New Features**
  * Added automatic cleanup of outdated indexer latency records.
  * Cleanup runs periodically in the background, removing records older than 14 days.
  * Cleanup is scheduled after searches and limited to prevent unnecessary repeated work.

* **Bug Fixes**
  * Improved maintenance of latency data to keep performance information current.
  * Reduced redundant cleanup activity while preserving regular data housekeeping.

### DIFF
--- a/internal/services/crossseed/service_search_test.go
+++ b/internal/services/crossseed/service_search_test.go
@@ -118,6 +118,10 @@ func (s *failingEnabledIndexerStore) RecordLatency(context.Context, int, string,
 	return nil
 }
 
+func (s *failingEnabledIndexerStore) CleanupOldLatency(context.Context, time.Duration) (int64, error) {
+	return 0, nil
+}
+
 func (s *failingEnabledIndexerStore) RecordError(context.Context, int, string, string) error {
 	return nil
 }
--- a/internal/services/dirscan/service_indexer_probe_test.go
+++ b/internal/services/dirscan/service_indexer_probe_test.go
@@ -64,6 +64,10 @@ func (s *failingDirScanIndexerStore) RecordLatency(context.Context, int, string,
 	return nil
 }
 
+func (s *failingDirScanIndexerStore) CleanupOldLatency(context.Context, time.Duration) (int64, error) {
+	return 0, nil
+}
+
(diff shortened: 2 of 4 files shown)


s0up4200 (approved):
thanks!

PR #2503, opened 2026-08-31
fix(web): avoid canvas readback for dynamic favicon

heathlarsen:
## Description

Firefox intercepts canvas readback when fingerprinting resistance is enabled because it can expose details such as GPU, drivers, fonts, and rendering behavior. Because qui generates the dynamic favicon by drawing SVG into a canvas and exporting it with `toDataURL("image/png")`, Firefox will return protected canvas output and resulting in a broken favicon.

This change keeps the existing canvas-generated PNG favicon and adds a dynamic SVG favicon candidate. Firefox with strict fingerprinting protections can use the SVG favicon path, avoiding corrupted canvas readback output, while browsers without SVG favicon support still have the PNG candidate.

SVG favicon support reference (but PNG fallback should cover missing support): https://caniuse.com/link-icon-svg

## How has this been tested?

Built and run the Docker image.

Verified in Firefox with strict privacy/fingerprinting protections enabled that the browser tab icon renders correctly instead of as garbled pixels.

Verified that changing themes still updates the favicon colour.

**FIREFOX:**
Before - with fingerprinting disabled, does not support Canvas producing PNG:
image

After - Is fixed by this change as it will use the SVG icon:
Minimal Theme:
image
autobrr Theme:
image

**CHROME:**
Before - was working before this change:
image

After - still working after this change:
Minimal Theme:
image
autobrr Theme:
image

## Checklist

- [x] My PR title follows the Conventional Commits (https://www.conventionalcommits.org/en/v1.0.0/#summary) format (it becomes the squashed commit message)

## AI disclosure

Yes. Codex was used to diagnose the Firefox canvas/fingerprint interaction and draft the small favicon implementation change. The code change is minimal and was reviewed/tested manually by a professional software developer.

## Bug Fixes
- Improved dynamic favicon updates for faster, more reliable display.
- Preserved SVG favicon quality while continuing to support PNG rendering.
- Improved favicon handling across browsers and display contexts.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -2,7 +2,7 @@
 <html lang="en" class="dark">
   <head>
     <meta charset="UTF-8" />
-    <link rel="icon" type="image/x-icon" href="data:," />
+    <link rel="icon" type="image/x-icon" href="data:," data-dynamic-favicon="svg" />
     <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, viewport-fit=cover" />
     <meta name="apple-mobile-web-app-capable" content="yes" />
--- a/web/src/hooks/useDynamicFavicon.ts
+++ b/web/src/hooks/useDynamicFavicon.ts
@@ -22,8 +22,6 @@ export function useDynamicFavicon() {
       canvas.height = size
       const ctx = canvas.getContext("2d")
 
-      if (!ctx) return
-
       // Get the dark mode primary color from the current theme
       const currentTheme = getThemeById(theme)
       if (!currentTheme) return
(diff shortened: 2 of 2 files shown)


heathlarsen:
@s0up4200 just FYI my original commit had me fully replacing the PNG with a SVG. I force-pushed to overwrite this with a version that produces both. This is because of it might've resulted in a loss of compatibility with older browsers. If you prefer the SVG only option I have the changes stashed.

s0up4200:
> @s0up4200 just FYI my original commit had me fully replacing the PNG with a SVG. I force-pushed to overwrite this with a version that produces both. This is because of it might've resulted in a loss of compatibility with older browsers. If you prefer the SVG only option I have the changes stashed.

Should be good as is! Thanks for the PR!

I pushed a small cleanup to your branch btw, removed a dead cancelled check and tagged the index.html placeholder

s0up4200 (approved):

PR #2558, opened 2026-09-04
fix(crossseed): resume fast verification rechecks that finish between polls

### LINKED ISSUE #2554: Verification recheck guard misses fast rechecks, leaving 100%-complete manual matches paused
> *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** The verification recheck guard in the recheck resume worker never observes the checking state for torrents whose recheck completes between polls, causing 100%-complete manual cross-seed matches to stay paused until the 60-minute absolute timeout.

**Current behavior:**
When a cross-seed torrent is added via manual match (or title rescue / relaxed season-episode-group), `candidateRequiresVerification` returns true, and the torrent is queued with `verificationRequired: true`. The recheck resume worker polls qBittorrent every `recheckPollInterval` (3 s) looking for `checkingUp` or `checkingDl` to set `sawChecking`. For torrents with 100% data overlap, qBittorrent finishes the recheck in well under 3 seconds — the torrent transitions checking → paused between two polls. Because `sawChecking` is never set, the guard at `pendingResumeSatisfied` keeps returning `true` ("keep waiting"), and the torrent sits in the queue for the full `recheckAbsoluteTimeout` (60 min). At timeout the entry is removed with the warning "Verification recheck transition was not observed, torrent left paused for manual review" and the torrent is never resumed.

**Desired behavior:**
A verification-required torrent whose recheck completes too fast for polling to observe should still be auto-resumed when the recheck result is 100% (progress ≥ 1.0, amountLeft ≤ 0). The safety property — never resume an unverified torrent — must be preserved: the torrent was explicitly rechecked by qui before entering the queue, so a completed result at 100% after a reasonable minimum elapsed time is sufficient evidence that the recheck ran.

**Key interfaces:**
- `pendingResume.verificationRequired` / `pendingResume.sawChecking`: the guard pair. `sawChecking` is only set when a poll observes `checkingUp` or `checkingDl`, which fast rechecks skip.
- `pendingResumeSatisfied()`: the decision function. When `verificationRequired && !sawChecking` it unconditionally returns `true` (keep waiting), with no fallback for a completed-at-100% torrent.
- `recheckPollInterval` (3 s): the polling cadence that defines the observation window.
- `recheckAbsoluteTimeout` (60 min): the timeout that eventually abandons the entry.
- `candidateRequiresVerification()`: returns true for manual matches (`req.ManualTargetHash != ""`), title rescues, and relaxed season/episode/group search decisions.

**Acceptance criteria:**
- [ ] A verification-required torrent that reaches 100% progress (progress ≥ 1.0, amountLeft ≤ 0) in paused/stopped state is auto-resumed even when no checking state was observed during polling.
- [ ] The guard still fails closed when the torrent is not at 100% and no checking state was observed (no regression on the safety property).
- [ ] The `monitorOnly` path (title rescue monitor) is unaffected — it must still observe or complete without resuming.
- [ ] Existing tests in `relaxed_structure_recheck_test.go` continue to pass.
- [ ] A new test covers the fast-recheck scenario: verification-required entry, no checking state observed, torrent at 100% → resumed.

**Out of scope:**
- Changing the polling interval or switching to an event-driven recheck notification.
- Modifying `candidateRequiresVerification` criteria.
- Changing the absolute timeout duration.

---

From Discord: https://discord.com/channels/881212911849209957/881967548143403058/1545382421736722462

<details>
<summary>Discord transcript</summary>

**skomerko** (10:38 UTC):
I added 3 torrents with 100% overlap (manual cross-seed), but none of them automatically resumed. In each case, qb logs show: added → resumed → stopped. Auto-resume is on. <:WheresMocca:1001686010868027463>  Noticed one warning in Qui logs: "Verification recheck transition was not observed, torrent left paused for manual review", "elapsed": 3602301.768231

**Qui-Gon** (11:46 UTC):
> replying to **skomerko**: I added 3 torrents with 100% overlap (manual cross-seed), but none of them automatically resumed. In each case, qb logs
Triage failed: no usable result from the model. Nothing was filed.

</details>

thejdubb02:
## Description

A manual cross-seed match is queued with `verificationRequired`, and the recheck resume worker only sets `sawChecking` when a poll observes `checkingUp`/`checkingDl`. A 100%-overlap recheck can finish in under the 3s `recheckPollInterval`, so the checking state is never observed, the guard in `processPendingRecheckResume` keeps returning "keep waiting", and the torrent sits paused until the 60-minute `recheckAbsoluteTimeout` ("Verification recheck transition was not observed, torrent left paused for manual review"). It is never resumed.

This narrows that guard. A non-monitor verification entry that is settled at 100% (progress >= 1, amountLeft <= 0, in a paused/stopped state) once `recheckFastCompleteMinElapsed` has passed since queuing falls through to the existing resume path, which still requires `recheckResumeStablePolls` stable polls before it acts. `monitorOnly` still never resumes, and a sub-100% result still fails closed.

Against the acceptance criteria: a fast-recheck 100% resumes; sub-100% with no checking still fails closed; `monitorOnly` is unaffected; the `relaxed_structure_recheck_test.go` tests still pass; and a new test covers the fast-recheck scenario.

Two things worth your call:

- The elapsed threshold. I used `recheckFastCompleteMinElapsed = 30s`. The brief asks for "a reasonable minimum elapsed time", and the value is a judgment about how long a recheck qui issued needs to have run. qBittorrent reports an optimistic 100% before a recheck runs, so this delay is what separates "recheck completed fast" from "recheck has not started". If a torrent can sit paused at an optimistic 100% for longer than this on your setups (a long recheck queue), a larger value is safer. Happy to change it, or to key it off a different signal if you have one.
- I could not field-test this against a live qBittorrent (no instance with a fast-recheck cross-seed scenario on hand). The closest check is the unit tests, which drive `processPendingRecheckResume` through the exact poll sequence.

Fixes #2554

## How has this been tested?

`go test -race -count=1` on `internal/services/crossseed` (the new tests plus the existing verification and recheck tests pass), `go build`, `go vet`, `gofmt`, and `golangci-lint` on the package. The season-pack link tests (`TestApplySeasonPackWebhook_*`) fail in my sandbox because it has no reflink/hardlink support; they fail identically on an unmodified checkout, so they are environmental and unrelated to this change.

## Checklist

- [x] If this changes the database schema, I have added migrations for both SQLite and PostgreSQL (not applicable, no schema change)

## AI disclosure

Yes. This was written by an AI coding agent (Anthropic Claude, Opus 4.8) working from the ready-for-agent brief: it traced the recheck state machine against the source, wrote the change and the tests, and verified them locally (build, vet, gofmt, golangci-lint, go test -race). Essentially the whole diff is AI-authored and then verified as above before submission.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved automatic resumption after fast verification while preserving stability and timing checks.
  * Prevented interrupted verification states from triggering resumption or retaining outdated confirmations after a restart.
  * Ensured incomplete, ambiguous, or unsettled verification states do not resume prematurely.
  * Allowed eligible paused or stopped items to resume after fresh verification, confirmation, and the required waiting interval.
  * Kept items paused when verification remains in progress, even after the fast-completion interval elapses.

### DIFF
--- a/internal/services/crossseed/recheck_resume_test.go
+++ b/internal/services/crossseed/recheck_resume_test.go
@@ -878,85 +878,64 @@ func TestProcessPendingTitleRescueMonitorWaitsForFullProgress(t *testing.T) {
 	require.Empty(t, sync.bulkActions)
 }
 
-func TestProcessPendingRecheckResumeResumesFastVerificationRecheck(t *testing.T) {
+func TestProcessPendingRecheckResumeFastVerificationRecheck(t *testing.T) {
 	t.Parallel()
 
 	// A 100%-overlap recheck can finish between polls, so checking is never observed.
-	// Before the minimum elapsed time the entry keeps waiting; after it, a settled 100%
-	// must auto-resume instead of waiting out the absolute timeout (issue #2554).
+	// The entry must wait out the minimum elapsed time, resume a settled 100% after it,
+	// then reach the confirm-and-drop exit rather than the absolute timeout (issue #2554).
 	sync := &recheckResumeSyncManager{}
 	service := &Service{
 		syncManager:      sync,
 		recheckResumeCtx: context.Background(),
 	}
 	budget := int64(0)
-	settled := qbt.Torrent{
-		Hash:       "hash1",
-		Progress:   1,
-		AmountLeft: 0,
-		State:      qbt.TorrentStatePausedUp,
+	newPending := func(addedAt time.Time) *pendingResume {
+		return &pendingResume{
+			instanceID:           1,
+			hash:                 "hash1",
+			budgetBytes:          &budget,
+			verificationRequired: true,
+			addedAt:              addedAt,
+		}
 	}
+	settled := qbt.Torrent{Hash: "hash1", Progress: 1, AmountLeft: 0, State: qbt.TorrentStatePausedUp}
+	elapsedAgo := time.Now().Add(-2 * recheckFastCompleteMinElapsed)
 
 	// Too soon after queuing: the recheck may not have run, so stay paused.
-	tooSoon := &pendingResume{
-		instanceID:           1,
-		hash:                 "hash1",
-		budgetBytes:          &budget,
-		verificationRequired: true,
-		addedAt:              time.Now(),
-	}
-	keep := service.processPendingRecheckResume(1, "hash1", tooSoon, settled)
-	require.True(t, keep)
+	tooSoon := newPending(time.Now())
+	require.True(t, service.processPendingRecheckResume(1, "hash1", tooSoon, settled))
 	require.Zero(t, tooSoon.resumeAttempts, "a 100% result before the minimum elapsed time must not resume")
 	require.Empty(t, sync.bulkActions)
 
-	// Enough time has passed for the recheck to have run: resume after stable polls.
-	elapsed := &pendingResume{
-		instanceID:           1,
-		hash:                 "hash1",
-		budgetBytes:          &budget,
-		verificationRequired: true,
-		addedAt:              time.Now().Add(-2 * recheckFastCompleteMinElapsed),
-	}
-	keep = service.processPendingRecheckResume(1, "hash1", elapsed, settled)
-	require.True(t, keep)
+	// Below 100% after the elapsed gate: still fails closed, no resume.
+	belowFull := newPending(elapsedAgo)
+	underFull := settled
+	underFull.Progress = 0.99
+	require.True(t, service.processPendingRecheckResume(1, "hash1", belowFull, underFull))
+	require.Zero(t, belowFull.resumeAttempts, "an unverified sub-100% torrent stays queued, not resumed")
+	require.Empty(t, sync.bulkActions)
+
+	// Settled 100% after the elapsed gate: resume after the stable polls.
+	elapsed := newPending(elapsedAgo)
+	require.True(t, service.processPendingRecheckResume(1, "hash1", elapsed, settled))
 	require.Zero(t, elapsed.resumeAttempts, "the unobserved-checking path still waits for stable polls")
 
-	keep = service.processPendingRecheckResume(1, "hash1", elapsed, settled)
-	require.True(t, keep, "the worker keeps the entry until resume is confirmed")
+	require.True(t, service.processPendingRecheckResume(1, "hash1", elapsed, settled),
+		"the worker keeps the entry until resume is confirmed")
 	require.Equal(t, 1, elapsed.resumeAttempts, "a settled 100% after the minimum elapsed time resumes")
+	require.True(t, elapsed.awaitingResumeConfirmation)
 	require.Equal(t, []string{"resume:hash1"}, sync.bulkActions)
-}
 
-func TestProcessPendingRecheckResumeFastPathStaysClosedBelowFull(t *testing.T) {
-	t.Parallel()
-
-	// The fast-recheck fallback only trusts a 100% result. Below 100% with no checking
-	// observed it must still fail closed, even after the minimum elapsed time (no
-	// regression on the safety property, issue #2554).
-	sync := &recheckResumeSyncManager{}
-	service := &Service{
-		syncManager:      sync,
-		recheckResumeCtx: context.Background(),
-	}
-	budget := int64(0)
-	pending := &pendingResume{
-		instanceID:           1,
-		hash:                 "hash1",
-		budgetBytes:          &budget,
-		verificationRequired: true,
-		addedAt:              time.Now().Add(-2 * recheckFastCompleteMinElapsed),
-	}
-
-	keep := service.processPendingRecheckResume(1, "hash1", pending, qbt.Torrent{
-		Hash:       "hash1",
-		Progress:   0.99,
-		AmountLeft: 0,
-		State:      qbt.TorrentStatePausedUp,
-	})
-	require.True(t, keep, "an unverified sub-100% torrent stays queued, not resumed")
-	require.Zero(t, pending.resumeAttempts)
-	require.Empty(t, sync.bulkActions)
+	// The torrent is now running: the worker must confirm the resume and drop the entry,
+	// not poll it for the full timeout and mislog it as left paused.
+	running := settled
+	running.State = qbt.TorrentStateUploading
+	require.True(t, service.processPendingRecheckResume(1, "hash1", elapsed, running),
+		"kept for one running poll while the resume is confirmed")
+	require.False(t, service.processPendingRecheckResume(1, "hash1", elapsed, running),
+		"a stable running state confirms the resume and drops the entry")
+	require.Equal(t, []string{"resume:hash1"}, sync.bulkActions, "no duplicate resume once running")
 }
 
 func TestProcessPendingRecheckResumeForgivenessRetriesAfterNegativeVerdict(t *testing.T) {
--- a/internal/services/crossseed/recheck_resume_test.go
+++ b/internal/services/crossseed/recheck_resume_test.go
@@ -878,6 +878,87 @@ func TestProcessPendingTitleRescueMonitorWaitsForFullProgress(t *testing.T) {
 	require.Empty(t, sync.bulkActions)
 }
 
+func TestProcessPendingRecheckResumeResumesFastVerificationRecheck(t *testing.T) {
+	t.Parallel()
+
+	// A 100%-overlap recheck can finish between polls, so checking is never observed.
+	// Before the minimum elapsed time the entry keeps waiting; after it, a settled 100%
+	// must auto-resume instead of waiting out the absolute timeout (issue #2554).
+	sync := &recheckResumeSyncManager{}
+	service := &Service{
+		syncManager:      sync,
+		recheckResumeCtx: context.Background(),
+	}
+	budget := int64(0)
+	settled := qbt.Torrent{
+		Hash:       "hash1",
+		Progress:   1,
+		AmountLeft: 0,
+		State:      qbt.TorrentStatePausedUp,
+	}
+
+	// Too soon after queuing: the recheck may not have run, so stay paused.
+	tooSoon := &pendingResume{
+		instanceID:           1,
+		hash:                 "hash1",
+		budgetBytes:          &budget,
+		verificationRequired: true,
+		addedAt:              time.Now(),
+	}
+	keep := service.processPendingRecheckResume(1, "hash1", tooSoon, settled)
+	require.True(t, keep)
+	require.Zero(t, tooSoon.resumeAttempts, "a 100% result before the minimum elapsed time must not resume")
+	require.Empty(t, sync.bulkActions)
+
+	// Enough time has passed for the recheck to have run: resume after stable polls.
+	elapsed := &pendingResume{
+		instanceID:           1,
+		hash:                 "hash1",
+		budgetBytes:          &budget,
+		verificationRequired: true,
+		addedAt:              time.Now().Add(-2 * recheckFastCompleteMinElapsed),
+	}
+	keep = service.processPendingRecheckResume(1, "hash1", elapsed, settled)
+	require.True(t, keep)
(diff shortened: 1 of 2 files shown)


s0up4200 (changes requested):

I reviewed `1f9429a8` against `develop`: 2 files, +97/-2, both in `internal/services/crossseed`.

## Standards

**Hard violations**

1. `service.go` ~495: the `pendingResume.verificationRequired` doc comment still says "A transition missed between polls fails closed and stays paused." The fast path makes that false. The struct doc is the state-machine documentation, and the PR did not update it.
2. `service.go` ~7082: the worker still logs "Verification recheck transition was not observed, torrent left paused for manual review" on `verificationRequired && !sawChecking`. After a fast-path resume the torrent is running, so the message is wrong. The hunk comment "fall through to the normal resume path" overstates it. The confirmation branch is unreachable on that path, see Spec.
3. "Keep diffs minimal in high-churn areas": the block comment inserted mid-const-group forces gofmt to realign `maxMissingFilesResumeAttempts`, a churn line with no meaning. gofmt is clean and there is precedent, so this is minor. A trailing comment, or placement at the end of the group, avoids it.

**Judgement calls**

- Duplicated code: `torrent.Progress >= 1 && torrent.AmountLeft <= 0` is exactly what `pendingResumeSatisfied` returns for `verificationRequired` (line 6594). The memoised `satisfied()` closure two lines above already yields this verdict.
- Code shape: the hunk nests a boolean plus `if !fastRecheckComplete { return true }` inside the existing guard. A single flat condition reads better. The function already has `progress := torrent.Progress`; the hunk uses `torrent.Progress`.
- The const comment "short of the 60m timeout" hardcodes a sibling constant's value. Name `recheckAbsoluteTimeout` instead.

**Tests**

- Placement, naming, and `t.Parallel()` match the neighbours. Both tests hand-build `Service`, `pendingResume`, and `qbt.Torrent` literals, duplicating the fixture in `requireVerificationPendingWaitsForObservedFullCheck` (line 726) and the table in `TestProcessPendingRecheckResumeConfirmationStates`. `FastPathStaysClosedBelowFull` belongs as one row of the first test, not a second function.
- The first test stops at `resumeAttempts == 1` and never polls a running state. That shape hides the defect below.

## Spec

**Premise check.** All three `queueVerificationRecheckResume` call sites sit in the success branch of a synchronous `BulkAction(..., "recheck")`. The spec's safety argument ("torrent was explicitly rechecked by qui before entering the queue") holds for the normal path.

**(a) Missing or partial**

- The PR delivers half of "A verification-required torrent ... is auto-resumed". The worker issues the resume call, but the entry never reaches the confirm-and-drop exit. After `resumePendingRecheck` sets `awaitingResumeConfirmation`, the next poll sees a running state. `isPausedOrStopped` is false, so `fastRecheckComplete` is false, and line 6776 returns `true` before the `awaitingResumeConfirmation` branch runs. `resumeConfirmedPolls` stays 0. The worker polls the hash for 60 minutes, then logs it as "not observed, left paused" while it is uploading. I confirmed this with the probe test below: two running polls after the resume give `keep=true true confirmedPolls=0`. The retry half still works. If qBittorrent stops the torrent again, the paused 100% state re-enters the fast gate, reaches `isPausedOrStopped && satisfied()`, and calls `resumePendingRecheck` again.
- The new test masks this. Its final assertion, "the worker keeps the entry until resume is confirmed", passes on correct and broken code alike, because it never feeds a running-state poll.

The probe test I ran on the branch:

```go
func TestProbeFastPathConfirm(t *testing.T) {
	sync := &recheckResumeSyncManager{}
	service := &Service{syncManager: sync, recheckResumeCtx: context.Background()}
	budget := int64(0)
	settled := qbt.Torrent{Hash: "hash1", Progress: 1, AmountLeft: 0, State: qbt.TorrentStatePausedUp}
	req := &pendingResume{instanceID: 1, hash: "hash1", budgetBytes: &budget, verificationRequired: true,
		addedAt: time.Now().Add(-2 * recheckFastCompleteMinElapsed)}
	require.True(t, service.processPendingRecheckResume(1, "hash1", req, settled))
	require.True(t, service.processPendingRecheckResume(1, "hash1", req, settled))
	require.Equal(t, 1, req.resumeAttempts)
	require.True(t, req.awaitingResumeConfirmation)
	running := settled
	running.State = qbt.TorrentStateUploading
	// two stable running polls should confirm and drop the entry
	keep1 := service.processPendingRecheckResume(1, "hash1", req, running)
	keep2 := service.processPendingRecheckResume(1, "hash1", req, running)
	t.Logf("keep after running polls: %v %v confirmedPolls=%d", keep1, keep2, req.resumeConfirmedPolls)
	require.False(t, keep2, "entry should be dropped once resume is confirmed")
}
```

**(b) Not asked for**

None. `recheckFastCompleteMinElapsed` is the spec's "reasonable minimum elapsed time". It adds a fixed >=30 s plus two stable polls to every fast verification resume. The spec allowed a delay; it did not ask for one this long.

**(c) Implemented but questionable**

- The 30 s gate adds little for the stated safety property. qBittorrent's `forceRecheck` on a stopped torrent zeroes pieces synchronously inside the WebUI request, so once qui's recheck call returns, a paused 100% is a post-check result. Queued-for-checking torrents report `checkingUP/DL`, not paused, so `isPausedOrStopped` already excludes them.
- One reachable hole: a qBittorrent hard restart before it saves the post-recheck fastresume. `checkingResumeData` resets `sawChecking` but not `addedAt`. A torrent that reloads with skip-check pieces (100%, stopped, never hashed) passes the fast gate once 30 s from the original add have elapsed. Narrow, and the gate does not close it.
- A zero-value `addedAt` in a hand-built `pendingResume` opens the fast gate through a huge `time.Since`. Tests only; `queuePendingResume` always stamps it.

The targeted tests in `recheck_resume_test.go` and `relaxed_structure_recheck_test.go` pass on the branch.

---

Standards: 3 hard, 3 judgement, 2 test-shape; worst is the stale doc comment plus the misleading timeout warning. Spec: 2 partial, 0 creep, 3 questionable; worst is that the fast-path resume never confirms, so the entry lingers 60 minutes and logs a false "left paused" warning, and the PR test masks it. Fix shape: move the fast-path check so the `awaitingResumeConfirmation` branch runs first, or set `sawChecking = true` when the fast gate passes.

thejdubb02:
Thanks, this is a great review. You are right about the confirm-and-drop gap, and my test hid it by stopping at the resume call. Fixed and pushed (2d6f99f).

The bug: after the fast path calls resumePendingRecheck, awaitingResumeConfirmation is set, but the next running-state poll re-entered the verificationRequired guard and returned before the awaitingResumeConfirmation branch, so the entry never confirmed and would poll to recheckAbsoluteTimeout and mislog. Fix: the guard is now verificationRequired && !sawChecking && !awaitingResumeConfirmation, so once resume is issued the confirmation branch takes over, confirms on stable running polls, and drops the entry. The test now runs the full lifecycle through two running (uploading) polls and asserts the entry is dropped with no duplicate resume. That also resolves the timeout log (hard `#2`): a resumed entry drops on confirmation well before the timeout, so "transition was not observed, left paused" is now only reached by an entry that genuinely never resumed, where it is accurate.

Other points, all addressed:

- Struct doc (hard `#1`): updated to describe the fast-path fallback instead of "fails closed and stays paused".
- Const churn (hard `#3`): moved recheckFastCompleteMinElapsed below maxMissingFilesResumeAttempts so gofmt keeps the group aligned.
- Duplicated 100% check: now uses the memoised satisfied() closure. I kept it inside the verificationRequired guard rather than lifting it, so satisfied() stays lazy and does not trigger the forgiveness file fetch for budget entries.
- Const comment: names recheckAbsoluteTimeout instead of hardcoding 60m.
- Tests: FastPathStaysClosedBelowFull is folded into the one function as the below-100% case, with the fixtures built by a small local helper.

On the delay (spec b): you are right that 30s plus the two stable polls is on the long side, and I would rather you set it than guess. The elapsed gate does the "reasonable minimum elapsed time" job from the brief, and the stable polls are the existing !sawChecking gate in the resume path. I am happy to lower the constant, or to exempt the fast path from the extra stable-poll wait so the elapsed gate is the only delay. Which would you prefer? I also want to be honest that a settled 100% after the gate is a heuristic, not proof the piece check ran (a recheck still queued behind others is the case where it can be wrong), so the value is a safety-versus-latency call on your side.

Verification: go build, go vet, gofmt, golangci-lint, and go test -race on internal/services/crossseed all pass, including the existing verification and relaxed-structure tests. The season-pack link tests fail in my sandbox for lack of reflink/hardlink support and fail identically on an untouched develop, so they are environmental.

s0up4200 (changes requested):

I reviewed all three commits at `51cbd5df`. Resume confirmation now works. One case still leaves torrents paused.

## Standards

I found no standards violations or unnecessary code. The comments, repeated completion check, and constant placement are fixed.

## Spec

**[P2] Normal recheck initialization permanently disables the fallback.** [service.go:6763](https://github.com/autobrr/qui/blob/51cbd5df4b9c7eb2dbbc08bbec6297692530480f/internal/services/crossseed/service.go#L6763)

The spec requires auto-resume "even when no checking state was observed during polling." Libtorrent also uses `checkingResumeData` during a normal forced recheck, before it starts checking pieces. That state alone does not prove a restart. [Upstream implementation](https://github.com/arvidn/libtorrent/blob/v2.0.11/src/torrent.cpp#L2150-L2236)

I traced this sequence in the source:

1. Qui requests a recheck.
2. A poll observes `checkingResumeData` and sets `recheckInterrupted=true`.
3. Piece checking finishes between polls.
4. Later polls report stopped, progress `1`, and remaining bytes `0`.

The 30-second wait cannot clear that flag, so the torrent stays paused until the 60-minute timeout. The original bug remains for this sequence. Please distinguish normal recheck startup from an interrupted check and test both without weakening the restart protection.

The ordinary fast path now confirms the resume and removes the queue entry. I found no unrelated changes.

The [second-poll test request](https://github.com/autobrr/qui/pull/2558#discussion_r3938764988) is still open. I would not block on it because I found no unsafe resume in the current code.

I traced the code but did not run tests, build the app, or reproduce this live. At review time, test, lint, and release reported `action_required`. CI had not checked this revision. [Test workflow](https://github.com/autobrr/qui/actions/runs/34018681534)

Standards: 0 findings. Spec: 1 finding, P2, valid fast rechecks can still remain paused.

thejdubb02:
Thanks, addressed in 004c99d7.

You are right that a normal forced recheck passes through checkingResumeData before it hashes, so that state alone is not a restart. The 51cbd5df branch flagged every verification-required checkingResumeData poll as interrupted, which is what disabled the fast path. I gated recheckInterrupted on sawChecking: a checkingResumeData poll only counts as an interruption when we already observed a piece check (the continuity break). A checkingResumeData poll with no prior piece check is treated as normal recheck startup, so the fast path fires after the fast-complete delay and the torrent confirms and drops instead of waiting out the 60 minute timeout.

Both new cases are rows in TestProcessPendingRecheckResumeConfirmationStates, with addedAt set past recheckFastCompleteMinElapsed so the elapsed delay is not what is being exercised:

- "fast resume when only resume-data validation is observed": a fast recheck whose only checking state is checkingResumeData resumes, then confirms and drops. This fails without the change (resume is never issued, the entry lingers).
- "interrupted piece check stays paused after fast-complete delay": CheckingUp, then checkingResumeData, then a settled 100% stays paused even past the delay, so the restart protection is unchanged.

On the narrow case you noted in the earlier review (a qBittorrent restart before any piece check that reloads skip-check 100% resume data): this change leaves that as-is, it still passes the fast gate once the delay has elapsed. Closing it means detecting a qBittorrent restart across polls rather than from one torrent's state, which felt out of scope here. Happy to look at it separately if you want it closed.

Verified locally: gofmt and go vet clean, and the crossseed recheck-resume tests pass with -race.

s0up4200 (approved):
Looks good to merge. Tested locally with qBittorrent 5.2.3 in Docker. The complete torrent resumed after the fast recheck. The incomplete torrent stayed stopped, including after a crash and restart.

PR #2589, opened 2026-09-06
fix(database): register the sqlite connection hook at init

ColinHebert:
## Description

`go test -race ./internal/services/orphanscan/` intermittently reported a data race inside the shared test-database helper. The orphan scan code was not the cause; the parallel tests in that package only perturbed timing enough to expose a latent problem in how the SQLite connection hook was registered.

`modernc.org/sqlite` appends to an unsynchronized package-level slice in `RegisterConnectionHook` (`driver.go:237`) and reads that same slice on every `Driver.Open` (`driver.go:212`). qui registered the pragma hook lazily behind a `sync.Once` inside `New`, so the first `New` in a test binary wrote the slice while other tests were already opening connections through their own `sql.Open("sqlite", ...)` calls — several packages do this directly, e.g. `internal/services/orphanscan/delete_test.go`.

This moves registration into the package `init()`. The write then happens during package initialization, before any goroutine exists, so there is nothing left to race with. The `sync.Once` and its call site in `New` are removed; the diff is net −2 lines.

### Incidental fix to `qui db migrate`

`qui db migrate --from-sqlite … --to-postgres …` opens the source database directly via `openSQLiteDB` (`internal/database/sqlite_to_postgres.go:222`) and never calls `New`, so under lazy registration the hook was never registered in that process at all and the connection ran on stock SQLite defaults. Measured on a generated `qui.db` with the same `mode=ro` DSN:

| pragma | before | after |
| --- | --- | --- |
| `busy_timeout` | `0` | `5000` |
| `cache_size` | `-2000` (2 MB) | `-64000` (64 MB) |
| `mmap_size` | `0` | `268435456` |
| `foreign_keys` | `0` | `1` |

That command full-scans 58 tables, so the larger cache and mmap help exactly that workload, and a non-zero `busy_timeout` means it no longer fails immediately with `SQLITE_BUSY` if anything else holds the file. `foreign_keys` is moot on a read-only connection.

### Notes for reviewers

- The race itself was **not** reachable in production. Every prod entry point (`serve`, `create-user`, `change-password`) opens SQLite exactly once at the top of a single cobra command, so there was never a second opener to race against.
- The only production behaviour that changes is the `db migrate` table above.
- On a Postgres-only deployment the hook is now registered but never fires, since no SQLite connection is ever opened.
- Test packages that do a raw `sql.Open("sqlite", ":memory:")` and import `internal/database` now always get the pragmas, where previously it depended on whether some earlier test had called `New`. That is more deterministic; all six such packages were run under `-race` and pass.
- No documentation update needed: no config, CLI surface, or user-facing behaviour changes.

## How has this been tested?

This machine has no local Go toolchain, so everything below ran in a `golang:1.27` container.

**The race, before and after.** A throwaway reproducer put a `database.New` and an unrelated `sql.Open` on two goroutines released together, run as fresh processes so each got one shot at the first registration:

| tree | fresh processes | races |
| --- | --- | --- |
| `develop` | 60 | **11** |
| this branch | 60 | **0** |

The failures carried exactly the reported stack pair — write at `driver.go:237` via `sync.Once.doSlow`, previous read at `driver.go:212` inside `Driver.Open`. The reproducer was temporary and is not part of this PR.

**Stress runs.** `go test -race -count=20 ./internal/services/orphanscan/` → ok, 13.9s. `go test -race -count=20 ./internal/database/` → ok, 1647s (needs `-timeout 40m`; the default 10m timeout will kill it).

**Always-on pragmas break nothing.** `-race -count=1` passes on every package that opens SQLite directly: `dbinterface`, `models`, `services/arr`, `services/automations`, `services/crossseed`. `internal/database` and `internal/services/orphanscan` are covered by the `-count=20` stress runs above.

**Live run.** Built `./cmd/qui`, ran `generate-config` then `serve`: 95 migrations applied, server up, `qui.db-wal` and `-shm` present, confirming the init-registered hook applied `journal_mode = WAL`.

**`db migrate` against a real Postgres 16 container.** `--dry-run` completed clean, 58 tables reported, exit 0. This was the risk case: `isReadOnlyDSN` correctly detects `file:…?mode=ro`, so the hook skips `journal_mode`, `synchronous` and `page_size` rather than attempting a write against a read-only database. The pragma table above was measured with a probe binary on that same database.

**Lint/format.** `make precommit` cannot run on this host (no `go`, `gofmt`, `golangci-lint`, `pnpm`), so its Go components ran in the container instead: `gofmt -l` clean, `go vet` clean, `go fix -diff` clean, `golangci-lint v2.13.0` on `./internal/database/...` → 0 issues. Frontend lint is not applicable; there are no `web/` changes.

## Checklist

## AI disclosure

Yes. Claude Opus 5 via Claude Code did the investigation, the code change, and the verification runs, working from a race report supplied by the maintainer. The change is 16 insertions / 18 deletions in one file and was reviewed by the maintainer before merge.

## Summary by CodeRabbit

* **Refactor**
  * Improved database initialization timing to ensure connection settings are applied consistently.
  * Preserved existing database behavior, including connection hooks, timeout handling, read-only detection, configuration, and error reporting.
  * No changes were made to the public interface or expected end-user functionality.

* **Tests**
  * Added regression coverage confirming database connection settings are applied when connections are opened directly.


ColinHebert:
@coderabbitai generate docstrings

s0up4200 (approved):

PR #2650, opened 2026-09-09
fix(gojackett): stop retrying once the caller's context ends

### LINKED ISSUE #2649: test(crossseed): data race between FilterLogLevels and a leaked jackett search goroutine
CI's `test-postgres` job (full suite under `-race`) failed on PR #1917 with a data race in `internal/services/crossseed`:

```
WARNING: DATA RACE
Read at 0x0000038ee0e0 by goroutine 12090:
  github.com/rs/zerolog/log.Warn()
  github.com/autobrr/qui/internal/services/jackett.(*Service).executeIndexerSearch()
      internal/services/jackett/service.go:2020
  ...
  github.com/autobrr/qui/pkg/gojackett.(*Client).retryDo()
      pkg/gojackett/http.go:139
  ...
  github.com/autobrr/qui/internal/services/jackett.(*searchScheduler).executeTask.func2()
      internal/services/jackett/scheduler.go:709

Previous write at 0x0000038ee0e0 by goroutine 12335:
  github.com/autobrr/qui/internal/services/crossseed.TestMatchEpisodeCandidates_FilterLogLevels.func1()
      internal/services/crossseed/seasonpack_test.go:758
```

`seasonpack_test.go:758` is the test swapping `log.Logger` to capture output. The reader is a jackett search goroutine started by an earlier test in the same package that is still running its retry loop and logs "Failed to search indexer" when it gives up.

Root cause: `pkg/gojackett.(*Client).retryDo` calls `retry.Do` without `retry.Context(ctx)`, so once the caller's context ends the request in flight fails immediately but the loop still runs the remaining attempts with backoff, a few seconds in total. `TestSearchTorrentMatches_CallerDeadlineIsNotPartialSuccess` (500 ms caller deadline against a blocking indexer) is the shape that leaves such a goroutine behind, and `seasonpack_test.go` sorts after `search_*_test.go`, so `FilterLogLevels` runs inside that window.

This is also a production bug: a cancelled or timed-out search keeps retrying an indexer nobody is waiting on.

The retry loop also calls `retry.Delay(time.Second * 3)` inside the closure, where the returned option is discarded, so the intended 3 s delay never applies. Not changed here; noted for whoever touches retry timing next.

Fix in the linked PR: pass `retry.Context(ctx)` to `retry.Do`, with a test that fails in 3.4 s without it and passes in about 100 ms with it.

Only the Postgres job sees this because it is the only PR job that runs with `-race` (`make test-postgres`); the release workflow's PR test step runs without it.

nitrobass24:
## Description

`retry.Do` in `pkg/gojackett` ran with the default background context, so a request whose caller had already timed out still went through all five attempts with backoff. In production that is up to a few seconds of retries against an indexer nobody is waiting on. In the test suite it leaks a search goroutine past its test, and its final "Failed to search indexer" log races the next test that swaps the global logger, which is the data race CI reported in `TestMatchEpisodeCandidates_FilterLogLevels`.

The fix is `retry.Context(ctx)` on the `retry.Do` call. The new `TestRetryDoStopsWhenContextEnds` takes 3.4 s and fails without it, and passes in about 100 ms with it.

Not changed: the `retry.Delay(time.Second * 3)` call inside the retry closure discards its option and has never applied. Left alone so this PR does not change retry timing; it is noted on the issue.

Closes #2649

## Checklist

## Test plan

- [x] `go test -race -count=1 ./pkg/gojackett/ ./internal/services/jackett/` pass; the new test fails without the one-line fix
- [x] `make precommit`, `go build ./...`
- [x] `internal/services/crossseed` under `-race` locally: no data race reported (the partial-pool tests that fail on this macOS host fail identically on develop and are unrelated)

## Summary by CodeRabbit

- **Bug Fixes**
  - Network retry attempts now stop promptly when the caller’s context is canceled or times out, preventing unnecessary continued requests.

### DIFF
--- a/internal/services/crossseed/seasonpack_test.go
+++ b/internal/services/crossseed/seasonpack_test.go
@@ -9,6 +9,7 @@ import (
 	"encoding/base64"
 	"maps"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"testing"
@@ -698,6 +699,15 @@ func TestMatchEpisodeCandidates_ExcludesIncompleteEpisodes(t *testing.T) {
 // candidates. The loop reads every cached episode, so a reason that means "belongs to
 // another show or another pack" must stay off the default level.
 func TestMatchEpisodeCandidates_FilterLogLevels(t *testing.T) {
+	// Run in a separate process so search goroutines cannot race with this test's logger.
+	if os.Getenv("QUI_TEST_FILTER_LOG_LEVELS_CHILD") != "1" {
+		cmd := exec.CommandContext(t.Context(), os.Args[0], "-test.run=^TestMatchEpisodeCandidates_FilterLogLevels$", "-test.timeout=30s")
+		cmd.Env = append(os.Environ(), "QUI_TEST_FILTER_LOG_LEVELS_CHILD=1")
+		output, err := cmd.CombinedOutput()
+		require.NoError(t, err, "%s", output)
+		return
+	}
+
 	parsedPack := rls.ParseString("Cool.Show.S03.1080p.WEB.x264-GRP")
 	packRelease := &parsedPack
 
--- a/pkg/gojackett/http.go
+++ b/pkg/gojackett/http.go
@@ -157,6 +157,7 @@ func (c *Client) retryDo(ctx context.Context, req *http.Request) (*http.Response
 		}),
 		retry.Attempts(5),
 		retry.MaxJitter(time.Second*1),
+		retry.Context(ctx),
 	)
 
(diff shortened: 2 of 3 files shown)


s0up4200:
Added a separate process for `TestMatchEpisodeCandidates_FilterLogLevels`. The retry fix stops the backoff, but a canceled search can still log after its caller returns. The separate process prevents that search from sharing the logger that this test replaces.

I reproduced the race with a temporary 100 ms delay after cancellation. The fix passed 20 runs with the same delay under the race detector, then passed after I removed the delay.

PR #2660, opened 2026-09-10
fix(automations): allow workflows to clear categories

### LINKED ISSUE #2658: Automation category action rejects Uncategorized, so a workflow cannot clear a category
## Agent Brief

**Category:** bug
**Summary:** The automation "Move to category" action offers Uncategorized but refuses to save it, and the rule processor would skip it anyway, so no workflow can clear the category of a torrent.

**Current behavior:**
The category action in the workflow dialog renders a select whose first entry is Uncategorized. The dialog maps that entry to an empty category string, and the form category value starts empty, so Uncategorized is what the select displays the moment the action is added. Saving fails with the "Select a category" toast, because submit validation reads an empty category as "the user picked nothing".

The gate is not only in the UI. `CategoryAction` would carry `Category: ""`, and rule processing applies the category action only when `Category` is non-empty, so a stored rule would silently match nothing. Two smaller defects sit on the same value: the "include affected cross-seeds" switch renders only when the category value is non-empty, and the save confirmation preview interpolates the raw category name, so it would read `to category ""`.

The plumbing underneath is already fine. `SetCategory` on the sync manager hands the category straight to the qBittorrent WebAPI `torrents/setCategory` endpoint, which clears the category of a torrent for an empty value; backup restore already relies on that.

**Desired behavior:**
A workflow action can clear the category of a torrent by targeting Uncategorized.

- Selecting Uncategorized and saving succeeds.
- Validation still catches the real error case: the user added the category action and never picked anything. That needs an unpicked state distinct from a deliberate Uncategorized choice, for example by leaving the form category value undefined until the user picks and showing the select placeholder until then.
- The processor plans the category change when the target is the empty category, gating on `Enabled` rather than on a non-empty name. The existing no-op filter still skips torrents that are already uncategorized.
- Cross-seed options (include affected cross-seeds, skip if a cross-seed sits in given categories) behave the same for Uncategorized as for a named category.
- The confirmation preview names the target with the localized Uncategorized label instead of empty quotes.
- Rules stored before this change with a named category behave exactly as before.

**Key interfaces:**
- `CategoryAction`: `Category == ""` must mean "remove the category", not "action off". `Enabled` stays the only on/off switch.
- Rule processing for the category action: currently gated on a non-empty `Category`; gate on `Enabled` instead.
- `SetCategory` on the sync manager: already clears the category for an empty value, so it is expected to need no change.
- WorkflowDialog form state (`exprCategory`) and its submit validation: must separate "not chosen" from "Uncategorized", including when an existing rule is loaded for editing.

**Acceptance criteria:**
- [ ] Adding a Move to category action, selecting Uncategorized, and saving succeeds with no "Select a category" toast.
- [ ] Adding a Move to category action and saving without touching the select is still rejected with the existing toast.
- [ ] Reopening a saved Uncategorized rule shows Uncategorized selected, and saving again keeps it.
- [ ] A backend test shows the processor plans a category change to the empty category for a torrent that has one, and plans nothing for a torrent that is already uncategorized.
- [ ] Include affected cross-seeds and skip-if-cross-seed-in-categories are reachable and honored when the target is Uncategorized.
- [ ] The save confirmation preview shows the localized Uncategorized label rather than empty quotes.
- [ ] Rules stored with a named category are unaffected.

**Out of scope:**
- The CATEGORY condition in the query builder, which already handles Uncategorized correctly.
- The Set Category dialog used for manual torrent actions.
- How categories themselves are created, renamed, or deleted.

From discussion #2657

s0up4200:
## Description

Allows automation workflows to clear torrent categories by selecting Uncategorized, while still rejecting an untouched category field.

Closes #2658

## How has this been tested?

Ran the production binary with temporary data and a localhost qBittorrent stub. The browser rejected an untouched category field, saved and reopened Uncategorized, exposed cross-seed controls, and displayed Uncategorized in the confirmation. A live dry run planned one category change and skipped the already uncategorized torrent. Automatic approval review blocked activation without dry run, so no category write was sent.

The follow-up also rejected an unselected Category beside Pause on Save, Enable, and Dry-run now; selecting Uncategorized saved both actions. The production workflow list displayed Uncategorized for an empty target and archive for a named target.

## Performance

Compared `e19a6ac1` with `e06c931d` using a temporary Go benchmark: one rule, 10,000 synthetic torrents, three runs per target, Linux amd64, Ryzen 7 8745HS, Go 1.27.1, no race detector. Median named-category processing was 5.632 ms before and 5.615 ms after (-0.3%), with 11.75 MB and 50,100 allocations per pass unchanged.

The latest develop merge (`33b71428`, merged as `d7a37b63`) changes only test infrastructure; the measured paths are identical. The validation follow-up (`5d0032ba`) adds scalar UI checks and skips automatic previews for unselected categories.

The empty target previously skipped action planning: 4.069 ms, 10.87 MB, 50,020 allocations. It now plans the requested changes: 5.475 ms, 11.75 MB, 50,100 allocations, comparable to a named category. The frontend changes scalar form state and labels without adding dependencies, scans, subscriptions, or requests.

Verified the production UI with synthetic torrents. No screenshot attached.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Automation category actions can target **Uncategorized** to remove a torrent’s category.
  - Uncategorized selections are supported when saving, previewing, or running automations.
  - Category actions now require a category selection before proceeding; **Uncategorized** remains a valid selection.
  - Automation overviews display the **Uncategorized** label for torrents without a category.

- **Documentation**
  - Updated user and API documentation to explain category removal and selection behavior.

- **Tests**
  - Added coverage for Uncategorized actions, validation, previews, and rule processing.

### DIFF
--- a/documentation/docs/features/automations.md
+++ b/documentation/docs/features/automations.md
@@ -599,7 +599,7 @@ Quick troubleshooting:
 
 ### Category
 
-Move torrents to a different category.
+Move torrents to a different category. Select **Uncategorized** to remove their category.
 
 Options:
 
--- a/internal/api/handlers/automations.go
+++ b/internal/api/handlers/automations.go
@@ -337,11 +337,6 @@ func (h *AutomationHandler) validatePayload(ctx context.Context, instanceID int,
 	}
 	payload.Conditions.Normalize()
 
-	// Validate category action has a category name
-	if payload.Conditions.Category != nil && payload.Conditions.Category.Enabled && payload.Conditions.Category.Category == "" {
-		return http.StatusBadRequest, "Category action requires a category name", errors.New("category name required")
-	}
-
 	// Validate export to instance action
 	if payload.Conditions.ExportToInstance != nil && payload.Conditions.ExportToInstance.Enabled {
 		if payload.Conditions.ExportToInstance.TargetInstanceID <= 0 {
--- a/internal/api/handlers/automations_test.go
+++ b/internal/api/handlers/automations_test.go
@@ -18,6 +18,27 @@ import (
 	"github.com/autobrr/qui/internal/services/automations"
 )
 
+func TestAutomationValidatePayload_Category(t *testing.T) {
+	for _, category := range []string{"", "archive"} {
+		t.Run("target="+category, func(t *testing.T) {
+			handler := NewAutomationHandler(nil, nil, nil, nil, nil)
+			payload := &AutomationPayload{
+				Name:           "Category rule",
+				TrackerPattern: "*",
+				Conditions: &models.ActionConditions{
+					Category: &models.CategoryAction{Enabled: true, Category: category},
+				},
+			}
+
+			status, message, err := handler.validatePayload(t.Context(), 1, payload)
+			require.NoError(t, err)
+			require.Zero(t, status)
+			require.Empty(t, message)
+			require.Equal(t, category, payload.Conditions.Category.Category)
+		})
+	}
+}
+
 func TestAutomationDryRunNow(t *testing.T) {
 	newRequest := func(body string) *http.Request {
 		req := httptest.NewRequest(http.MethodPost, "/api/instances/1/automations/dry-run", strings.NewReader(body))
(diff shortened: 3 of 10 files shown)


s0up4200:
@codex review

PR #2676, opened 2026-09-12
chore(i18n): remove unreferenced locale blocks

ColinHebert:
## Description

Removes seven locale blocks that no component resolves. They are duplicates or leftovers: translators keep maintaining strings that never render, and the copies have already drifted from the live text they shadow — the orphan scan copy interpolates `{{maxFiles}}` where the live key uses `{{max}}`, so a translator "fixing" it would have been editing a placeholder that is wrong anyway.

| Removed block | Keys | Why it is dead |
| --- | --- | --- |
| `automations:orphanScan` | 37 | The Automations page renders `OrphanScanOverview`, which resolves `instances:preferences.orphanScanOverview` |
| `automations:reannounce` | 30 | Same shape: `ReannounceOverview` resolves `instances:preferences.reannounceOverview` |
| `instances:proxySettings` | 30 | `ProxySettingsForm.tsx` was deleted in #1957; the strings stayed |
| `dashboard:speedLimits` | 5 | The dialog reads `torrents:dialogs.speedLimits` |
| `instances:preferences.workflows.toast` | 5 | Superseded by `preferences.workflowsOverview.toast` (20 keys) |
| `instances:preferences.workflows.deleteDialog` | 4 | Byte-identical to `preferences.workflowsOverview.deleteDialog` |
| `automations:queryBuilder.deleteModes` | 4 | Superseded by `instances:preferences.workflowDialog.delete.*` |

115 key paths per locale across ten locales — 1150 strings, 1379 deleted lines. The 9 added lines are trailing-comma fixups where a removed block was last in its parent; no other line changed.

`dashboard.json` needed different handling: the `en` and `cs` files store `÷ — →` as `\uXXXX` escapes, which a JSON round-trip would have silently unescaped. Those blocks were removed textually instead, asserting per file that the parsed result equals *original minus target* so the rewrite could not touch anything else.

The surviving blocks keep serving the strings these copies shadowed. `preferences.workflows.*` is only partly removed — the parent still holds live keys (`minuteSuffix`, `allTrackers`, `actionCount`).

## How has this been tested?

`pnpm check:i18n` passes (exit 0): "All translation keys resolved", and every locale reports `Errors: 0 (missingKeys: 0, extraKeys: 0, interpolation: 0, htmlTags: 0, emptyStrings: 0, encoding: 0)`. The coverage checkers compare each locale's key set against English, so this also confirms all ten locales dropped the blocks together. `pnpm build` succeeds; 37 tests pass across the preferences, i18n and pages suites.

Live run: built the binary against the fresh bundle, started `qui serve`, and fetched every served JS chunk (3.1 MB). "Proxy Configuration", "Select proxy type for routing connections" and the stale `{{maxFiles}}` placeholder are gone from what the server sends; "Orphan Scan", "Reannounce", "Speed Limits", "Automations" and "Auto-cleanup enabled" all still ship.

Because a deletion is only as good as the proof nothing reads it, the unused claim was verified three ways beyond the i18n checks — `check-i18n-keys.mjs` proves used → exists, not the inverse:

1. **All 115 paths against every string literal in the repo** (14,227 distinct literals, 499 source files, plus Go and static assets). 83 have no trace at all. Of the 32 with a trace, 27 were single common words (`title`, `failed`, `delete`) matching unrelated object properties, and 5 were substrings of longer live keys — `reannounce.title` inside `preferences.reannounce.title`, `speedLimits.title` inside `dialogs.speedLimits.title`, and so on. Exact-key uses: 0.
2. **All 102 non-literal translation call sites, enumerated** — the only way a key could escape step 1. Every template literal carries a static prefix (or a `const`-resolved one such as `` `${s}.${…}` `` where `s = "preferences.workflowsOverview.summary"`), and none matches a removed prefix. Variable keys (`t(ACTION_LABEL_KEYS[action])`, `t(item.titleKey)`) read constant tables whose values are string literals, which step 1 already covered. The manually dispatched `_plural` keys are written `t(cond ? "x_plural" : "x")` — both plain literals; `orphanFilesFound_plural` and `pathsIgnored_plural` appear zero times outside the locale JSON, and i18next resolves the `_plural` sibling from the same namespace and prefix as its base key, which is the live `instances` block.
3. **History and backend.** No non-locale file ever referenced these prefixes except the commits that removed their consumers (#1957, #2634). The Go backend emits no i18n key paths; `reannounce.activity` is an activity `Kind` constant consumed by a `switch`, and the removed block had no `activity` key.

The limit worth stating: this is static analysis, so it cannot rule out a key assembled from two runtime fragments with no static prefix. Step 2 enumerated every non-literal call site and no such site exists.

## Performance

Deleting locale keys can only affect bundle size, so that is what was measured. Paths: `web/src/i18n/locales/{en,zh-CN,zh-TW,fr,de,cs,it,ko,uk,pt-BR}/{automations,instances,dashboard}.json`.

Baseline `c23dc25c` (merge-base with develop) vs `e8175ddf`, both built with `pnpm build` (production, Vite 8 + PWA) in the same container image (node:24) against the same `node_modules`. Measured over the whole `web/dist` tree, gzip level 9.

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| dist total, raw | 10,332,860 B | 10,272,060 B | **−60,800 B (−0.59%)** |
| dist total, gzip | 3,088,957 B | 3,070,331 B | **−18,626 B (−0.60%)** |
| `automations` chunks, gzip | 40,004 B | 26,916 B | −13,088 B |
| `instances` chunks, gzip | 223,334 B | 219,766 B | −3,568 B |
| `index` chunk, gzip | 564,704 B | 563,118 B | −1,586 B |
| `dashboard` chunks, gzip | 24,908 B | 24,522 B | −386 B |

Conclusion: an improvement, no regression. English is eager-loaded in `index`, which is why that chunk shrinks too; the rest comes out of the lazy-loaded per-namespace locale chunks. No runtime code path changed, so there is nothing else to measure.

None — no rendered string changed. Every removed key was unreachable, which is the point of the change.

## Checklist

No `documentation/` update needed: none of these strings ever rendered, so no documented behaviour changes.

## AI disclosure

Yes. Written with Claude Code (Claude Opus 5) — the removals, the verification passes, and this description. Reviewed by me before opening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Localization**
  * Removed outdated translations for reannounce, orphan scanning, proxy settings, speed limits, deletion modes, and workflow deletion messages across supported languages.
  * Retained workflow action-count translations where applicable.


s0up4200:
@codex review

s0up4200 (approved):

PR #2677, opened 2026-09-12
fix(i18n): render the correct plural form for six count strings

ColinHebert:
## Description

Six count strings rendered the wrong plural form. Two different causes, one fix.

**Four keys had no plural forms i18next could resolve.** They carried a base and a `_plural` sibling, and `_plural` belongs to the pre-v4 JSON format: i18next 21.0.0 "introduce[d] new v4 json format (intl plural forms with ordinal support)", and 24.0.0 "remove[d] old i18next JSON formats" with the "Intl API [...] mandatory now and will not fallback anymore" ([CHANGELOG](https://github.com/i18next/i18next/blob/master/CHANGELOG.md)). We are on 26.4.0, where the suffix gets no special handling at all — the installed `node_modules/i18next` dist contains zero occurrences of `_plural` or `compatibilityJSON`. So `t(key, { count: 3 })` fell through to the base: "Only 3 selected categories will be matched." came out as "Only 3 selected **category** will be matched."

| Key | Call site |
| --- | --- |
| `instances:preferences.workflowsOverview.tagActions` | `WorkflowsOverview.tsx:1899` |
| `instances:preferences.completionOverview.selectedCategories` | `CompletionOverview.tsx:428` |
| `instances:preferences.completionOverview.selectedTags` | `CompletionOverview.tsx:442` |
| `instances:preferences.completionOverview.selectedIndexers` | `CompletionOverview.tsx:455` |

**Two keys were dispatched by hand, which hardcodes the English plural rule.** `dashboard:trackerBreakdown.importDialog.conflictCount` and `...toasts.importedCount` were written `t(n !== 1 ? "…_plural" : "…")` in `Dashboard.tsx`. That is correct for two-form languages and harmless for zh/ko, but Czech and Ukrainian were routed through the English rule as well:

| | count 2 | count 21 |
| --- | --- | --- |
| cs, before | `2 konfliktů` — the 5+ genitive, at a `few` count | — |
| uk, before | `2 конфліктів` — should be `конфлікти` (`few`) | `21 конфліктів` — 21 takes `one` in Ukrainian |

Both now go through i18next with CLDR suffixes, and the ternaries are gone.

All six keys are converted in all ten locales, with the forms each coverage checker enforces: `_one`/`_other` for en, de, fr, it, pt-BR; `_other` alone for zh-CN, zh-TW, ko; `_one`/`_other`/`_few`/`_many` for uk. Existing translations were mapped to the category their grammar already was rather than re-translated — for uk that means genitive-plural strings became `_many` and nominative-plural ones `_few`.

**Czech needed a form that did not exist.** i18next resolves cs counts 2-4 to `_few`, which does not fall back to `_other`, so a bare one/other pair renders the raw key string there. Before this change cs at least showed readable (if ungrammatical) text, so shipping one/other only would have been a regression. cs gets `_few` for 2-4 plus a proper genitive `_other` for 0 and 5+ — `5 vybraných tagů bude zahrnuto`, not `5 vybrané tagy budou zahrnuty`.

`web/AGENTS.md` said "Legacy `_plural` keys are manually dispatched and must exist in all locales". After this PR none are dispatched, so that line now states which suffixes each language needs and that `_plural` no longer resolves.

### Not in this PR

`instances:preferences.orphanScanOverview.pathsIgnored` has the same broken `_plural` shape and is **not** fixed by any open PR — it survives in #2675's head and in #2676's head. I left it alone to keep this branch off the files those two are moving, but it needs a follow-up (it converts cleanly against both; I checked). The sibling keys in that block *are* covered: #2675 converts `orphanFilesFound` → `orphanItemsFound_one/_other` and `filesCount` → `itemsCount_one/_other` as part of its files→items rename.

`automations:orphanScan.pathsIgnored` and `orphanScan.orphanFilesFound` have the same shape and are deliberately skipped, because #2676 deletes that whole unreferenced block — the Automations page renders `OrphanScanOverview`, which resolves the `instances` copy instead. Fixing strings that are about to be deleted would have bought nothing.

Both of those PRs merge with this one cleanly: a three-way merge of every file this branch touches, against #2675 (`claude/orphan-cleanup-terminology-7b083c`) and #2676 (`e8175dd`), reports zero conflicts. Any order works.

**The cs and uk strings are worth a native-speaker look.** The existing translations were preserved, but the new `_few`/`_many` forms are mine.

## How has this been tested?

`pnpm check:i18n` passes (exit 0), as do `pnpm build`, `pnpm lint` (0 errors) and 37 tests across the i18n, preferences and pages suites.

Beyond CI, resolution was exercised against real i18next with the shipped locale files — **10 locales × 6 keys × counts {0,1,2,3,4,5,11,21,101}, no raw keys and no wrong-category strings**. Spot checks: en 5 → "Only 5 selected categories will be matched."; cs 1/2/5 → "Jen 1 vybraný tag bude zahrnut." / "Jen 2 vybrané tagy budou zahrnuty." / "Jen 5 vybraných tagů bude zahrnuto."; cs conflicts 1/2/5/21 → `1 konflikt` / `2 konflikty` / `5 konfliktů` / `21 konfliktů`; uk conflicts → `1 конфлікт` / `2 конфлікти` / `5 конфліктів` / `21 конфлікт`.

Live run: built the binary against a fresh bundle, started `qui serve`, and fetched the served assets over HTTP. `conflictCount_one`/`_other`/`_few`/`_many` and `importedCount_*` all ship in the dashboard chunks; the cs `konflikty` form ships; `conflictCount_plural` and `importedCount_plural` return zero hits across every eager chunk. Same check passed for the `instances` keys — the eager English chunk carries `tagActions_one`/`_other` and `selectedCategories_one`/`_other`, and the lazy locale chunks carry `selectedTags_few` (cs) and `selectedIndexers_many` (uk).

Seeing the strings rendered needs a configured qBittorrent instance with workflows or a completion rule, plus a tracker-customization import for the dashboard pair, so the wording in those panels is worth a human look.

## Performance

Locale JSON plus two call sites that lost a ternary. Bundle size is the only metric that can move; the removed conditionals are not on a hot path.

Paths: `web/src/i18n/locales/*/instances.json`, `web/src/i18n/locales/*/dashboard.json`, `web/src/pages/Dashboard.tsx`. Baseline `c23dc25c` (merge-base with develop) vs `f1dc7f2e`, both `pnpm build` (production, Vite + PWA) in the same container image (node:24) against the same `node_modules`, measured over the whole `web/dist` tree.

| Metric | Before | After | Delta |
| --- | --- | --- | --- |
| dist total, raw | 10,332,860 B | 10,333,224 B | **+364 B (+0.0035%)** |
| dist total, gzip (sum of per-file) | 3,092,433 B | 3,092,526 B | **+93 B (+0.003%)** |

An increase, and an accounted-for one: the diff adds 24 strings (cs `_few`, uk `_few`/`_many`) that did not exist before, against two ternaries removed. Locale JSON sources grew by +483 B, and the raw bundle by +364 B — the difference being the deleted conditional code.

Method note: gzip is the **sum of per-file gzip -9**, not gzip of the concatenated tree. Concatenating first makes the number depend on `find` ordering, and content hashes change between builds, which produced a spurious ~2 KB swing on an earlier run. Per-file summing is order-independent. Repeat builds of identical input were byte-identical, so these figures are signal, not noise.

Conclusion: no regression worth acting on. Correct plural rendering in two languages costs 93 gzipped bytes.

None. The change is to count wording inside the Workflows, Completion and tracker-breakdown panels, all of which need a configured instance to render.

## Checklist

Documentation: `web/AGENTS.md` is updated in this PR, as above. No `documentation/docs/` change — that documents product behaviour, not i18n key format, and no documented string changed.

Follow-ups this PR does not take on: `instances:preferences.orphanScanOverview.pathsIgnored` (above), and roughly 20 pre-existing cs plurals that are missing `_few` — `preferences.workflows.actionCount` at count 2 renders the literal string `preferences.workflows.actionCount` on develop today.

## AI disclosure

Yes. Written with Claude Code (Claude Opus 5) — the locale changes, the Czech and Ukrainian plural forms, the verification and this description. Reviewed by me before opening.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pluralized text handling for workflow actions, completion filters, and dashboard import messages.
  * Added language-appropriate plural forms for Czech and Ukrainian, including “few” and “many” cases where applicable.
  * Updated pluralization across German, English, French, Italian, Brazilian Portuguese, Korean, Simplified Chinese, and Traditional Chinese.
  * Improved count-based wording for Ukrainian dashboard and completion messages.

* **Documentation**
  * Updated localization guidance to reflect current pluralization rules.


s0up4200:
@codex review

PR #2727, opened 2026-09-15
feat(sshpool): SSH connection layer, host-key pinning, and credential endpoints

### LINKED ISSUE #2722: feat(sshpool): SSH connection layer, host-key pinning, and credential endpoints
Design: `docs/remote-backend-design.md`, rollout step 3, first slice. Builds on the SSH schema and credential store from #1917. No filesystem operations run over SSH in this PR; the `fsops.Pool` keeps returning `ErrRemoteBackendNotImplemented` for remote mode.

## Scope

**`internal/sshpool`**
- `Dial(ctx, instance)`: `net.JoinHostPort`, dial timeout, key auth from `GetDecryptedSSHKey`. Passphrase keys are already rejected at write time.
- Host-key callback with three outcomes: unpinned (first contact, return the presented key), pinned and matching (the pinned key's algorithms go first in `HostKeyAlgorithms`, the rest stay allowed so a key-type change reports as a mismatch rather than a failed negotiation), pinned and different (typed mismatch error carrying both keys). A pin that fails to decrypt is never "unpinned": `ssh-test` reports `pin_unreadable` with the presented key (or without one when the host is also unreachable; missing credentials still answer 400 first) and no probe; the replace route is the way out, and an endpoint change drops the pin as it always does.
- Capability probe, read-only: open the sftp subsystem and record `statvfs@openssh.com`, `hardlink@openssh.com`, `limits@openssh.com`; open one exec session running `true` to record whether exec is allowed at all (a key restricted to `internal-sftp` still opens the session, prints a refusal and exits non-zero), then a second running `find --version && stat --version` to record whether the userland is GNU. Reflink probing waits for the exec tier PR since it needs a path.
- Every dial and probe honors ctx cancellation (the probe closes the one-shot connection on cancel) and closes the connection when done. The pin routes answer 409 from the row before dialing. No pool, keepalive, or reconnect yet.
- Adds `github.com/pkg/sftp` (needed for the extension list, and by the next PR regardless).

**`internal/models`**
- `ReplaceHostKeyPin(ctx, instanceID, host, port, marshaledKey)`: refuses an unpinned instance; otherwise the mirror of `SetHostKeyPin`.

**API** (all on an existing instance; the AAD binds ciphertext to the row id)
- `PUT /api/instances/{id}/ssh-credentials` `{ host, port, username, privateKey }` → `SetSSHCredentials`; 204. A host or port change drops the pin, as the store already does.
- `DELETE /api/instances/{id}/ssh-credentials` → `ClearSSHCredentials`. Pin is kept.
- `POST /api/instances/{id}/ssh-test`: dials with stored credentials only. Response:
  ```json
  {
    "status": "unpinned" | "pinned" | "mismatch" | "pin_unreadable" | "error",
    "hostKey": "<base64 ssh wire format>",
    "fingerprint": "SHA256:…", "keyType": "ssh-ed25519",
    "pinnedFingerprint": "SHA256:…", "pinnedKeyType": "…",
    "capabilities": { "sftp": true, "statvfs": true, "hardlink": true, "limits": false, "exec": true, "gnuUserland": true },
    "error": "…"
  }
  ```
  The probe runs on first contact and on a matching pin. It never runs on mismatch or on an unreadable pin. `pinned*` fields appear only on mismatch.
- `POST /api/instances/{id}/ssh-host-key` `{ hostKey }`: re-dials, requires the host to present exactly `hostKey`, then `SetHostKeyPin`. 409 when already pinned.
- `POST /api/instances/{id}/ssh-host-key/replace` `{ hostKey }`: same, but `ReplaceHostKeyPin`; 409 when unpinned. The frontend reaches this only from the mismatch screen. No previous key is required: the user accepts the new key the same way a terminal client does.
- `InstanceResponse` gains `sshHost`, `sshPort`, `sshUsername`, `sshHostKeyPinned` (bool), `filesystemMode` (`none` | `local` | `remote`). Never the key material.
- OpenAPI updated for all of the above; `make test-openapi` passes.

Stateless by design: first contact returns the key, confirm echoes it back and the server re-dials to check the host still presents it. Nothing is held in memory between the two calls and a restart between them is harmless.

## Out of scope
- Any `fsops.Backend` implementation, connection pooling, reconnect, keepalive (next PR).
- Reflink probe, exec-based operations (exec tier PR).
- Frontend (step 4).
- Schema changes. None needed.

## Tests
In-process only: an `x/crypto/ssh` server in the test with the `pkg/sftp` server package as the sftp subsystem and a fake exec handler. Cover: first contact returns key and probe; confirm pins; second test reports pinned; mismatch fails closed with both fingerprints and no probe; tampered pin ciphertext reports `pin_unreadable` with the presented key, no probe, and `replace` clears it; the same pin against an unreachable host still reports `pin_unreadable`; key-type change is a mismatch; `ssh-host-key` refuses when pinned and `replace` refuses when unpinned; confirm refuses when the host presents a different key than the one echoed; a `command="internal-sftp",restrict` style server (exec refused) yields `exec: false`; ctx cancellation unblocks a hung dial.

## Field test
Against a real seedbox: `PUT ssh-credentials`, `ssh-test` shows fingerprint and tier, `ssh-host-key` pins it, a second `ssh-test` reports `pinned` with the same tier. Record commands and output in the PR.

## Checks
`make precommit`, `go test -race -count=1 ./internal/sshpool/... ./internal/models/... ./internal/api/...`, `make test-openapi`, `make build`.

nitrobass24:
## Description

First slice of the remote filesystem backend (`docs/remote-backend-design.md`, rollout step 3): the SSH connection layer and the API that lets a user store SSH credentials on an instance, see the host-key fingerprint and the probed capability tier, confirm the pin, and replace it after a mismatch. No filesystem operation runs over SSH yet; `fsops.Pool` still returns `ErrRemoteBackendNotImplemented` for remote mode. The frontend and the SFTP backend follow in their own PRs.

**`internal/sshpool`** (new). One-shot dials with the stored key. The host-key callback has three outcomes: first contact returns the presented key, a matching pin proceeds, a different key is a typed `MismatchError` carrying both keys. A pin that fails to decrypt is never treated as first contact. The pinned key's algorithms go first in `HostKeyAlgorithms` and the rest stay allowed, so a key-type change reports as a mismatch instead of a failed negotiation. A read-only probe records the sftp extensions (`statvfs`, `hardlink`, `limits`), whether exec is allowed, and whether the userland is GNU. One socket deadline bounds the handshake and the probe; a cancelled request closes the connection. Adds `github.com/pkg/sftp`.

**`internal/models`.** `ReplaceHostKeyPin`, the named operation for the confirmed-mismatch path, sharing the compare-and-set write with `SetHostKeyPin`. `SetSSHCredentials` validation errors now carry `ErrInvalidSSHCredentials` so the handler can tell a bad submission from a cipher or database failure.

**API**, all on an existing instance:

| Route | Behaviour |
|---|---|
| `PUT /api/instances/{id}/ssh-credentials` | store host, port, username, private key; 204. A host or port change drops the pin |
| `DELETE /api/instances/{id}/ssh-credentials` | clear credentials, keep the pin; 204 |
| `POST /api/instances/{id}/ssh-test` | dial with stored credentials; 200 with `status` `unpinned`, `pinned`, `mismatch`, `pin_unreadable` or `error`, the presented key, fingerprint, key type, both fingerprints on a mismatch, and capabilities when the key is trusted or unpinned |
| `POST /api/instances/{id}/ssh-host-key` | re-dial, require the host to present exactly the echoed key, pin it; 409 when already pinned |
| `POST /api/instances/{id}/ssh-host-key/replace` | the same over an existing pin; 409 when unpinned |

Confirmation is stateless: the client echoes the key `ssh-test` showed and the server re-dials to check the host still presents it. `InstanceResponse` gains `sshHost`, `sshPort`, `sshUsername`, `sshHostKeyPinned` and `filesystemMode`; the key never leaves the server. OpenAPI covers all of it.

`pin_unreadable` is reported with the presented key and no probe, so a corrupt pin has a way out through the replace route rather than a bare error. When the host is also unreachable the status still wins, without a key. Missing credentials answer 400 first.

Tests run against an in-process SSH server (`internal/testutil/sshtest`) with the pkg/sftp server as the subsystem and a fake exec handler in GNU, BSD, sftp-only and hanging modes. They cover first contact, pinning, mismatch with no session opened, key-type change, tampered pin, both 409 directions, confirm against the wrong key, cancellation during a hung handshake and a hung command, and the deadline on both.

Documentation: none under `documentation/docs/` in this PR. The feature is not reachable from the UI until the frontend PR, which will carry the user docs. `docs/remote-backend-design.md` is updated where the code settled a detail.

Fixes #2722

## How has this been tested?

Field test against a real seedbox is pending and will be recorded here before the PR leaves draft: `PUT ssh-credentials`, `ssh-test` for the fingerprint and tier, `ssh-host-key` to pin, a second `ssh-test` reporting `pinned`, and the fingerprint compared with `ssh-keyscan` out of band.

## Performance

No hot path is touched. The five endpoints are user-initiated, each dials once and closes, and `fsops.Pool` still routes remote mode to an error. The only change on an existing path is five field copies in the instance list response, which is cheaper than the client health lookup already on that path. No measurement needed.

## Checklist

- [x] If this changes the database schema, I have added migrations for both SQLite and PostgreSQL (no schema change)

## AI disclosure

Written with Claude Code (Claude Fable 5.1). Nearly all of the code was AI-written from a spec grilled and approved by hand, then passed through an adversarial review workflow and four rounds of two-axis review, with each fix approved by hand.


nitrobass24:
Closing: commits carried AI attribution trailers that AGENTS.md forbids. Reopened with rewritten history.

PR #2728, opened 2026-09-15
feat(sshpool): SSH connection layer, host-key pinning, and credential endpoints

### LINKED ISSUE #2722: feat(sshpool): SSH connection layer, host-key pinning, and credential endpoints

nitrobass24:
## Description

First slice of the remote filesystem backend (`docs/remote-backend-design.md`, rollout step 3): the SSH connection layer and the API that lets a user store SSH credentials on an instance, see the host-key fingerprint and the probed capability tier, confirm the pin, and replace it after a mismatch. No filesystem operation runs over SSH yet; `fsops.Pool` still returns `ErrRemoteBackendNotImplemented` for remote mode. The frontend and the SFTP backend follow in their own PRs.

**`internal/sshpool`** (new). One-shot dials with the stored key. The host-key callback has three outcomes: first contact returns the presented key, a matching pin proceeds, a different key is a typed `MismatchError` carrying both keys. A pin that fails to decrypt is never treated as first contact. The pinned key's algorithms go first in `HostKeyAlgorithms` and the rest stay allowed, so a key-type change reports as a mismatch instead of a failed negotiation. A read-only probe records the sftp extensions (`statvfs`, `hardlink`, `limits`), whether exec is allowed, and whether the userland is GNU. One 15 s socket deadline bounds the connect, the handshake and the probe together, so a host that answers and then stalls fails the request instead of holding it; probe output is capped at 8 KiB; a probe the connection does not survive, whether the request was cancelled or the deadline fired, is an error and never a partial capability report; a cancelled request closes the connection. Adds `github.com/pkg/sftp`.

**`internal/models`.** `ReplaceHostKeyPin`, the named operation for the confirmed-mismatch path, sharing the compare-and-set write with `SetHostKeyPin`. `SetSSHCredentials` validation errors now carry `ErrInvalidSSHCredentials` so the handler can tell a bad submission from a cipher or database failure.

**API**, all on an existing instance:

| Route | Behaviour |
|---|---|
| `PUT /api/instances/{id}/ssh-credentials` | store host, port, username, private key; 204. A host or port change drops the pin |
| `DELETE /api/instances/{id}/ssh-credentials` | clear credentials, keep the pin; 204 |
| `POST /api/instances/{id}/ssh-test` | dial with stored credentials; 200 with `status` `unpinned`, `pinned`, `mismatch`, `pin_unreadable` or `error`, the presented key, fingerprint, key type, both fingerprints on a mismatch, and capabilities when the key is trusted or unpinned |
| `POST /api/instances/{id}/ssh-host-key` | re-dial, require the host to present exactly the echoed key, pin it; 409 when already pinned |
| `POST /api/instances/{id}/ssh-host-key/replace` | the same over an existing pin; 409 when unpinned |

Confirmation is stateless: the client echoes the key `ssh-test` showed and the server re-dials to check the host still presents it. `InstanceResponse` gains `sshHost`, `sshPort`, `sshUsername`, `sshHostKeyPinned` and `filesystemMode`; the key never leaves the server. OpenAPI covers all of it.

All three dialing routes tell an unreachable host from a stored key qui cannot read: the first carries the transport's own text (`ssh-test` in its `error` field, the pin routes as a 502 with a fixed message), the second is logged and answered with a fixed message (`ssh-test` as `error`, the pin routes as 500).

`pin_unreadable` is reported with the presented key and no probe, so a corrupt pin has a way out through the replace route rather than a bare error. When the host is also unreachable the status still wins, without a key. Missing credentials answer 400 first.

Tests run against an in-process SSH server (`internal/testutil/sshtest`) with the pkg/sftp server as the subsystem and a fake exec handler in GNU, BSD, sftp-only and hanging modes. They cover first contact, pinning, mismatch with no session opened, key-type change, tampered pin, both 409 directions, confirm against the wrong key, cancellation during a hung handshake and a hung command, and the deadline on both.

Documentation: none under `documentation/docs/` in this PR. The feature is not reachable from the UI until the frontend PR, which will carry the user docs. `docs/remote-backend-design.md` is updated where the code settled a detail.

Closes #2722

## How has this been tested?

Against a commercial shared seedbox (the same host as the design doc's Field Validation section), with this branch running in the test qui container and a dedicated passphrase-less ed25519 key authorized there. Host, user and key are not shown; the key never appeared in any response.

```
PUT  /api/instances/3/ssh-credentials            -> 204
POST /api/instances/3/ssh-test                   -> 200
  status=unpinned  keyType=ecdsa-sha2-nistp256
  fingerprint=SHA256:wkYs6InaRJ//aGk6UBdColVkLTWd2aClFNg36eP4Aq0
  capabilities: sftp statvfs hardlink limits exec gnuUserland  (all true)
ssh-keyscan -t ecdsa <host> | ssh-keygen -lf -   -> SHA256:wkYs6InaRJ//aGk6UBdColVkLTWd2aClFNg36eP4Aq0  (matches)
POST /api/instances/3/ssh-host-key {hostKey}     -> 204
POST /api/instances/3/ssh-host-key {hostKey}     -> 409 "ssh host key is already pinned"
POST /api/instances/3/ssh-test                   -> 200  status=pinned, same fingerprint and capabilities
GET  /api/instances                               -> instance 3: sshHostKeyPinned=true filesystemMode=remote, no key material
```

With the same key re-authorized under the recommended `command="internal-sftp",restrict` template:

```
POST /api/instances/3/ssh-test                   -> 200  status=pinned, same fingerprint
  capabilities: sftp statvfs hardlink limits  (true)   exec gnuUserland  (false)
```

So a real OpenSSH forced command reads as the SFTP-only tier, and the pin is unaffected by the key's restrictions.

Credential lifecycle, same host:

```
DELETE /api/instances/3/ssh-credentials          -> 204   GET: sshHostKeyPinned=true filesystemMode=none
POST   /api/instances/3/ssh-test                 -> 400 "SSH credentials are not configured for this instance"
PUT    /api/instances/3/ssh-credentials          -> 204   GET: sshHostKeyPinned=true filesystemMode=remote
POST   /api/instances/3/ssh-test                 -> 200   status=pinned, same fingerprint, no re-confirmation
```

Two things the live host showed that the in-process server cannot: `limits@openssh.com` is advertised (this box is past OpenSSH 8.6 now), and first contact negotiated the host's ECDSA key rather than its ed25519 one, so that is what got pinned; later connects put ECDSA first and report `pinned`, as designed. Mismatch, unreadable pin, and the sftp-only tier are covered by the in-process tests.

## Performance

No hot path is touched. The five endpoints are user-initiated, each dials once and closes, and `fsops.Pool` still routes remote mode to an error. The only change on an existing path is five field copies in the instance list response, which is cheaper than the client health lookup already on that path. No measurement needed.

## Checklist

- [x] If this changes the database schema, I have added migrations for both SQLite and PostgreSQL (no schema change)

## AI disclosure

Written with Claude Code (Claude Fable 5.1). Nearly all of the code was AI-written from a spec grilled and approved by hand, then passed through an adversarial review workflow and four rounds of two-axis review, with each fix approved by hand.

## Summary by CodeRabbit

- **New Features**
  - Added SSH-based remote filesystem support for instances.
  - Added controls to save or remove SSH credentials.
  - Added SSH connectivity testing with host-key details, fingerprints, filesystem capabilities, and distinct error statuses.
  - Added host-key confirmation and replacement workflows.
  - Instance responses now show SSH connection metadata, pin status, and filesystem access mode.
  - Added handling for unreadable pins and changed SSH endpoints.
- **Documentation**
  - Updated API and architecture documentation for SSH endpoints, validation, capabilities, and host-key behavior.


nitrobass24:
Field tested on my live instance.

s0up4200 (changes requested):

## Standards

I found no hard breach of a documented standard. These are judgement calls:

1. `docs/architecture.md`: the module map does not name the new `internal/sshpool` package. One line that names the dialer closes the gap.
2. `internal/web/swagger/openapi.yaml:6928`: one em dash in the `hostKey` description. The file on `develop` has none.
3. `internal/sshpool/sshpool_test.go:294-297`: two comment paragraphs sit over one test, and the second one replaces the first. Drop the first.
4. `internal/api/handlers/instances.go:271-276`, `327-332`, `374-379`: the same five SSH fields appear in all three response builders. The file already does this for `SortOrder` and `IsActive`, so I ask for no change here.
5. `internal/api/handlers/instances_ssh.go:167` and `internal/models/instance_ssh.go:206`: the `replace bool` flag goes through two layers, and each layer switches on it. Two states read fine. A third state turns this into a repeated switch.
6. `internal/api/handlers/instances_ssh.go:79`: the 400 body reads "invalid ssh credentials: ssh port 70000 out of range". The sentinel is right, but the prefix is noise for the client.
7. `internal/api/handlers/instances_ssh.go:211`: the confirm route returns the raw error text in a 502 body. A cipher fault from `GetDecryptedSSHKey` shows "read ssh key: … message authentication failed", which the credentials route hides on purpose (`TestUpdateCredentialsReportsStoreFailureAsServerError`).
8. `TestReplaceHostKeyPinRefusesAnUnconfirmedEndpoint`: the fixture host is `stale.example.com`. The rule asks for `example.invalid` or a `.test` name. The test never dials it.

## Spec

No requirement from #2722 is missing. Every test case in the issue has a test in the diff. Three places look wrong:

1. `internal/sshpool/sshpool.go:161` with `internal/sshpool/probe.go:26-60`: a cancelled probe returns no error and a partly filled capability set. `TestCancellationUnblocksProbe` asserts `Exec:false` on a host that allows exec. This does no harm while the handler discards the response. It misleads as soon as something caches the tier.
2. `internal/sshpool/sshpool.go:145-151`: with an unreadable pin, any failure before the handshake other than `ErrSSHKeyNotConfigured` reports `pin_unreadable`. A stored private key that does not parse (`internal/sshpool/sshpool.go:225-228`) lands there too. The spec gives that precedence to an unreachable host only.
3. `internal/api/handlers/instances_ssh.go:211`: every confirm error that is not a mismatch becomes a 502, and the OpenAPI text says 502 means "The host could not be reached". A malformed stored key or a cancelled request also gets 502.

Standards: 8 judgement calls, worst is the cipher fault text in the 502 body. Spec: 0 missing, 3 look wrong, worst is the cancelled probe that returns partial capabilities as success.

nitrobass24:
Thanks, all addressed in 33e4cc35, 97238094 and 871fedc2.

**Standards**
1. `docs/architecture.md` names `internal/sshpool`.
2. Em dash removed from the OpenAPI `hostKey` description.
3. First comment paragraph over `TestCancellationUnblocksProbe` dropped.
4, 5. Left as you asked.
6. The credentials 400 body no longer carries the sentinel prefix.
7. Confirm and replace no longer echo error text. A new `sshpool.ErrConnect` wraps dial and handshake failures; it answers 502 with a fixed message. A stored key that will not decrypt or parse is logged and answered 500 with a fixed message, matching the credentials route. A cancelled request gets no response. OpenAPI 502 and 500 text updated on both routes.
8. Fixture host is `stale.example.invalid`.

**Spec**
1. A probe the connection does not survive is now an error, not a partial report. The probe ends with one liveness request, since a refused sub-probe and a dead socket fail the same way; `Test` returns the caller's context error when that is the cause. Covers both cancellation and the socket deadline. `TestCancellationUnblocksProbe` and `TestTimeoutBoundsProbe` assert a nil report.
2. `pin_unreadable` without a key is reported only for `ErrConnect`. A stored key that will not parse is reported as itself; `TestUnparseableKeyOutranksUnreadablePin`.
3. Same fix as Standards 7.

**One extra.** `ssh-test` had the same leak in its 200 `error` field, so it now follows the same split: transport text for `ErrConnect`, a fixed message otherwise. `TestConfirmHidesStoredKeyFault` covers both routes.

**Looked at and left as is**, so they do not read as oversights: the same five SSH fields in the three response builders and the two-layer `replace` flag (your 4 and 5); the four-arm error switch that `ssh-test` and the pin routes share in shape but not in responses; `SSHHostKeyEncrypted != ""` at five sites rather than an `Instance` method; the `TrimPrefix` on the credentials error rather than a typed error; a handshake that fails before the host-key exchange counting as unreachable for the unreadable-pin precedence; `ssh-rsa` SHA-1 absent from the algorithm list for non-RSA pins. Each is a few lines to change and I will take any of them if you want it in this PR; otherwise the walk-filter and pool changes in the next slice are where the shared shapes get a third caller.

s0up4200 (approved):

PR #2766, opened 2026-09-19
fix(api): preserve content length for torrent file downloads

schwarz:
## Description

Torrent content downloads went through the HTTP compressor, which drops Content-Length and strips the incoming Range header. Browsers could not show download progress or resume a file.

Compression now applies per route group instead of at the router root. The torrent content download route and the three SSE routes (`/stream`, `/logs/stream`, `/instances/{instanceID}/rss/events`) register in a group that keeps the raw writer, so `http.ServeContent` reports the file size and serves byte ranges, and stream events flush without the compressor's buffer. Every other route, including JSON, log files, Swagger, health, the proxy, and the web UI, still compresses. `/stream` keeps its own gzip through `gzipSessionWriter`.

The old Accept-header bypass and the path predicate are gone. The SSE routes are now uncompressed regardless of the Accept header, and a non-SSE route requested with `Accept: text/event-stream` compresses like any other.

## How has this been tested?

The compression tests in `internal/api/server_compression_test.go` rebuild each route's middleware chain from the production router around a stub handler. They cover the full and range download, the three SSE routes with and without `Accept: text/event-stream` (event delivered at flush, write deadline reachable), JSON and log file responses still gzipped, authentication on the uncompressed group, and two top-level routes (`/health` and the web root) that must stay on the compressed router. The download test proves the middleware chain, not the handler; the handler ends in `http.ServeContent`, so the chain is what needed proving.

Field test on a local rig (throwaway qBittorrent 5.x container seeding a 3,000,000 byte file, qui built from this branch): `GET .../files/0/download` with `Accept-Encoding: gzip` returned `200`, `Content-Length: 3000000`, `Accept-Ranges: bytes`, no `Content-Encoding`, body byte-equal to the file. `Range: bytes=100-199` returned `206`, `Content-Length: 100`, `Content-Range: bytes 100-199/3000000`. The same two requests against a develop binary (`v1.29.0-38-g3df380b9`) returned `200` with `Content-Encoding: gzip`, `Transfer-Encoding: chunked`, no Content-Length, and the whole file for the range request. `/api/logs/stream` and `/api/instances/1/rss/events` delivered their first event within 3 s uncompressed; `/api/stream?activity=1` delivered a heartbeat with its own gzip. `/api/openapi.json` and `/` still returned `Content-Encoding: gzip`.

## Performance

Compared merge-base `07b62da4` with `0510d8ff` using a temporary benchmark with each revision's middleware, Chi, ServeContent, and the production gzip settings. Measurements were repeated five times.

For the video sample, the response grew from 321.14 MiB to 323.12 MiB: an increase of only 1.98 MiB (0.62%). HTTP compression provided little additional size reduction for this already compressed video. A 4 MiB incompressible binary sample also showed no meaningful size increase. Regular JSON API responses produced identical compressed response sizes.

Round 2 (route groups): JSON, log, and SSE timings stayed within measurement noise. Full incompressible downloads improved from 0.896 to 0.558 ms; range requests now send only the requested bytes. Route wrappers add 8,842 allocated bytes (+0.059%) and 171 allocations per router at startup, with no measurable construction-time regression.

The bandwidth impact depends on the file contents: highly compressible files, such as plain text, transfer larger without compression. Downloads continue to stream without buffering the entire file.

## Checklist

## AI disclosure

Yes, hic sunt dracones. The new implementation and test code were AI-written using Codex & Astra/Sol, with human direction and review. Codex also carried out the performance measurements and wrote most of this PR description. The maintainer follow-up commit (boundary comments, top-level route coverage in the test, field test) was AI-assisted with Claude.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed compression-related issues affecting torrent downloads, event streams, and log streams.
  * Preserved reliable streaming behavior, including timely flushing and write deadlines.
  * Ensured protected download and streaming endpoints continue requiring authentication.

* **Performance**
  * Enabled gzip compression for larger API, documentation, health, and log responses when supported.
  * Kept small responses and real-time streams uncompressed to reduce overhead and maintain compatibility.


s0up4200 (changes requested):

Thanks, the fix is right and the tests prove it. I checked the premise against `httpcompression@v0.0.9`: adapter.go:89 deletes the request `Range` header before the handler runs, and response_writer.go:161 deletes `Content-Length` when compression starts. Both new subtests fail on develop and pass on this branch.

## Standards

**The path matcher rebuilds what chi knows.** `isTorrentContentDownload` at `internal/api/server.go:814` derives the mount prefix from `baseURL`, then splits and checks six segments by hand. `r.Mount(apiMount, apiRouter)` at `internal/api/server.go:771` computes that same prefix, so one fact now lives in two places and whoever changes one breaks the other. The three `!= ""` segment checks add nothing: a path with an empty segment cannot reach the download handler, so bypassing compression for it changes no behavior.

Two ways out, pick either:

- Ask the router. `router.Find(chi.NewRouteContext(), req.Method, req.URL.Path)` returns the route pattern, so the predicate becomes one string compare against `"/api/instances/{instanceID}/torrents/{hash}/files/{fileIndex}/download"` under the mount, and it follows `baseURL` for free.
- Skip the predicate. Move the compressor off the root mux and apply it per group with `r.With`, leaving the download route unwrapped.

**Optional: three sibling routes have the same defect.** `internal/api/handlers/backups.go:1135` (backup run and backup blob downloads) and `internal/api/handlers/logs.go:166` call `http.ServeContent` behind the same compressor, so they lose `Content-Length` and cannot serve ranges either. I do not need that in this PR. If you pick the per-group approach, covering them costs nothing extra, so take it if it falls out.

## Spec

The description says the new route check will *"keep the compression bypass aligned with the production endpoint."* The subtest at `internal/api/server_test.go:112` hardcodes the same path literal in both assertions, so nothing comes from the router. A route move trips `router.Find`, but no assertion feeds the predicate what the router serves. That makes it a tripwire, which has some value. If you go with the `router.Find` predicate above, the check holds by construction and can go. Either way I would move it out of `TestAllEndpointsDocumented`, whose subject is swagger coverage, into `server_compression_test.go`.

Standards: 1 finding plus 1 optional. Spec: 1 finding, the route check. Worst on each axis is the hand-rolled matcher and the hardcoded route check.

schwarz:
AI: 

- Routing: Compression now uses Chi route groups. Torrent content downloads and SSE routes bypass the general compressor; the main stream retains its own gzip support. The manual path matcher and header-based bypass are gone.
- Tests: Removed the compression tripwire from Swagger coverage. Compression tests now use middleware from the production router and cover downloads, ranges, SSE, authentication, and continued JSON/log compression.
- Scope: Left the optional backup/log changes out. Log downloads intentionally retain compression. Backup-run downloads generate archives directly, so removing compression alone would not give them range support.
- Performance: JSON, log, and SSE timings remained within measurement noise. Full incompressible downloads improved from 0.896 to 0.558 ms; range requests now send only the requested bytes. Route wrappers add 8,842 allocated bytes
  (+0.059%) and 171 allocations per router at startup, with no measurable construction-time regression.

Precommit, race tests for the affected packages, the build, and live checks passed. The remaining maintainer decision is whether that startup allocation tradeoff is acceptable.

s0up4200 (approved):

